PR #4, opened 2017-08-01
c++14 bump

rmsr:
* bump to c++14 (minimum g++ 5 and clang 3.8)
* fix stuff so compilers is happy
* add compiler version checking

### DIFF
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -1,6 +1,27 @@
 # hey emacs, this is -*- python -*-
 from SCons.Script.SConscript import SConsEnvironment
-import os.path;
+import os.path
+import subprocess
+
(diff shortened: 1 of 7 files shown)


code/SConstruct
```diff
@@ -1,6 +1,27 @@
 # hey emacs, this is -*- python -*-
 from SCons.Script.SConscript import SConsEnvironment
-import os.path;
+import os.path
+import subprocess
+
+CXX = 'g++'
+#CXX = 'clang'
```
cizra: Probably clang++?

rmsr: Nope, clang is explicitly a c++ compiler. Good thought though.

PR #6, opened 2017-08-02
Clean up build system

rmsr:
- Rewrite scons scripts
- Add makefile & install script
- Add travis-ci builds

also misc fixes attached


code/SConstruct
```diff
@@ -1,139 +1,205 @@
...
+    optimize=   off     Use -O3 optimization vs -O1
+    pretty=     on      Prettify output
+    shared=     on      Build MUD code as shared libraries
+    twitter=    off     Enable support for posting shouts to Twitter
+
+Options can be specified on the commandline or via environment variables.
+
+The defaults are optimized for developing with Sneezy rather than for
+deploying it. Good production flags: CXX=g++-6 harden=1 lto=1 shared=0
+(and optimize=1 if you don't need good crash debugging).
```
cizra: heh heh heh.

code/SConstruct
```diff
@@ -1,139 +1,205 @@
...
+flag_harden =   get_bool_option('harden',      True)
+flag_lto =      get_bool_option('lto',         False)
+flag_optimize = get_bool_option('optimize',    False)
+flag_pretty =   get_bool_option('pretty',      True)
+flag_shared =   get_bool_option('shared',      True)
+flag_twitter =  get_bool_option('twitter',     False)
 
+opt_includes =       get_option('CPPPATH',     '')
+opt_libdirs =        get_option('LIBPATH',     '')
+opt_olevel =         get_option('olevel',      '3' if flag_optimize else '1')
```
cizra: Does GCC reorder code with -O1?

code/SConstruct
```diff
@@ -1,139 +1,205 @@
...
+    # buffer overflow checker, only works for -O1 or greater
+    env.Append(CXXFLAGS='-D_FORTIFY_SOURCE=2')
+    # stack overflow checker
+env.Append(CXXFLAGS='-fstack-protector-strong')
+
+
+## optimize: compiles slow, bad for debugging, good for CPU
+opt_level = '-O1'
+if flag_optimize:
+    opt_level = '-O3'
```
cizra: Last I heard, the general consensus was that -O2 is the most reliable optimization level and anything above that should be profiled to prove effective.

rmsr: I think the biggest danger from -o3 is when people rely on "undefined behavior" in corner cases in the C spec, like signed integer overflow, where the compiler is free to say turn all your ints to 0xdeadbeef.

code/SConstruct
rmsr: Hmm, maybe a little, thanks for this poke, apparently there's -Og which is "optimize without breaking debugging".

code/SConstruct
rmsr: technically you should be reviewing #7 instead of this one, github won't let me reopen this.

rmsr: Also rediscovered -Og which tries to optimize without wrecking debug. But only clang 4 and above support it. So, we go -O0 to -Og when possible to -O1 for harden to -O2 for optimize (-O3 is harder on icache and might be more machine-specific). Testing now, will commit soon.

PR #7, opened 2017-08-02
Improve build system

rmsr:
- Rewrite scons scripts
- Add makefile & install script

also misc fixes attached


rmsr:
I have no solaris box nor mac to test on. I will set up a travis-ci mac build though! Which, haha, is the easiest way to get access to modern compilers and libs. I don't think solaris is worth keeping, I'll just clear that stuff out.

rmsr:
Pulled .travis.yml file for now, needs more lovin' and the rest of this stuff might as well merge, pending whatever OSX issues might be extant.

rmsr:
Oops, wrong branch, but I'm much closer to a satisfactory travis file.

rmsr:
Major changes to the scons files since your approve, my head whirls, I'm sure there's some corner cases that I've broken, but... the travis file is in, linux builds are working. Obviously I've had no joy with osx. The nicest new thing is that sneezy now creates all those empty lib directories if they're missing, so no more scripting up mkdirs during installation! Of course, I pulled in a new boost dependency for that, so be aware. Naturally with all the rebasing there's no good delta against the previous pull request.

PR #17, opened 2017-08-13
Clang scanner gameplay changes

rmsr:
Things like missing break, unused variables revealing a thinko, etc.

All of these affect gameplay in various ways, I don't think any of them have serious impact.

### DIFF
--- a/code/code/cmd/cmd_deathstroke.cc
+++ b/code/code/cmd/cmd_deathstroke.cc
@@ -98,19 +98,19 @@ static int deathstroke(TBeing *caster, TBeing *victim)
       if (caster->reconcileDamage(victim, 0,sktype) == -1)
         return DELETE_VICT;
     }
-    /* with great failure, comes great sorrow *grin* */
-    percent = ((10 - (caster->getArmor() / 50)) << 1);
-    percent += caster->getDexReaction() * 5;
-    percent -= victim->getAgiReaction() * 5;
+
+    // with great failure, comes great sorrow *grin*
     // what happens here is that the mob gets a shot at the players vitals
     // ... fair is fair right? 
+    percent = ((10 - (caster->getArmor() / 50)) << 1);
+    percent += victim->getDexReaction() * 5;
+    percent -= caster->getAgiReaction() * 5;
     int bKnown2 = victim->getSkillValue(SKILL_DEATHSTROKE);
     if (bKnown2 > 0 &&
-        victim->bSuccess(bKnown2 + 20, SKILL_DEATHSTROKE) &&
+        victim->bSuccess(bKnown2 + percent, SKILL_DEATHSTROKE) &&
          (i = victim->specialAttack(caster,SKILL_DEATHSTROKE)) &&
-         i != GUARANTEED_FAILURE) 
-    {
-      /* monster hits player vitals while player is exposed */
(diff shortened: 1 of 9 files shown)


Hordeking:
I haven't used clang, so I can't really tease out what this one is about? Is he fixing something so it compiles in clang? Is he just saying he was using clang when he made these changes? @cizra, What do you make of this set of changes? I would think the behavior changes take effect no matter what compiler one uses.

rmsr:
These are defects identified by the clang code sniffer tool: https://clang-analyzer.llvm.org/

rmsr:
To be more specific, clang-scan found various bits of code that failed the sniff test, even if they did compile. These changes attempt to fix spots where (what I perceive to be) the developer's intention was foiled by a thinko or code slippage.

rmsr:
Can I get some reviews on this one? I don't mind if someone NAKs specific commits because they think it distorts gameplay.

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -98,19 +98,20 @@ static int deathstroke(TBeing *caster, TBeing *victim)
...
-    percent = ((10 - (caster->getArmor() / 50)) << 1);
-    percent += caster->getDexReaction() * 5;
-    percent -= victim->getAgiReaction() * 5;
+
+    // with great failure, comes great sorrow *grin*
     // what happens here is that the mob gets a shot at the players vitals
     // ... fair is fair right? 
+    percent = ((10 - (caster->getArmor() / 50)) << 1);
+    percent += victim->getDexReaction() * 5;
+    percent -= caster->getAgiReaction() * 5;
```
cizra: Why'd you flip them? Dex is offense, Agi is defense, so it makes sense to check victim's Agi when attacking.

lib/txt/news
```diff
@@ -4,31 +4,18 @@
 07-04-17 : The world command now reports the correct number of distinct
```
cizra: Why did this file suddenly change so much?

code/code/misc/magicutils.cc
```diff
@@ -1325,10 +1325,7 @@ void TBeing::rawBlind(int level, int duration, saveTypeT save)
   aff.type = SPELL_BLINDNESS;
   aff.bitvector = AFF_BLIND;
   aff.level = level;
-  aff.duration = duration;
-
-  if (save)
-    duration /= 2;
+  aff.duration /= (save ? 2 : 1);
```
cizra: Eschew obfuscation

lib/txt/news
rmsr: cleaned up date formatting, out of order entries, indentation, and trailing whitespace.

rmsr: honestly I can't believe it still applies - we need more news entries for player-visible changes

code/code/cmd/cmd_deathstroke.cc
rmsr: this calculation is for the counterattack, where the "victim" is the attacker. the previous code was a copy-paste error.

cizra: Ohhh right!

rmsr:
De-obfuscated the magicutils comment, should be good to go yes?

PR #22, opened 2017-10-23
Removing exceptions to fundamental game rules from inactive immortals.

Hordeking:
I'm adminning the new server instance. Removed immortals from the list who don't really need to be on that list anymore. Added self and another active immortal.

### DIFF
--- a/code/code/misc/immortal.cc
+++ b/code/code/misc/immortal.cc
@@ -6010,14 +6010,17 @@ void TBeing::doAccount(const sstring &arg)
       // account, so i'm adding myself. - dash
       sstring tmp_name = getName();
 
-      if ((tmp_name != "Batopr") &&
-          (tmp_name != "Lapsos") &&
-          (tmp_name != "Damescena") &&
-          (tmp_name != "Peel") &&
-          (tmp_name != "Jesus") &&
-          (tmp_name != "Dash") &&
-          (tmp_name != "Maror") &&
-          (tmp_name != "Angus")) {
+      if (//(tmp_name != "Batopr") &&
+          //(tmp_name != "Lapsos") &&
+          //(tmp_name != "Damescena") &&
+          //(tmp_name != "Peel") &&
+          //(tmp_name != "Jesus") &&
+            (tmp_name != "Dash") &&
+          //(tmp_name != "Maror") &&
+          //(tmp_name != "Angus")
+            (tmp_name != "Juno") &&
+            (tmp_name != "Aion")
+         ) {
         sendTo("Sorry you suck too much to do this.\n\r");
         return;
       }


cizra:
Not sure what you're adminning -- sneezymud.com:7900 has been down for months. Care to elaborate?

Hordeking:
sneezymud.org 7900

We got the server up and running and hosted. We managed to get ahold of Andrew and basically whoever is running his server farm updated the firewall and didn't bother opening 7900. We're not holding our breath and went ahead and set up the new server. Aion has been trying to get ahold of you on fb, but you disappeared from there. Here's the group if you're interested. https://www.facebook.com/groups/345197135614921/

cizra:
Anyone can set up a dozen Sneezys -- what matters is continuity of the player data. I see my account is not working, thus you probably didn't carry over the data from Andrew's box? What about the wiki and stuff?

Hordeking:
No, you're right. Andrew seems to be willing to do this, but generically seems to not really be very responsive to anyone but Intuition. RMSR is on the same fb group I posted earlier, but didn't respond when tagged while I was working on getting the code to build. I think you fixed the same warnings/errors in disc_air.cc as I had to in order to build.

Anyway, you're welcome to come on and play with it for a bit. We're having some trouble with the shops not listing anything (something wrong with the sql query, I think). Here's the "coming soon" announcement on the facebook group. https://www.facebook.com/groups/345197135614921/permalink/1178605135607446/

cizra:
Back to the topic: why must we hardcode a list of player names anywhere? Why not find a proper solution like in #18.

Hordeking:
I'm in complete agreement with you on that one. I made these changes and committed them to my own repo before seeing your change. I got the impression those hardcoded names were somehow being excepted from certain checks normally applied to immortals for admin purposes (Lev 60 admins). I'd like nothing more than to rip out anything hardcoded like that. Having things like that hard coded is dangerous. I was actually looking for hardcoded logins and passwords when I found that bit The commenting out old ones and adding myself (since it didn't look like a hardcoded account) was essentially so someone had admin power, and since at the time I edited it, it was us two trying to troubleshoot and bootstrap, I just changed that bit of code.

We need all the help we can get from you on this and other stuff, since I've always been a pretty low level player, and I don't know the code very well at all. I want to see this mud great again.

cizra:
Can we please disentangle the hardcoded list thing from the spell/prayer effects fix? The fix should be merged, the hardcode should not be merged, IMHO.

Hordeking:
I also fixed a bad argument to a verify function call. It was trying to verify immmortals, and that created a directory immmortals.

cizra:
Sorry to rain on the parade, but
1) please separate the verify fix, it should be committed, and
2) the rest should not. The hardcoded usernames should be scrapped, not updated.

PR #28, opened 2017-11-05
Fixing web links

Hordeking:
Lots of broken softlinks. I fixed it. lib was probably one level further up when these links were created.


cizra:
Please separate the fix for web links from the pile of other commits.

Hordeking:
It was originally separate. I merged the branches on my own repo, then pushed it to my repo :(
So now when you look at it, you see it as it is now, not with the branch separate.

A beginner mistake.

PR #29, opened 2017-11-05
Buffer overflow fix gcc7

Hordeking:
GCC7 is smarter than previous GCCs. It did some code analysis voodoo and determined that there was a very likely potential for a buffer overflow and produced a warning/error about it. This change fixes it.


cizra:
Please separate the useful commits from the irrelevant like this:

```
git rebase -i upstream/master
# Vim opens. Erase all lines except the fix for overflows. Save and quit.
# verify outgoing stuff
git log ..upstream/master
git push --force
# GitHub pull request should update automatically
```

code/code/misc/guild.cc
```diff
@@ -1174,7 +1174,7 @@ char * display_permission(unsigned int perms) {
 }
 
 char * display_guild_flags(unsigned int flags) {
-  char buf[256];
+  char buf[512];
```
Hordeking: This is really important here. The longest possible output string is longer than 256 bytes.

cizra:
I'd like to get all except the inactive immortals thing. That code ought to be rethunk somehow.

PR #35, opened 2017-11-09
Fixing bad directory verification.

Hordeking:
This verification not only checks for the existence of a directory, it also creates it if it doesn't exist. So the directory it looks for was given with a typo, so it creates that, and then the code goes on to use the directory that didn't have the typo in the name. Could crash if it verifies immmortals, when immortals doesn't exist in the first place.


cizra:
My fix for repairs should be scrapped in favor of this approach. Also every mkdir in the setup instructions.

PR #36, opened 2017-11-09
Enabled multiplay by default, masks a crashbug

cizra:


cizra:
The compliance checks are crashy. By disabling them by default, we won't hit the crashing codepath.

PR #38, opened 2017-11-09
Fixed a buffer overflow

cizra:


cizra:
Disagreed:
```
If the resulting string would be longer than n-1 characters, the remaining characters are discarded and not stored, but counted for the value returned by the function.
```
http://www.cplusplus.com/reference/cstdio/vsnprintf/

See the `n-1`, it leaves room for the terminator. Besides, all the examples I found use the exact size. Besides, testing showed that the string _is_ terminated all right.

Hordeking:
The reason I mention it is that it almost seems as if that particular behavior is undefined. Essentially, my question was "does vsnprintf use buf_size to make sure that the null doesn't get put past the end", and the answer that I got was essentially "it depends on the implementation".

At any rate, you say you've tested it, so I'll approve it. I can believe that vsnprintf would make sure not to overflow if you explicitly tell it how big the buffer is.

PR #44, opened 2017-11-11
Added in-mud ability to manipulate the global load potential.

Hordeking:
It looks like a lot of changes, but all it really does is allows us to manipulate the global load potential from within the mud, rather than having to "tweak" it by 0.05 each time or manually editing the stats file when the mud is down.

### DIFF
--- a/code/code/misc/immortal.cc
+++ b/code/code/misc/immortal.cc
@@ -7,6 +7,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 #include <stdio.h>
+#include <cmath>
 
 #include "extern.h"
 #include "handler.h"
@@ -4344,22 +4345,73 @@ void TBeing::doInfo(const char *arg)
     else if (is_abbrev(arg1, "tweak")) {
       if (!hasWizPower(POWER_INFO_TRUSTED)) {
         sendTo("You should not attempt to change that.\n\r");
-	return;
+        return;
       }
       char arg2[80];
       arg = one_argument(arg,arg2, cElements(arg2));
-      if (is_abbrev(arg2, "loadrate")) {
+      if (is_abbrev(arg2, "loadrates")) {
         char opt[80];
-	arg = one_argument(arg,opt, cElements(opt));
-	if (is_abbrev(opt, "up"))
-	  stats.equip += .05;
-	else
-	  stats.equip -= .05;
-	save_game_stats();
-      } else {
-	sendTo(format("loadrate is %f\n\r") % stats.equip);
+	    arg = one_argument(arg,opt, cElements(opt));
+	    if (is_abbrev(opt, "up")) {
+	        stats.global_lp_target += .05;
+            stats.global_lp_target_changerate = fabs((stats.global_lp_target - stats.equip)/30);
+            save_game_stats();
+        }
+	    else if (is_abbrev(opt, "down")) {
+	        stats.global_lp_target -= .05;
+            stats.global_lp_target_changerate = fabs((stats.global_lp_target - stats.equip)/30);
+            save_game_stats();
+        }
+        else if ( is_abbrev(opt, "set" ) ){
+            sendTo(opt); sendTo("\n\r");
+
+            sstring usage = "Usage: info tweak loadrate set <target> [transition time in seconds]\n\r";
+
+            if (!*arg || !arg) {
+                sendTo(usage);
+                return;
+            } 
+            arg = one_argument(arg,opt, cElements(opt));
+            if (!is_number(opt) || convertTo<double>(opt)<0) {
+                sendTo("Invalid value given for global load potential. Must be a number >= 0.0.\n\r "+usage);
+                return;
+            }
+            stats.global_lp_target = convertTo<double>(opt);
+
+            if (!*arg || !arg)
+            {
+                // No change period given-just set it immediately. We're done.
+                stats.equip = stats.global_lp_target;
+                save_game_stats();
+                sendTo(format("Target Global Load Potential set to %f\n\r")%stats.global_lp_target);
+                return;
+            }
+
+            arg = one_argument(arg,opt, cElements(opt));
+            if (!is_number(opt) || convertTo<double>(opt)<=0) {
+                stats.global_lp_target = stats.equip;   //Restore the old target global load potential. This one was a wash.
+                sendTo("Invalid value given for transition time. Must be a positive number of seconds.\n\r "+usage);
+                return;
+            }
+            int change_period_seconds = convertTo<int>(opt);
+            stats.global_lp_target_changerate = fabs((stats.global_lp_target - stats.equip)/change_period_seconds);
+            sendTo(format("Target Global Load Potential set at %f")%stats.global_lp_target);
+            sendTo(format(" to change over %f seconds.\n\r")%convertTo<int>(opt));
+
+            save_game_stats();
+        }
+        else
+            sendTo(format("Global load probabiity is %f\n\r") % stats.equip);
       }
-    } 
+      else
+      {
+        sstring str = "Usage: Display or modifies the global load potential\n\r\n\r";
+        str += "\tinfo tweak loadrate\n\r";
+        str += "\tinfo tweak loadrate [up|down]\n\r";
+        str += "\tinfo tweak loadrate set <target global load potential> [transition time in seconds]\n\r";
+        sendTo(str);
+      }
+    }
     else if (is_abbrev(arg1, "deaths")) {
       if (!hasWizPower(POWER_INFO_TRUSTED)) {
         sendTo("You cannot access that information.\n\r");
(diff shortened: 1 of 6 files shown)


code/code/misc/parse.cc
```diff
@@ -2081,16 +2083,8 @@ void argument_interpreter(sstring argument, sstring &first_arg, sstring &second_
...
-  if (str.empty())
-    return (0);
-
-  for (look_at = 0; *(str.c_str() + look_at) != '\0'; look_at++) {
-    if ((*(str.c_str() + look_at) < '0') || (*(str.c_str() + look_at) > '9'))
-      return (0);
-  }
-  return (1);
+    boost::smatch what;  //Not really used here, since we're just doing go/no-go
+    return boost::regex_match(str, what, boost::regex("^[\\+-]?\\d+(\\.\\d+)?$"));
```
Hordeking: This is the big one to watch, since it's used in a few places to validate input. Previously, it only returned true for text integers, but false for anything that had anything but a zero thru nine in it. So while it correctly invalidated words and letters and other garbage, it also came back as invalid for floating point numbers, negative numbers, and so on. This change fixes that. If the number is preceeded by a plus or minus, it will now be validated correctly. If it has a decimal, it must have at least one digit to the right, otherwise it will come back as invalid. And of course, no decimal point at all with/without signs will also work (signed/unsigned integer).

code/code/misc/statistics.cc
```diff
@@ -268,7 +276,7 @@ void save_game_stats(void)
          gold_modifier[GOLD_SHOP_RESPONSES].getVal(),
          gold_modifier[GOLD_DUMP].getVal());
 
-    fprintf(fp, "%f\n", stats.equip);
+    fprintf(fp, "%f %f %f\n", stats.equip, stats.global_lp_target, stats.global_lp_target_changerate);
```
Hordeking: Outputs the new stats to the log now, for persistence across boots.

code/code/misc/immortal.cc
```diff
@@ -4344,22 +4345,73 @@ void TBeing::doInfo(const char *arg)
...
+                return;
+            }
+            stats.global_lp_target = convertTo<double>(opt);
+
+            if (!*arg || !arg)
+            {
+                // No change period given-just set it immediately. We're done.
+                stats.equip = stats.global_lp_target;
+                save_game_stats();
+                sendTo(format("Target Global Load Potential set to %f\n\r")%stats.global_lp_target);
```
cizra: Changes like this deserve a logline

code/code/misc/immortal.cc
```diff
@@ -4344,22 +4345,73 @@ void TBeing::doInfo(const char *arg)
...
+            }
+            int change_period_seconds = convertTo<int>(opt);
+            stats.global_lp_target_changerate = fabs((stats.global_lp_target - stats.equip)/change_period_seconds);
+            sendTo(format("Target Global Load Potential set at %f")%stats.global_lp_target);
+            sendTo(format(" to change over %f seconds.\n\r")%convertTo<int>(opt));
+
+            save_game_stats();
+        }
+        else
+            sendTo(format("Global load probabiity is %f\n\r") % stats.equip);
```
cizra: typo

code/code/misc/parse.cc
cizra: I'd've used `R"(^[\+-]?\d+(\.\d+)?$)"` -- less confusing backslashes this way.

code/code/misc/statistics.cc
```diff
@@ -175,8 +175,16 @@ int init_game_stats(void)
 
     SetupStaticGoldModifiers();
 
-    if (fscanf(fp, "%f\n", &stats.equip) != 1) {
-      vlogf(LOG_BUG, "bad value for equipment load rate");
+    if (fscanf(fp, "%lf\n", &stats.equip) != 1) {
+      vlogf(LOG_BUG, "bad value for equipment load potential");
+    }
+
+    if (fscanf(fp, "%lf\n", &stats.global_lp_target) != 1) {
```
cizra: Pretty sure this will break compatibility. New code won't be able to read old stats files. Why don't we put it into the DB, where this wouldn't be an issue, and moreover where we want to move things anyway?

code/code/misc/statistics.cc
cizra: So you _read_ the new format but _write_ the old one?

code/code/sys/socket.cc
```diff
@@ -567,16 +568,35 @@ procTweakLoadRate::procTweakLoadRate(const int &p)
...
+  if ( stats.global_lp_target_changerate <= 0 ) return;     // If this is zero, the procedure won't change anything anyway.
+
+
+  // target_delta = (stats.global_lp_target - original_load)/(24hr/day*3600s/hr*delta_time_in_days)
+
+  // If the current loadrate is within one delta of the target, we will
+  // call it good. Clamp the load rate to target just to be safe and
+  // prevent hunting.
+  if( 
+        (stats.global_lp_target-stats.global_lp_target_changerate <= stats.equip)
```
cizra: The parentheses here are not needed (though you can retain them if you think it helps)

code/code/sys/socket.cc
```diff
@@ -567,16 +568,35 @@ procTweakLoadRate::procTweakLoadRate(const int &p)
...
+    stats.equip = stats.global_lp_target;
+    stats.global_lp_target_changerate = 0.0f;
+    save_game_stats();
+    vlogf(LOG_MISC, format("procTweakLoadRate: desired load potential of %f achieved.") % stats.equip);
     return;
   }
 
-  stats.equip -= 0.00019166666666666666;
+  // Determines if we need to go up or down to get closer to the target load potential.
+  (stats.equip > stats.global_lp_target) ? (stats.equip -= stats.global_lp_target_changerate) : (stats.equip += stats.global_lp_target_changerate);
```
cizra: An `if` statement would look less confusing

code/code/sys/socket.cc
```diff
@@ -567,16 +568,35 @@ procTweakLoadRate::procTweakLoadRate(const int &p)
...
   save_game_stats();
-  vlogf(LOG_LOW, format("procTweakLoadRate: adjusted load rate to %f") %
+  vlogf(LOG_MISC, format("procTweakLoadRate: adjusted load potential to %f") %
```
cizra: It's a `lf` now

code/code/sys/socket.cc
```diff
@@ -1810,7 +1830,7 @@ int TMainSocket::gameLoop()
   scheduler.add(new procFactoryProduction(Pulse::MUDDAY));
 
   // pulse realhour
-//  scheduler.add(new procTweakLoadRate(Pulse::REALHOUR)); // desired load rate achieved
+  scheduler.add(new procTweakLoadRate(Pulse::ONE_SECOND));
```
cizra: Uhh, do we need to run it _every_ second? I do realize the checks in it are cheap, but still.

code/code/misc/statistics.cc
Hordeking: No, it writes to the new format as well. It reads three values, and writes out three values.

Hordeking: I just happened to stick a remark on the spot where the read is. Check lines 178-188.

code/code/misc/parse.cc
Hordeking: I didn't even see this until just now. I did exactly that a little bit earlier.

code/code/misc/statistics.cc
Hordeking: I'm not sure there's really a good way around this. I manually added the two numbers into the stats file to make this work. We could put them at the end if we wanted, would that be satisfactory? My preference is for keeping the related statistics together, which is why I did it the way I did.

Hordeking: I just moved these values to the end of the file, so any old versions reading the file won't break anything.

code/code/sys/socket.cc
Hordeking: I'll leave them for now. I've been bitten too many times by not having them and a different operator actually had higher precedence. That was a painful 4 hours of bug hunting.

code/code/sys/socket.cc
Hordeking: We can change it, but I prefer its compactness. in this context. If you _really_ want it changed, we can do that.

code/code/sys/socket.cc
Hordeking: We can probably change this to something else. Optimally, I would like to be able to remove this proc from the scheduler when not in use. I'd need to add a function to search for it and do that. I haven't looked very hard in how to access the scheduler instance from other locations besides the main game loop, but once we do that, we can just add and remove as we like.

As for it being cheap, if the change rate is set to zero, the first thing it does is returns, so it isn't much of a hit when not in use.

code/code/misc/statistics.cc
```diff
@@ -278,6 +289,8 @@ void save_game_stats(void)
     }
 
     fprintf(fp, "%ld\n", (long) stats.first_login);
+
+    fprintf(fp, "%.18lf %.18lf\n", stats.global_lp_target, stats.global_lp_target_changerate);
```
cizra: Writing 2 values on 1 line here, but reading them from 2 lines above

PR #46, opened 2017-11-14
    Reduced spam

Hordeking:
- Reduced log spam related to gradual load potential changes.
- Removed extraneous sendTo that was intended for debugging.


code/code/sys/socket.cc
```diff
@@ -589,7 +589,7 @@ void procTweakLoadRate::run(const TPulse &) const
     stats.equip = stats.global_lp_target;
     stats.global_lp_target_changerate = 0.0f;
     save_game_stats();
-    vlogf(LOG_MISC, format("procTweakLoadRate: desired load potential of %lf achieved.") % stats.equip);
+    //vlogf(LOG_MISC, format("procTweakLoadRate: desired load potential of %lf achieved.") % stats.equip);
```
cizra: I think this log line should be left in. Remove the one on line 599 instead.

PR #53, opened 2017-11-17
Fix the alias table

cizra:


code/code/sys/migrations.cc
```diff
@@ -35,7 +35,7 @@ void runMigrations() {
                     "create table alias ("
                     "id int primary key auto_increment not null, "
```
Hordeking: Is a primary key automatically unique? If not, should we make it so?

PR #74, opened 2017-12-06
Global burn rate

sntwo:
Added burn rate to modify the rate burning rounds damage objects.  Changeable in game in info tweak.

### DIFF
--- a/code/code/misc/statistics.cc
+++ b/code/code/misc/statistics.cc
@@ -69,6 +69,8 @@ int init_game_stats(void)
   stats.global_lp_target_changerate = 0.0;  // This is how fast the load potential changes.
   stats.max_exist = 1.2;  // this affects the MAX number of a thing allowed
 
+  stats.burnrate = 0.4;  // this affects the rate at which things burn
+
   // 1.40 resulted in 16-20 days playtime to L50
   // 1.05 resulted in 25-30 day to L50 (4.1)
   // 0.80 had reasonable rages for 4.5 beta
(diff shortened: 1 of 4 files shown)


code/code/misc/statistics.cc
```diff
@@ -207,6 +209,10 @@ int init_game_stats(void)
       vlogf(LOG_BUG, "bad value for rate of change of equipment load potential");
     }
 
+    if (fscanf(fp, "%lf\n", &stats.burnrate) != 1) {
```
cizra: Why not store it in DB? You could reuse the `configuration` table, it's a key-value mapping -- or even better, create a new table for all stats. See `code/sys/migrations.cc` for a comfy way to do that.

code/code/misc/statistics.cc
```diff
@@ -69,6 +69,8 @@ int init_game_stats(void)
   stats.global_lp_target_changerate = 0.0;  // This is how fast the load potential changes.
   stats.max_exist = 1.2;  // this affects the MAX number of a thing allowed
 
+  stats.burnrate = 0.4;  // this affects the rate at which things burn
```
cizra: Shouldn't the default be 1?

code/code/misc/statistics.h
```diff
@@ -107,6 +109,7 @@ class GameStats {
       piety_gained(0.0),
       piety_gained_attempts(0),
       equip(0.0),
+      burnrate(0.0),
```
cizra: uhhh.. 1?

code/code/misc/statistics.cc
sntwo: This makes sense.  I was trying to keep my first commit simple so I just copied the stats.load format.  There are many magic constants scattered around that could be pulled into this.  I'll look into it.

PR #83, opened 2017-12-10
Global settings db

sntwo:
Makes a table for global constants (global tweaks) and allows modification in game.  Moved stats.equip and stats.burnrate into said table and also added an obj freeze modifier.  Many more constants to come!


code/SConstruct
```diff
@@ -259,4 +259,4 @@ env.SConscript(
 
 
 ## etags
-PhonyTargets(env, tags='@etags code/*/*')
+PhonyTargets(env, tags='@etags code/*/*')
```
cizra: Why you eated my newline :(

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
+//////////////////////////////////////////////////////////////////////////
+//
+// SneezyMUD - All rights reserved, SneezyMUD Coding Team
+//
+//  tweaks.cc
+//
+//////////////////////////////////////////////////////////////////////////
```
cizra: I think we're AGPL or somesuch, these days -- we reserve really few rights. Best remove this whole block

code/code/misc/tweaks.h
```diff
@@ -0,0 +1,78 @@
+//////////////////////////////////////////////////////////////////////////
+//
+// SneezyMUD - All rights reserved, SneezyMUD Coding Team
```
cizra: almost no rights reserved

code/code/misc/tweaks.h
```diff
@@ -0,0 +1,78 @@
...
+//  the desired value.
+//
+//  Note that if you are changed the value of related tweaks at the same 
+//  time it might be hard to figure out the effect in game.
+//
+//////////////////////////////////////////////////////////////////////////
+
+
+#ifndef __TWEAK_H
+#define __TWEAK_H
```
cizra: Every compiler under the sun understands `#pragma once`, which is 2 lines less boilerplate AND safer against user error

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+
+tweakEntry::tweakEntry(int i,double v, double t, double r) :
+  id(i),
+  cvalue(v),
+  tvalue(t),
+  rate(r)
+{
+}
+
+tweakEntry::~tweakEntry()
```
cizra: What for?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+{
+  return tweaks.at(t);
+}
+
+tweakTypeT & operator++(tweakTypeT &c, int)
+{
+  return c = (c == MAX_TWEAK_TYPES) ? TWEAK_NONE : tweakTypeT(c+1);
+}
+
+tweakInfoT::~tweakInfoT()
```
cizra: What for?

code/code/sys/migrations.cc
```diff
@@ -61,6 +60,30 @@ void runMigrations() {
...
+                    "tweak_type int not null, "
+                    "tweak_value float(20) not null, "
+                    "tweak_target float(20) not null, "
+                    "tweak_rate float(20) not null, "
+                    "datecreated datetime not null default CURRENT_TIMESTAMP)"
+                    );
+
+            //this doesn't work / migrations run before stats load      
+            //sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values ('1','%f','%f','%f')", stats.equip, stats.global_lp_target, stats.global_lp_target_changerate);
+            //another way would be to parse the stats file for current stats.equip value etc., but maybe an admin can just set the value after migration...
```
cizra: The right way would be to load the stats file, if not unreasonably complex. If too complex, hardcode the defaults in constructor with a clean but migrated DB.

code/code/sys/migrations.cc
```diff
@@ -61,6 +60,30 @@ void runMigrations() {
...
+            
+            //imm says stats.equip is 112.5
```
cizra: Why not defaults in tweakInfoT constructor, instead?

code/code/sys/migrations.cc
```diff
@@ -61,6 +60,30 @@ void runMigrations() {
...
+            sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values (1,112.5,0.7,0.0)");//loadrate
+
+            //burnrate and freezerate are 1.0 anyway.
+            //sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values (2,1.0,1.0,0.0)");//burnrate
+            //sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values (3,1.0,1.0,0.0)");//freezedamrate
```
cizra: We don't need these lines

code/code/misc/tweaks.h
```diff
@@ -0,0 +1,78 @@
...
+  TWEAK_FREEZEDAMRATE,
+  MAX_TWEAK_TYPES
+};
+
+class tweakEntry {
+  public:
+  int id;
+  double cvalue;
+  double tvalue;
+  double rate;
```
cizra: No value in naming a value a value. Let's use `current`, `target`, `rate` instead

code/code/sys/socket.cc
```diff
@@ -561,42 +561,50 @@ void procMobHate::run(const TPulse &) const
...
+  TDatabase db(DB_SNEEZY);
 
-  if ( stats.global_lp_target_changerate <= 0 ) return;     // If this is zero, the procedure won't change anything anyway.
 
+  for(tweakTypeT i=TWEAK_LOADRATE;i<MAX_TWEAK_TYPES;i++){
+    
+    tweakEntry* t = tweakInfo[i];
+    if (t->rate == 0.0) {
+      continue;
+    }
```
cizra: This ...

code/code/sys/socket.cc
```diff
@@ -561,42 +561,50 @@ void procMobHate::run(const TPulse &) const
...
 
-  // target_delta = (stats.global_lp_target - original_load)/(24hr/day*3600s/hr*delta_time_in_days)
+    if ( t->tvalue <= 0 ) return;     // If this is zero, the procedure won't change anything anyway.
```
cizra: ... and this style could be unified a bit. Also, the extra spaces inside `()` don't follow the majority of style

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+{
+}
+
+tweakInfoT::tweakInfoT(): 
+    tweaks(MAX_TWEAK_TYPES + 1, nullptr), 
+    loaded(false)
+{
+  loaded=false;
```
cizra: not needed

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+  return c = (c == MAX_TWEAK_TYPES) ? TWEAK_NONE : tweakTypeT(c+1);
+}
+
+tweakInfoT::~tweakInfoT()
+{
+}
+
+tweakInfoT::tweakInfoT(): 
+    tweaks(MAX_TWEAK_TYPES + 1, nullptr), 
+    loaded(false)
```
cizra: not needed, it's nicer to write `bool loaded = false` straight in header

code/code/misc/tweaks.h
cizra: Perhaps hardcode defaults right here as NSDMI?

code/code/misc/tweaks.h
```diff
@@ -0,0 +1,78 @@
...
+  tweakEntry *operator[] (const tweakTypeT);
+
+  tweakInfoT();
+  ~tweakInfoT();  
+};
+
+extern tweakInfoT tweakInfo;
+extern tweakTypeT & operator++(tweakTypeT &c, int);
+
+#endif
```
cizra: Emacs, huh?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+           "where tweak_id = (select max( tweak_id ) from globaltweaks b where a.tweak_type = b.tweak_type)"
+  );
+  //enter in the proper values. Is it possible for the query to fail and the mud still be running?
+  int i = 0;
+  while(db.fetchRow()){
+    tweakTypeT tweak_id=(tweakTypeT) convertTo<int>(db["tweak_type"]);
+    int tid=(int) convertTo<int>(db["tweak_id"]);
+    double val=(double) convertTo<double>(db["tweak_value"]);
+    double tar=(double) convertTo<double>(db["tweak_target"]);
+    double rate=(double) convertTo<double>(db["tweak_rate"]);
```
cizra: The `(double)` casts probably aren't necessary?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+  for(tweakTypeT i=TWEAK_NONE;i<MAX_TWEAK_TYPES;i++){
+    tweaks[i] = new tweakEntry(0,1.0, 1.0, 0.0);
+  }
+
+  //the globaltweaks table contains multiple rows per tweak type
+  //We only want to keep the last tweak update of each tweak type in memory, so we do a greatest n per group query
+  db.query("select tweak_id, tweak_type, tweak_value, tweak_target, tweak_rate from globaltweaks a "
+           "where tweak_id = (select max( tweak_id ) from globaltweaks b where a.tweak_type = b.tweak_type)"
+  );
+  //enter in the proper values. Is it possible for the query to fail and the mud still be running?
```
cizra: yes with clean DB

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+
+  //On load fill out all tweaks with 1.0.  This will make adding new tweaks easier by eliminating the need to write a migration.
+  for(tweakTypeT i=TWEAK_NONE;i<MAX_TWEAK_TYPES;i++){
+    tweaks[i] = new tweakEntry(0,1.0, 1.0, 0.0);
+  }
+
+  //the globaltweaks table contains multiple rows per tweak type
+  //We only want to keep the last tweak update of each tweak type in memory, so we do a greatest n per group query
+  db.query("select tweak_id, tweak_type, tweak_value, tweak_target, tweak_rate from globaltweaks a "
+           "where tweak_id = (select max( tweak_id ) from globaltweaks b where a.tweak_type = b.tweak_type)"
```
cizra: I didn't get that part. Why multiple rows per tweakable?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+  loaded=true;
+  return i;
+}
+
+tweakInfoT tweakInfo;
+
+//recieved after immortal enters 'info tweak <arg>'
+void tweakInfoT::doTweak(TBeing *b, const char *arg)
+{
+  char arg2[256];
```
cizra: Instead of these yummy overflowables, let's use `sstring::word()`

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+        handleTweak(b, t, arg);
+        return;
+      }
+    }
+  }
+
+  //if they got here they typed info tweak gibberish
+  showTweaks(b);
+  showTweakOptions(b);
+  return;
```
cizra: Can we do validation in the beginning, and if it passes fine, assume no errors?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+  tweakEntry *tweak = tweakInfo[t];
+  TDatabase sneezy(DB_SNEEZY);
+
+  //only one arguements given after type
+  if (!*arg) { 
+    if (is_abbrev(opt, "up")) {
+      tweak->tvalue += .05;
+      tweak->rate = fabs((tweak->tvalue - tweak->cvalue)/30);
+      b->sendTo(format("You tweaked %s up a tad.\r\n") % getTweakName(t));
+      sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values (%i,%f,%f,%f)",
```
cizra: Why on earth do we want history on these?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+      tweak->rate = fabs((tweak->tvalue - tweak->cvalue)/30);
+      b->sendTo(format("You tweaked %s up a tad.\r\n") % getTweakName(t));
+      sneezy.query("insert into globaltweaks (tweak_type, tweak_value, tweak_target, tweak_rate) values (%i,%f,%f,%f)", 
+        t, tweak->cvalue, tweak->tvalue,tweak->rate);
+      sneezy.query("select max( tweak_id ) as tid from globaltweaks");
+      if (sneezy.fetchRow()) {
+        tweak->id = (int) convertTo<int>(sneezy["tid"]);
+      }
+    }
+	  else if (is_abbrev(opt, "down")) {
```
cizra: I'm sure we can collapse these 2 cases, up and down

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+    else {
+      showTweakOptions(b);
+    }
+  }
+
+  //multiple arguements given after type
+  else if ( is_abbrev(opt, "set") ){
+    sstring usage = "Usage: info tweak <parameter> set <target> [transition time in seconds]\r\n";
+
+    if (!arg || !*arg) {
```
cizra: Are you sure either can happen?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+      sneezy.query("select max( tweak_id ) as tid from globaltweaks");
+      if (sneezy.fetchRow()) {
+        tweak->id = (int) convertTo<int>(sneezy["tid"]);
+      }
+      return;
+    }
+
+    arg = one_argument(arg,opt, cElements(opt));
+    if (!is_number(opt) || convertTo<double>(opt)<=0) {
+      tweak->tvalue = tweak->cvalue;   //Restore the old target global load potential. This one was a wash.
```
cizra: Misconfigure then roll back? Let's validate first and then act

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+      b->sendTo("Invalid value given for transition time. Must be a positive number of seconds.\r\n "+usage);
+      return;
+    }
+
+    int change_period_seconds = convertTo<int>(opt);
+    tweak->rate = fabs((tweak->tvalue - tweak->cvalue)/change_period_seconds);
```
cizra: We disagreed with Juno about whether seconds is the right unit to choose, but she wrote the code so seconds it is

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+    case TWEAK_LOADRATE:
+      return "LoadRate";
+    case TWEAK_BURNRATE:
+      return "BurnRate";
+    case TWEAK_FREEZEDAMRATE:
+      return "FreezeDamChance";
+    case MAX_TWEAK_TYPES:
+      return "";
+  }
+  return "";
```
cizra: I'd assert here

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,268 @@
...
+  str += "\tinfo tweak <parameter> [up|down|history]\r\n";
+  str += "\tinfo tweak <parameter> set <target> <transition time in seconds>\r\n";
+  b->sendTo(str);
+}
+
+//no spaces - they will mess up parsing. ideally < 20 characters to be pretty
+const sstring tweakInfoT::getTweakName(tweakTypeT tt){
+  switch (tt) {
+    case TWEAK_NONE:
+    return "none";
```
cizra: I'd assert here.

code/code/misc/tweaks.h
```diff
@@ -0,0 +1,78 @@
...
+
+#include <map>
+
+#include "extern.h"
+#include "sstring.h"
+#include "being.h"
+
+
+enum tweakTypeT {
+  TWEAK_NONE  = 0,
```
cizra: What's this for?

code/code/misc/tweaks.cc
sntwo: A new row is added every time the tweakable is changed by hand so the history can be shown.  I didn't want to add a new row every second in procTweakRates, so for tweakables that have rates set, there is one row to mark the manual set point and a row to hold the current value.

code/code/misc/tweaks.cc
sntwo: If someone makes a tweakable for xp or something really important one might want to know...

code/code/sys/migrations.cc
sntwo: Stats.equip is buried in the middle of txt/stats so it would essentially be almost a copy of the whole file load.  Also I should probably purge stats.equip out of the file load and save while here, but I would like to put that off until I have moved more things out of stats.

cizra: Isn't it possible to call some `readStatsFile()` or such?

code/code/misc/tweaks.cc
```diff
@@ -0,0 +1,232 @@
...
+    sstring usage = "Usage: info tweak <parameter> set <target> [transition time in seconds]\r\n";
+
+    arg = one_argument(arg,opt);
+
+    if (!is_number(opt) || convertTo<double>(opt)<0) {
+      b->sendTo("Invalid value given for global potential. Must be a number >= 0.0.\r\n "+usage);
+      return;
+    }
+
+    int targ = convertTo<double>(opt);
```
cizra: Use double here, or better yet, `auto`
```
 20:53:28 H:17 V:333 R:0 > ⇨info tweak freezedamchance set 2
// Misc: Imm set FreezeDamChance to 2.000000 Target global FreezeDamChance set to 2.000000 

20:53:29 H:17 V:333 R:0 > ⇨info tweak freezedamchance set 0.5
// Misc: Imm set FreezeDamChance to 0.000000 Target global FreezeDamChance set to 0.000000
```

sntwo:
Woops, thought dismissing that last review will unblock the merge.

cizra:
Unblocking is manual.

cizra:
I still think there's stuff to fix here, but not enough to hold it back from master.

PR #87, opened 2017-12-13
Less spammy logs

cizra:


code/code/sys/gmcphandlers.cc
```diff
@@ -0,0 +1,138 @@
+#include "gmcphandlers.h"
+#include "being.h"
```
sntwo: code/sys/gmcphandlers.cc:6:20: fatal error: json.hpp: No such file or directory
 #include "json.hpp"

On docker in windows 10.

cizra:
You need to update Git submodules, I added a new one.

PR #108, opened 2018-01-18
Boot mods

cizra:


rmsr:
Someday we'll get rid of all those errors and warnings during zone loading too. Good thinking on the transaction wrapper.

rmsr:
Despite being in the hackers group, and approving your previous PR, I don't seem to have approve access for this one for some reason. Anyway, I approve!

PR #114, opened 2018-01-19
Remove obsolete cxxtest tests

rmsr:
I'm not sure unit tests are very interesting, we really need functional tests against a running MUD instance.


cizra:
I see value in unittesting complex and self-contained stuff like color conversions.

rmsr:
I don't disagree that we should also have unittests, but I think the test harness we end up with will NOT be cxxtest. We can recycle the contents of these tests easily from the repo history when we get to that point. Though I guess keeping them in HEAD will prevent us from forgetting about them. I guess I should withdraw this PR for now?

cizra:
I'd keep stuff until we have new stuff to replace it.

PR #115, opened 2018-01-19
Fixed socket code - was confused about how you should accept connections

cizra:


rmsr:
I was going to make some minor style requests, like removing the misleading v6_ prefix on some of the variables, (since we don't have any v4 sockets what's the diff) but really, why bother haha.

PR #116, opened 2018-01-19
Implemented Reboot command that exits with exitcode=1

cizra:


cizra:
I know what LGBT is, not sure about LGMT though...

cizra:
My best guess was "Let's Get That Merged" 😄

rmsr:
That's even better!

Hordeking:
I approve. It works. I should probably implement a shutdown 0 and a shutdown abort command.

PR #117, opened 2018-01-19
Relock containers on reboot if needed

cizra:


code/code/obj/obj_open_container.cc
```diff
@@ -65,12 +65,23 @@ void TOpenContainer::assignFourValues(int x1, int x2, int x3, int x4)
   setCarryVolumeLimit(x4);
 }
 
+int TOpenContainer::getDefaultContainerFlags() const {
+  TObj* prototypeRaw = read_object(objVnum(), VIRTUAL);
+  TOpenContainer* prototype = dynamic_cast<TOpenContainer*>(prototypeRaw);
+  int out = 0;
+  if (prototype != nullptr)
```
rmsr: this isn't what nullptr is for, just use the idomatic `if (prototype)`.

cizra:
Doors are read out of DB `room`, `roomexit`, `roomextra` tables during boot. Containers are generated during zone reset out of zonefiles. Totally different code paths.

code/code/obj/obj_open_container.cc
cizra: Brevity vs explicitness? I'm happy with either, so if it makes you feel better...

rmsr:
Hold off and let me investigate for a few days, if it turns out to be too much trouble we'll use this simpler fix.

rmsr:
I'll look at this tomorrow I promise

cizra:
Which tomorrow?

code/code/obj/obj_open_container.cc
rmsr: I think in this case explicitness is actually subtracting information from the code: now I have to assume that we compare against null because treating prototype as a boolean doesn't do what we need. In more complex examples this kind of assumption can lead to wasted time or bugs.

rmsr:
Today tomorrow! I'm working a new branch actually, with an alternate implementation, like the limbstun code.

rmsr:
I'm hoping #166 covers this need ok?

rmsr:
Closing in favor of #166

PR #119, opened 2018-01-20
make plasma mirror and thorn flesh trigger wimpy

sntwo:


sntwo:
Tested and it works to trigger wimpy.  However, I should note that the change in code is from the victim essentially hitting themself to the victim getting hit by the mage/shaman.  This means that the hatred mechanics are now activated by plasma mirror and thorn flesh damage, and the mage/shaman should get xp from that damage.  This might open up some exploit to getting good xp from these spells, or especially if someone has a plasma mirror affect item, but also makes them more dangerous to use...

rmsr:
I think all the things you describe are perfectly sensible.

PR #121, opened 2018-01-20
Remove inappropriate use of nullptr

rmsr:
nullptr is for use when a literal 0 carries incorrect type information, for instance a badly-designed method overload, one with an int arg and one with a pointer arg, where NULL matches against the int and not the pointer. Its presence implies typing issues that are not present in our codebase, and is thus confusing and non-idiomatic.


cizra:
I understand the distinction when calling overloads -- but I absolutely don't get why nullptr implies issues of any kind. For me, nullptr is the default way of writing a null pointer.

I can understand `if (!ptr)` is shorter, so I'm OK with these changes, but not with expressing null pointers as NULL/0.

rmsr:
So I checked some style guides, the ISO guide strongly implies using nullptr when possible, and the google one explicitly requires it. After sleeping on it, I think I like nullptr really.

However, testing explicitly against nullptr should still be used only when you actually need the extra typing information, as empty pointers of any kind cast to false in a boolean context. I've revised this PR to just address the comparison issue, and leave the nullptr call arguments in.

Heck, now I'd say we should do a global s/NULL/nullptr/g if it wouldn't mess up blame so much. Maybe when I get finally get sufficiently frustrated by the terrible indenting and run a code formatter on everything I'll rewrite to nullptr as well, haha.

PR #125, opened 2018-01-21
Fix forking and tmpfile writing in bug cmd

rmsr:
Previously sneezy would fork itself into the background, and try to write to the wildly obsolete /usr/tmp.


rmsr:
I think I will rewrite this using boost::process and no tmpfiles.

rmsr:
The author didn't want to block sneezy while writing to the tmpfile. If we had a more robust event loop (or if I was more familiar with its features I guess) we could do some async i/o with sendmail's stdin. For now I will retain the forking, but talk to sendmail thru a pipe rather than a tmpfile - sendmail is not guaranteed to complete its business in a timely manner.

rmsr:
boost::process is an undocumented bowel purge (literally not intended to be used directly), just sticking with this since it "works" and we'll be dumping it soon anyway.

code/code/misc/other.cc
```diff
@@ -4175,20 +4173,27 @@ void Descriptor::send_feedback(const char *subject, const char *msg)
...
-    {
-        fputs(message.c_str(), fp);
-        fclose(fp);
-        int err = system((format("/usr/sbin/sendmail -t < %s") % tempfile).str().c_str());
-        if (err)
-            vlogf(LOG_MISC, format("Call to sendmail returned nonzero status %d") % err);
-        remove(tempfile.c_str());
+  if (!vfork()) {
+    char tempfile[32];
+    strcpy(tempfile, "/tmp/sneezy_sendmail_XXXXXX.tmp");
```
cizra: Why not `char tempfile[] = "/tmp/";`

rmsr: mkstemps requires a full template to work with. it replaces the Xes with random values and then securely creates a new temp file from the resulting path.

rmsr:
Another alternative would be to append to a flatfile in the short term, that'd be a really quick fix! Should I do that?

code/code/misc/other.cc
cizra: Why can't we use a string literal to give it an initial value?

rmsr: mkstemps would then be trying to modify the RO string literal, no? I'm sure there's some handy C++ way to initialize the array with a copy of the literal instead of just pointing to it, but everyone seems to be using strcpy...

cizra: You're declaring a `char[]`, not a `const char*`, so it'll get its own copy. This works fine:

```cpp
int main() {
  char str[] = "bla\n";
  str[0] = 'f';
  cout << str;
}
```

rmsr: Ahhh forgot about that detail. Well, gonna redo this PR anyway so it don't matter. Thanks for being so patient.

rmsr:
Ugh, this is now misnamed and polluted with a bunch of crap. Closing in favor of a fresh PR.

PR #132, opened 2018-01-22
Losing your last fighting limb makes you pass out from pain

cizra:
…s fights end quicker

### DIFF
--- a/code/code/misc/limbs.cc
+++ b/code/code/misc/limbs.cc
@@ -1103,6 +1103,33 @@ wearSlotT pickRandomLimb(bool)
   return wearSlotT(num);
 }
 
+void TBeing::stun(spellNumT affect)
+{
+  if (task) {
+    stopTask();
+  }
+  if (spelltask) {
+    stopCast(STOP_CAST_NONE);
+  }
+
+  affectedData aff;
+  aff.type = affect;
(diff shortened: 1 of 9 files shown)


code/code/misc/limbs.cc
```diff
@@ -1103,6 +1103,33 @@ wearSlotT pickRandomLimb(bool)
...
+  aff.type = affect;
+  aff.duration = Pulse::UPDATES_PER_MUDHOUR / 3;
+  aff.bitvector = AFF_STUNNED;
+  affectTo(&aff, -1);
+}
+
+void TBeing::stunIfLimbsChoppedOff()
+{
+  if (bothArmsHurt()) {
+    act("The pain in your crushed limbs causes you to pass out.",
```
rmsr: If you decide against my disease-based code, this message should be more abstract, since it could get called after a limb is chopped off, in which case "crushed" doesn't make sense.

rmsr:
Another way we could accelerate the 'cleanup' of a crippled mob is for the extreme pain disease to reduce defense while it's active

rmsr:
Another thing I forgot to check is whether the Affliction code will prevent one from doubling up afflictions, or if we need to add handling for that.

rmsr:
are we closing this in favor of #152 ?

rmsr:
#152 is looking good, closing this in favor of

PR #139, opened 2018-01-24
doTell: convert most users to TBeing* instead of char*

rmsr:
_REMINDER: it is recommended to review my PRs per-commit rather than as a big lump, I try to make each commit reflect a "unit of progress"._

Previously doTell would always loop thru the charlist looking for the target. Also, there are a lot of different ways sneezy got a "tellable" name out of a TBeing object!

Now, just tell directly to the TBeing* dest, except for the actual "tell xxx yyy" command which does it the hard way via an overload.

Also includes a tiny cleanup of some (ranger haha) quest handling code in combat.cc.

### DIFF
--- a/code/code/misc/being.h
+++ b/code/code/misc/being.h
@@ -743,7 +743,6 @@ class TBeing : public TThing {
     double pietyGain(double);
     void goBerserk(TBeing *);
     void checkForQuestTog(TBeing *);
-    void sendCheatMessage(char *);
     void stopFighting();
     int canBeParalyzeLimbed();
     int checkIdling();
--- a/code/code/misc/being.h
+++ b/code/code/misc/being.h
@@ -597,7 +597,7 @@ class TBeing : public TThing {
     int numClasses() const;
 
     int triggerSpecial(TThing *, cmdTypeT cmd, const char *arg); 
-    int triggerSpecialOnPerson(TThing *, cmdTypeT cmd, const char *arg); 
+    int triggerSpecialOnPerson(TThing *, cmdTypeT cmd, const sstring &arg); 
     void sendCastingMessages(bool, bool, int, skillUseTypeT, int);
     void sendFinalCastingMessages(bool, bool, skillUseTypeT);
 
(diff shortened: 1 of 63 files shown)


code/code/misc/talk.cc
```diff
@@ -733,69 +737,78 @@ TBeing *findTellTarget(TBeing *me, const sstring &name, bool visible, bool mobs)
 }
 
 
-// returns DELETE_THIS on death of this
-// triggerSpecOnPerson prevents this from being constant
+TBeing *findAccountAlternate(TBeing *me, const sstring &name, bool visible)
+{
+    TBeing *vict;
+    TDatabase db(DB_SNEEZY);
+    db.query("select p1.name as name from player p1, player p2, account a where p2.name='%s' and a.account_id=p2.account_id and p1.account_id=a.account_id", name);
```
cizra: select player.name as name from player where account_id = %d and name <> %s

code/code/misc/being.h
```diff
@@ -1816,7 +1815,16 @@ class TBeing : public TThing {
     void lookingAtObj(TThing *);
     void doShout(const sstring &);
     int doWhisper(const sstring &);
+
+    int doTell(const TBeing *vict, const sstring &message, bool visible = TRUE) {
+        return doTell(const_cast<TBeing *>(vict), message, visible);
```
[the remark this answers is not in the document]
rmsr: Are you referring to the complex de-constifying, or the need to have a const->nonconst override?

code/code/misc/talk.cc
rmsr: more readable AND shorter, double win - can't eliminate the join because we don't have the account ID to hand (it's a tell to someone who isn't logged in)

rmsr: I also dedented it because I gave up on my dreams of four-space indents.

code/code/misc/being.h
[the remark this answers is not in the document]
cizra: `const_cast` sounds like inviting undefined behaviors. For instance, GCC tends to reorder code pretty aggressivesly, so you might end up first writing, later reading and finally ending up with the initial value you were supposed to overwrite. Let's do the proper thing and deconstify things that aren't const -- or let's make the method const and add `mutable` where needed.

rmsr:
bleaaah so messed up

PR #166, opened 2018-01-31
Relock & unjam doors and containers during zone reset

rmsr:
Currently this gets cycled pretty frequently, I thought about adding a tweak to reduce the rate a bit but the only thing that would work comfortably with the code right now is for it to be based on a percent chance per reset. Migrating the zonefiles to the db will require refactoring the reset code, I've left todo markers so I'll pick it up there.


code/code/task/task_picklock.cc
```diff
@@ -1,144 +1,122 @@
...
 #include "being.h"
 
-void TThing::pickPulse(TBeing *ch)
-{
-  ch->sendTo("Hey, where'd your lockpick go?!?\n\r");
-  ch->stopTask();
-}
-
-void TTool::pickPulse(TBeing *ch)
+void pick_pulse(TBeing *ch, TThing *pick)
```
cizra: namespace {}

code/code/task/task_picklock.cc
```diff
@@ -1,144 +1,122 @@
...
+  int skill = 2 * ch->getSkillValue(SKILL_PICK_LOCK)
+      + 3 * ch->task->timeLeft // longer they try, better chance
+      + ch->plotStat(STAT_CURRENT, STAT_FOC, 3, 18, 13)
+      + ::number(-15,15);
 
-  skill += ch->plotStat(STAT_CURRENT, STAT_FOC, 3, 18, 13);
-  skill += ::number(-15,15);
+  if (!tool->addToToolUses(-1)) {
+    act("Your $o snap in half!   Oops!", true, ch, tool, nullptr, TO_CHAR);
+    act("$n's $o snaps in half.", true, ch, tool, nullptr, TO_ROOM);
```
cizra: Probably should be singular?

code/code/task/task_picklock.cc
```diff
@@ -1,144 +1,122 @@
...
+    if (other &&
+        (back=other->dir_option[rev_dir[ch->task->status]]) &&
         back->to_room == ch->in_room)
       REMOVE_BIT(back->condition, EXIT_LOCKED);
     ch->stopTask();
+    return;
   }
+
+  if (!::number(0,2))
+    act("You wiggle $p in the lock, without luck. (yet...)", true, ch, tool, nullptr, TO_CHAR);
```
cizra: Me no speaking English natively, but perhaps "the lockpick"?

code/code/task/task_picklock.cc
```diff
@@ -1,144 +1,122 @@
...
   if (!ch->doesKnowSkill(SKILL_PICK_LOCK)) {
     ch->sendTo("I bet you wish you knew how to pick locks.\n\r");
     ch->stopTask();
     return FALSE;
   }
-  int pulses_to_wait;
+
   switch(cmd) {
     case CMD_TASK_CONTINUE:
+      TThing *pick;
```
cizra: It'd look nicer to initialize while creating the variable, then check value

lib/txt/news
```diff
@@ -1,3 +1,5 @@
+01-29-18 : Jammed locks don't affect doorbash.
```
cizra: The commit message about flagging the exit doesn't do justice to all the nice things happening here!

code/code/task/task_picklock.cc
rmsr: yeah, today I read something reminding me about throwing static on these kinds of things: I'm using that instead of a namespace as it's shorter. it used to be deprecated vs anon namespaces but it's back to a-ok again. added.

code/code/task/task_picklock.cc
rmsr: oh! there's format codes for this kind of thing, let me use them... except that lockpicks are plural in the strings and singular in the code - an unfortunate mismatch. I will add code to handle that. the string fixup is added first.

rmsr: added a TTool overload for isPluralItem, that checks if fname(name) ends in 's', which wraps this up. then if someone makes a singular lockpick we're covered.

rmsr: oh! I'd missed that there's a skeleton key item, with fname "key", and it'll show as singular! perfect!

code/code/task/task_picklock.cc
rmsr: I agree that with $p printing short_desc it's not strictly grammatical in this sentence, since most of the tools are plural, but I feel printing short_desc is the "sneeziest" thing to do here and consistent with much of our other act usage. With "a skeleton key" in the database, printing a specific noun here can mismatch with the held tool.

code/code/task/task_picklock.cc
rmsr: I ended up doing it this way because of weird switch scoping rules I changed it as you describe, but had to wrap the whole case body inside of brackets to do it. otherwise C++ doesn't let you declare and initialize a variable on the same line inside a switch, because the variable's visibility falls through to further cases. why the original code was actually legal is kind of baffling though, since it's functionally equivalent to the illegal code. oh well, fixed!

code/code/obj/obj_tool.cc
```diff
@@ -117,3 +117,7 @@ int TTool::objectSell(TBeing *ch, TMonster *keeper)
   return FALSE;
 }
 
+bool TTool::isPluralItem() const
+{
+  return (fname(this->name).back() == 's');
```
cizra: this-> is superfluous

PR #167, opened 2018-02-01
Fix buffer overflow in new account code

rmsr:
gcc stack protector seems to catch this, but still... Throw in a free fix for the account named NEW issue.

### DIFF
--- a/code/code/misc/extern.h
+++ b/code/code/misc/extern.h
@@ -349,7 +349,6 @@ extern int mapSpellnumToFile(spellNumT);
 extern spellNumT mapFileToSpellnum(int);
 extern bool applyTypeShouldBeSpellnum(applyTypeT);
 extern bool has_key(TBeing *ch, int key);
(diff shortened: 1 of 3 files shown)


code/code/misc/extern.h
```diff
@@ -349,7 +349,7 @@ extern int mapSpellnumToFile(spellNumT);
 extern spellNumT mapFileToSpellnum(int);
 extern bool applyTypeShouldBeSpellnum(applyTypeT);
 extern bool has_key(TBeing *ch, int key);
-extern int bogusAccountName(const char *arg);
+extern int bogusAccountName(const sstring arg);
```
cizra:
```cpp
extern bool bogusAccountName(const sstring& arg);
```

also I'd stick the extern into `client.cc` and remove from here.

rmsr: the less in extern.h the better for sure.

PR #173, opened 2018-02-02
SConstruct: init git submodules if necessary

rmsr:
Pretty simple change, but maybe there's more I should add?

### DIFF
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -1,6 +1,20 @@
 # hey emacs, this is -*- python -*- vim: ft=python
+from __future__ import print_function
+
 import sys
 import subprocess
+
+
+def info(*msg):
+    print('INFO:', *msg)
+
+def warning(*msg):
+    print('WARNING:', *msg, file=sys.stderr)
+
+def error_exit(*msg):
+    print('ERROR:', *msg, file=sys.stderr)
+    raise SystemExit(1)
+
 
 # NOTE site_scons/site_init.py runs before this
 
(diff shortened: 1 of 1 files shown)


code/SConstruct
```diff
@@ -73,6 +73,18 @@ opt_olevel =         get_option('olevel',      '')
 opt_cxx =            get_option('CXX',         'g++')
 
 
+## ensure submodules are ready to go
+submodules = subprocess.check_output(['git','submodule','status']).split()
+if any(sub.startswith('-') for sub in submodules):
+    print >>sys.stderr, 'Initializing submodule dependencies, please wait...'
```
cizra:
```py
from __future__ import print_function
import sys

print('lala', file=sys.stderr)
```

By the way, why do you consider this comment to be an error?

code/SConstruct
```diff
@@ -73,6 +73,18 @@ opt_olevel =         get_option('olevel',      '')
...
+    try:
+        subprocess.check_call(['git','submodule','init'])
+        subprocess.check_call(['git','submodule','update'])
```
cizra: `git submodule update --init`?

code/SConstruct
```diff
@@ -73,6 +73,18 @@ opt_olevel =         get_option('olevel',      '')
...
+    except submodule.CalledProcessError, e:
```
cizra:
```py
except Something as e:
```

code/SConstruct
rmsr: ooh nice flag, combined

code/SConstruct
rmsr: I don't use e anyway, removed

rmsr:
Addressed the good feedback, merging

Issue #174, opened 2018-02-03
Riposting a mob parry results in a 3rd-person act msg


rmsr:
```
Masstrizial parries your stab at her neck.
Isegrim uses his parry to execute a riposte.
```

PR #186, opened 2018-08-14
fix riposte success roll & 3rd person message

sntwo:
Issue #174, error in who rolls for riposte bSuccess

### DIFF
--- a/code/code/misc/combat.cc
+++ b/code/code/misc/combat.cc
@@ -3902,22 +3902,22 @@ int TBeing::oneHit(TBeing *vict, primaryTypeT isprimary, TThing *weapon, int mod
       if (monkDodge(vict, weapon, &dam, w_type, part_hit))
         mess_sent |= ONEHIT_MESS_DODGE;
       else if (thiefDodge(vict, weapon, &dam, w_type, part_hit))
-	mess_sent |= ONEHIT_MESS_DODGE;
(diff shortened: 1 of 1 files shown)


code/code/misc/combat.cc
```diff
@@ -3902,22 +3902,22 @@ int TBeing::oneHit(TBeing *vict, primaryTypeT isprimary, TThing *weapon, int mod
...
-	  act("$n uses $s parry to execute a riposte.",
-	      FALSE, this, 0, vict, TO_NOTVICT);
-	  // actual ripose attack is added in hit().  Klugey, but
-	  // easier than trying to create a seperate attack here.
-	  SET_BIT(vict->specials.affectedBy, AFF_RIPOSTE);
-	}
-	mess_sent |= ONEHIT_MESS_DODGE;
+	      if(vict->doesKnowSkill(SKILL_RIPOSTE) &&   // must know the skill
+	        (::number(0,99) < 50) &&                // only 50% of the time
+	        vict->bSuccess(SKILL_RIPOSTE)){
```
cizra: I agree that the victim's riposte skill should be checked, not the attacker's — however, how does that change the actor in message?

sntwo: woops, no change to message.  ammended commit

cizra: Can't see the fix to message, again

PR #187, opened 2018-11-16
Aion2501 patch 1

Aion2501:
Intended updates: Change MotD and remove an over-spawning load in Dead Forest zone.

### DIFF
--- a/lib/txt/motd
+++ b/lib/txt/motd
@@ -6,7 +6,7 @@
       Players with homes will be provided keys to their homes upon request.<z>
 
 
-     <Y>SNEEZYMUD CONTACT:<z>       <c>support@sneezymud.com<z>
+     <Y>SNEEZYMUD PATREON:<z>       <c>https://www.patreon.com/Sneezy<z>
      <Y>DISCORD CHANNEL:<z>       <c>https://discord.gg/YY2PgqP<z>
      <Y>FACEBOOK GROUP:<z>       <c>https://www.facebook.com/groups/SneezyMud/<z>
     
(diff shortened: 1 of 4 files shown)


lib/txt/motd
```diff
@@ -1,24 +1,23 @@
...
-      or events made for players will be returned. Players with homes 
-      will be provided keys to their homes.<z>
+      <R>Offical Status:<z> PWIPE 07/26/2018
+      Limit 1 Account. 2 simultaneous character logons per account. 
+      Multiclassing Re-Enabled. Access dual-classing at lvl 50.  
+      Players with homes will be provided keys to their homes upon request.<z>
 
 
      <Y>SNEEZYMUD CONTACT:<z>       <c>support@sneezymud.com<z>
+     <Y>DISCORD CHANNEL:<z>       <c>https://discord.gg/YY2PgqP<z>
```
cizra: Delete the email? Or add a working one, perhaps, but having a broken email is much worse than having no email at all.

lib/txt/motd
```diff
@@ -1,24 +1,23 @@
...
      <Y>WEBSITE:<z>        <c>Wiki Online @ wiki.sneezymud.com<z>
      <Y>NEWBIE HELP:<z>    <c>'help newbie'
      <Y>GAME HELP:<z>      <c>game commands: REQUEST, BUG, IDEA and TYPO.
 
-     <R>-------------<z>   <W>Multiplay is enabled and allowed.<z> 
-     <R> NEWS UPDATE <z>   <W>Prior issues resolved with multiplay<z> 
-     <R>             <z>   <W><z> 
-     <R>-------------<z>   <W>Please consolidate potions into containers<z>
+     <R>-------------<z>   <W>Multiplay is enabled. Limit 1 Account.<z> 
+     <R> NEWS UPDATE <z>   <W>2 simultaneous character logins allowed.<z>
```
cizra: I wonder if we should point out that the limit is changeable by, say, questing.

PR #202, opened 2018-12-05
migrate wiz data files to db

sntwo:
Migrate data files in lib/immortals to the db.

### DIFF
--- a/code/code/misc/wiz_data.cc
+++ b/code/code/misc/wiz_data.cc
@@ -16,68 +16,29 @@ extern "C" {
 #include "extern.h"
 #include "monster.h"
 #include "person.h"
-
-class wizSaveData {
-  public:
-    int setsev,
-        office,
-        blockastart,
-        blockaend,
-        blockbstart,
-        blockbend;
-
-    wizSaveData();
-};
-
-wizSaveData::wizSaveData() :
-  setsev(0),
-  office(0),
-  blockastart(0),
-  blockaend(0),
-  blockbstart(0),
-  blockbend(0)
-{}
+#include "database.h"
 
 void TBeing::wizFileRead()
 {
-  FILE *fp;
-  sstring buf, buf2;
   Descriptor *d = NULL;
-  wizSaveData saveData;
 
   // don't use isImmortal, save always
   if (!(GetMaxLevel() > MAX_MORT) || !(d = desc))
     return;
 
-  buf = format("immortals/%s/wizdata") % getName();
-  fp = fopen(buf.c_str(), "r");
-  if (!fp) {
-    buf2 = format("immortals/%s") % getName();
-    fp = fopen(buf2.c_str(), "r");
-    if (!fp) {	// no immort directory 
-      if (mkdir(buf2.c_str(), 0770)) {
-	sendTo("Unable to create a wizard directory for you.\n\r");
-	vlogf(LOG_FILE, format("Unable to create a wizard directory for %s.") %  getName());
-      } else
-	sendTo("Wizard directory created...\n\r");
-    } else
-      fclose(fp);	// player has no wizard data, but has immort directory 
+  TDatabase db(DB_SNEEZY);
 
-    return;
-  }
-  if (fread(&saveData, sizeof(saveData), 1, fp) != 1) {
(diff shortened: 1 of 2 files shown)


code/code/misc/wiz_data.cc
```diff
@@ -16,68 +16,29 @@ extern "C" {
...
 
-  d->severity    = saveData.setsev;
-  d->office      = saveData.office;
-  d->blockastart = saveData.blockastart;
-  d->blockaend   = saveData.blockaend;
-  d->blockbstart = saveData.blockbstart;
-  d->blockbend   = saveData.blockbend;
+  db.fetchRow();
+
+  d->severity     = (int) convertTo<int>(db["setsev"]);
```
cizra: Will it crash if there's no such row?

code/code/misc/wiz_data.cc
```diff
@@ -16,68 +16,29 @@ extern "C" {
...
+  d->office       = (int) convertTo<int>(db["office"]);
+  d->blockastart  = (int) convertTo<int>(db["blockastart"]);
```
cizra: `convertTo<int>` already returns `int`, so the cast is superfluous.

code/code/sys/migrations.cc
```diff
@@ -143,6 +143,55 @@ void runMigrations() {
                     "value text not null, "
                     "foreign key (player_id) references player (id) on delete cascade)"));
         },
+        [&](){
+            vlogf(LOG_MISC, "Moving wiz data over to db");
+            assert(sneezy.query(
+                    "create table if not exists wizdata ("
+                    "setsev int not null,"
+                    "office int not null,"
+                    "blockastart int not null,"
```
cizra: I can totally imagine an immortal with no office, or no assigned block. Why are they `not null`? Also, I'd add `default 0` to office.

code/code/sys/migrations.cc
```diff
@@ -143,6 +143,55 @@ void runMigrations() {
...
+                    "blockaend int not null,"
+                    "blockbstart int not null,"
+                    "blockbend int not null,"
+                    "player_id bigint(20) unsigned not null, "
+                    "primary key (player_id), "
+                    "foreign key (player_id) references player (id) on delete cascade)"));
+            assert(sneezy.query("select id, name from player"));
+            std::map<int, std::string> idToName;
+            while (sneezy.fetchRow())
```
cizra: 2 loops doing both the same thing, operating over players. I'd collapse them into one loop.

code/code/misc/wiz_data.cc
```diff
@@ -103,72 +64,39 @@ void TMonster::wizFileSave()
...
-    vlogf(LOG_BUG, format("Corrupt wiz save file for %s") % arg);
-    fclose(fp);
-    return;
-  } 
-  fclose(fp);
-  buf3 = format("The office of %s is %d.\n\r") % arg % saveData.office;
-  sendTo(buf3);
+  TDatabase db(DB_SNEEZY);
+
+  db.query("select office from wizdata where player_id = (select id from player where name = '%s')", arg.c_str());
```
cizra: My twitch reaction for this task would be
```sql
select office from wizdata w join player_id p on w.player_id = p.player_id where p.name = '%s'
```
Can't justify it, though, your way is totally fine.

code/code/misc/wiz_data.cc
```diff
@@ -103,72 +64,39 @@ void TMonster::wizFileSave()
...
+  db.fetchRow();
+  buf = format("The office of %s is %d.\n\r") % arg % (int)convertTo<int>(db["office"]);
```
cizra: Will it crash if you `office Sestrel`?

code/code/misc/wiz_data.cc
```diff
@@ -103,72 +64,39 @@ void TMonster::wizFileSave()
...
+  sendTo(buf);
```
cizra: We don't really need `buf`, you know.

code/code/misc/wiz_data.cc
cizra: Hmm, reading code, it seems it'll return 0.

PR #212, opened 2019-02-06
Adding Race: Half-Elf

Aion2501:
Slightly higher end racials, however will have to deal with height/gearing issues.


wershlak:
@Aion2501 What's up with this? was the intent here to create a new playable race? or just add to the list of races for builders to use? 

The stats look too good. Are they human sized or elf sized or both?

Aion2501:
> @Aion2501 What's up with this? was the intent here to create a new playable race? or just add to the list of races for builders to use?
> 
> The stats look too good. Are they human sized or elf sized or both?

Abandoned it because I couldn't resolve how to make it a playable race.
Also wanted to add the concept that they could end up elf size or human size depending on the roll, but unable to resolve that issue either. Sneezy elves are very short.

Over all just unfinished and needs help to implement.

drewhershey:
Closing as abandoned.

Issue #213, opened 2019-02-16
Weapon immunity rework


wershlak:
Had a discussion with Aion about immunities that from weapon attacks. 

Weapons are not itemized with +1,+2,+3 hit on them. In order to get a weapon with these stats it generally is preferable to find a non-magic weapon so it can be enhanced. Magical weapons shouldn't be a penalty. The system is kind of dumb if we don't itemize and scale the +hit up with weapon levels.

This is particularly painful for thieves since they cannot use glowing weapons to execute their sneak attacks. 

We proposed a simple fix.

- adjust the code so that non-magical weapons are subject to immune_nonmagic while magical weapons ignore it
- Monks with voplat >= 15% ignore immune_nonmagic
- change the enhance weapon spell so that it doesn't add the glow effect

This allows all weapons to be useful.

Future ideas:
- remove immune_plus1/2/3 from code and from mobs to clean up after ourselves.
- possibly convert immune_plus1 to immune_magic - with tight rules about how it is applied. No possibility of 100% immunity to magic and 100% immunity to non-magic.
- Change enhance weapon to add other benefits (stats/HP/Mana/procs).
- Allow enhance weapon to be cast on magical items, further enhancing or overwriting effects.

PR #214, opened 2019-02-16
Change weapon immunity

wershlak:
This implements changes discussed here: https://github.com/sneezymud/sneezymud/issues/213

was able to test by editing mobs to give them 100% immunity to non-magic and fighting 
- with magic items vs without and confirming damage is applied
- with voplat and without and confirming damage is applied

- confirmed that mobs with slash/blunt/pierce immunity still get applied

Also tested enhance weapon to be sure it still applies magic and +hit/dam but no longer adds light/glow

### DIFF
--- a/code/code/misc/combat.cc
+++ b/code/code/misc/combat.cc
@@ -4100,38 +4100,27 @@ int TBeing::preProcDam(spellNumT type, int dam) const
   return (dam);
 }
 
+// weaponCheck handles immunity to non-magic when reconciling damage.
+// Since magic items and skills can avoid it's affects in ways that
+// don't apply to other immunity types
+// NOTE: slash/blunt/pierce immunity is handled in preProcDam
 int TBeing::weaponCheck(TBeing *vict, TThing *o, spellNumT type, int dam)
 {
-  immuneTypeT imm_type = getTypeImmunity(type);
-  int imm_num=0, total=0;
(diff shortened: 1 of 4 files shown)


Aion2501:
I have discussed and support these changes.

code/code/misc/combat.cc
```diff
@@ -4100,38 +4100,27 @@ int TBeing::preProcDam(spellNumT type, int dam) const
   return (dam);
 }
 
+// weponCheck handles immunity to non-magic when reconciling damage.
```
cizra: typo

code/code/obj/obj_base_weapon.cc
```diff
@@ -741,9 +741,10 @@ int TBaseWeapon::enhanceMe(TBeing *caster, int level, short bKnown)
       affected[1].modifier += 1;
 
     addObjStat(ITEM_MAGIC);
-    addObjStat(ITEM_GLOW);
 
-    addGlowEffects();
+    // Remove glowy-ness from enhance
+    // addObjStat(ITEM_GLOW);
+    // addGlowEffects();
```
cizra: We have Git for history, no need to leave dead code.

wershlak:
Cleaned up typos and squashed. I don't have permission to merge this though.

PR #215, opened 2019-02-21
Silvered Flag/Bless Dmg Boost

Aion2501:
Adding silvered flag and damage boosts to silver/blessed weapons vs undead/lycanth/diabolic.


code/code/misc/combat.cc
```diff
@@ -1721,13 +1721,13 @@ int TBeing::extraDam(const TBeing *vict, const TBaseWeapon *weap) const
 
   if (vict->isUndead() || vict->isLycanthrope()) {
     if (weap->isObjStat(ITEM_BLESS))
-      plus += 1;
-    if (weap->getMaterial() == MAT_SILVER) 
-      plus += 1;
+      plus += 10;
+    if(weap->getMaterial() == MAT_SILVER || weap->isObjStat(ITEM_SILVERED))
```
wershlak: ITEM_SILVERED is not defined. The definitions are here. Looks like there's a couple unused though

https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/obj.h#L481

Aion2501: Ahh perfect! Added from here.

Aion2501: Will have to go through a testing period to see if its too strong, but I think bless and silver vs undead demons and lycanth should be felt significantly. The right tool for the right job.  

Bless is only available if you have cleric/deik in your class list. I dont think holy water works on weapons, though it should. And Silvered will be difficult to come by as I have it planned atm.

PR #216, opened 2019-02-25
Adding Silvered flag constant

Aion2501:


wershlak:
Looks like this can be merged.

Able to oedit and add silvered flag to eq. Shows up in stat and seems applied

👍

PR #223, opened 2019-06-03
Update periodic.cc

parkscs:
Tweaking the pyrophobia trait to scale inversely with level (0% chance at level 50, close to 100% chance of current proc rate at level 1).  My thought is simply if we have level 50 characters that will go into battle with an arch vampire, they shouldn't flee from a lamp post. :)

### DIFF
--- a/code/code/misc/periodic.cc
+++ b/code/code/misc/periodic.cc
@@ -1305,7 +1305,7 @@ int TBeing::updateHalfTickStuff()
   }
 
   // player is scared of flame (exempt or lessen for mobs?)
-  if(hasQuestBit(TOG_HAS_PYROPHOBIA) && (getPosition() > POSITION_SLEEPING) && !::number(0,1)){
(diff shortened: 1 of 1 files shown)


parkscs:
I ran it by Aion before submitting, but not sure about more of a consensus
than that.  My personal thought for going forward is to move away from fear
(which is a shit gameplay mechanic) and perhaps have them suffer a debuff
from fire damage that would temporarily reduce some stats and shut off
their regen.  In a game where I can suicide my character against a dragon
at level 1, I don't really see a good reason to take away ability to
control my character when facing lamp posts or even basic fire spells.  I'd
personally like to see a shaken debuff that reduces some stats, shuts off
the racial regen and maybe lasts an hour or two when fire damage is taken.
This still gives the race a downside against fire (and they already take
more damage from fire), but leaves fear out of it (at least at high
levels).

As for why not number(0,50)-level... my brain thinks in percentages, lol.
That's a better way to implement it. :)

cizra:
I don't buy that argument — but I don't care too much either, so in it goes.

PR #225, opened 2019-06-12
New weapon procs - molten + possessed

parkscs:
Proposing a couple of thematic weapon procs for weapons I'd like to see in the game.  One is aimed at a possessed spear with damage procs based around a phantasmal warrior - chance to cause bleed and rarest one has a chance to embed an item similar to bloodspike (todo is to create a thematic item, but using spike for now).

The molten one is aimed at a fiery 2h great axe, and is largely limited to 2h weapons (most abilities don't work if the weapon is not PAIRED).  Some damage procs and a bonebreak proc that is similar to wither limb for thematic reasons.  The damage procs include some healing balanced around this being limited to 2h weapons.  There are a few thematic additions too like an enhanced damage proc if the enemy is bleeding (and it consumes the bleed), enhanced damage against undead, and so on.


parkscs:
Let me tweak as discussed.

PR #226, opened 2019-06-12
[WIP] implement first two auras

wershlak:
DO NOT MERGE

Looking to get feedback on this early. There's still quite a bit to do here but I want to make sure the general idea and the hooks into the scheduler and combat doesn't have any hidden gotchas that I didn't consider. I've mostly used the object special procs as a guide on how this should work as I'm attempting to do similar things here with auras.


wershlak:
This is out of date. closing

PR #227, opened 2019-06-13
Removing "exceptions" to dispel magic for items

parkscs:
There are some legacy exceptions to dispel magic that I don't think make a lot of sense in the current state of the game.  This change will effectively allow all stats to be dispelled from items (previously, stats that affected moves, light, and so on were immune to dispel).  Consequently, this will allow enhance weapon to apply to all weapons.

### DIFF
--- a/code/code/disc/disc_alchemy.cc
+++ b/code/code/disc/disc_alchemy.cc
@@ -1329,7 +1329,8 @@ int dispelMagic(TBeing *caster, TObj * obj, int, short bKnown)
 
     // No point affecting objects already set to none, but legacy checks removed.
     for (i = 0; i < MAX_OBJ_AFFECT; i++) { 
-      if ((obj->affected[i].location != APPLY_NONE)) {
+      if ((obj->affected[i].location != APPLY_NONE) &&
+	  (obj->affected[i].location != APPLY_ARMOR))) {
         obj->affected[i].location = APPLY_NONE;
         obj->affected[i].modifier = 0;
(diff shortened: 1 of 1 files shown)


Aion2501:
I approve. Dispel should be able to remove any AFFECT from an object.

wershlak:
I verified this compiles and I can dispel magic on items. 
I'm not sure what the goal is with this change though. Can someone explain the problem this is meant to fix?

I dispelled magic on a shield that was magic. It removed the "magic" flag as well as all stats including the armor provided by the shield. Seems like a way for people to ruin their gear accidentally.

parkscs:
Let me look at the armor - that may be a problem.  The goal with the change
was to prevent some wonky behavior on dispelling weapons.  The code would
say it worked, but it would never dispel - and you can't enhance weapons
with stats on them.  So you can't dispel the stats, but you can't enhance
the weapon, and the text from dispel magic never gives you any indication
the spell failed (and in fact it didn't fail, it was just coded to ignore
some stats).

wershlak:
I looked at enhance real quick and it looks like it fails if there are any effects on the weapon. So like if something is +5dex but not "magic" it fails. The message in that case should be: "That item is too powerful for this spell to work on it."

Dispel removing those affects seems like a fair trade if you know what you're getting into. Except Armor but there aren't may weapons with armor on them. I think it makes sense to put the exception to armor back in.

Should dispel remove glow and hum flags as well as the magic flag? Those annoy me.

PR #229, opened 2019-06-30
Preventing weapon dislodge into stunned targets

parkscs:
Currently pierce weapons very frequently get stuck into stunned targets, and this is a bit silly (and annoying).  I think a v2.0 of this change would be to add skill checks to weapon retention/strength/telekinesis/etc. so that you can dislodge the weapon yourself (with bonus damage) from non-stunned targets, but this is a quick and easy quality of life change for now.

### DIFF
--- a/code/code/misc/combat.cc
+++ b/code/code/misc/combat.cc
@@ -5311,7 +5311,11 @@ int TBeing::dislodgeWeapon(TBeing *v, TThing *weapon, wearSlotT part)
 {           
   char buf[160];
   int rc;
-
+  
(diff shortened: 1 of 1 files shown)


code/code/misc/combat.cc
```diff
@@ -5311,7 +5311,11 @@ int TBeing::dislodgeWeapon(TBeing *v, TThing *weapon, wearSlotT part)
 {           
   char buf[160];
   int rc;
-
+  
+  // Quality of life tweak - there's no sense in weapons getting stuck in opponents that can't even move
+  if (v->isAffected(AFF_STUNNED))
```
cizra: Could use a flavor text

PR #232, opened 2019-07-13
New weapon procs: glacial and molten

parkscs:
Molten:
   - 1d8 fire damage
   - 2% chance of bonebreak (fire-themed with wither effect)
   - fire-themed flavor text

Glacial:
   - 1d8 frost damage
   - 0.4% chance to freeze opponent (stun for 1 second)
   - ice-themed flavor text

### DIFF
--- a/code/code/spec/spec_objs.h
+++ b/code/code/spec/spec_objs.h
@@ -37,7 +37,7 @@ const int SPEC_FACTIONSCORE_BOARD = 97;
 const int SPEC_GRAFFITI=135;
 const int SPEC_STOCK_BOARD  = 138;
 const int SPEC_BRICKQUEST = 148;
-const int NUM_OBJ_SPECIALS = 161;
+const int NUM_OBJ_SPECIALS = 163;
 
 struct TObjSpecs {
   bool assignable;
--- a/code/code/spec/spec_objs_weapons.cc
+++ b/code/code/spec/spec_objs_weapons.cc
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
   return TRUE;
 }
 
+// Summary of this weapon proc
+//    - 1d8 fire damage proc
+//    - 2% chance of bonebreak, with a fire theme (withers the limb due to burn damage)
+//    - Flavor text proc of dripping lava from the weapon
+int moltenWeapon(TBeing *vict, cmdTypeT cmd, const char *arg, TObj *o, TObj *)
+{
+  char limb[80];
+  int rc, dam = 0, chance;
+  TBeing *ch;
+  sstring buf;
+  wearSlotT slot;
+
+
+  if (!(ch = dynamic_cast<TBeing *>(o->equippedBy)))
+    return FALSE;       // weapon not equipped (carried or on ground)
+
+
+  // Flavor text pulse
+  if (cmd == CMD_GENERIC_PULSE)
+  { 
+    if (!::number(0,75)){
+      act("<r>Drops of molten hot lava fall from $n's $o onto the $g.<1>",
+          0, ch, o, 0, TO_ROOM);
+      act("<r>Your $o drips molten hot lava onto the $g.<1>",
+          0, ch, o, 0, TO_CHAR);
+    }
+  }
+
+
+  if(cmd == CMD_OBJ_HITTING)
+  {
+    chance = ::number(1,50);
+
+
+    if (!o || !vict)
+      return FALSE;
+
+
+    // Bonebreak proc + 10-20 damage if successful (max once/fight)
+    if (chance == 1)
+    {
+       if(!ch->canWither(vict, SILENT_YES))
+         return false;
+
+
+       bool ok = false;
+       bool found = false;
+
+
+
+
+       // Initial checks to prevent issues and for balancee reasons
+       for (slot = MIN_WEAR; slot < MAX_WEAR; slot++) 
+       {
+         if (notBreakSlot(slot, false))  // same ones, right?
+           continue;
+         if (!vict->slotChance(slot))
+           continue;
+         found |= (vict->isLimbFlags(slot, PART_MISSING));
+         ok = true;
+        }
+
+
+
+
+       if (found || !ok) 
+	 return false;
+
+
+       if (vict->isImmune(IMMUNE_HEAT, slot) )
+         return false;
+
+
+
+
+
+
+       for (slot = pickRandomLimb();; slot = pickRandomLimb())  
+       {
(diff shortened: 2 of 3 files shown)


code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+         if (notBreakSlot(slot, true))
+           continue;
+         if (!vict->slotChance(slot))
+           continue;
+         break;
+       }
+       sprintf(limb, "%s", vict->describeBodySlot(slot).c_str());
+
+
+       TThing *t;
```
cizra: Move down to where you actually use it

code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+
+
+       if (!vict->hasPart(slot)) {
+         vlogf(LOG_COMBAT, format("BOGUS SLOT trying to be made PART_USELESS: %d on %s") % 
+       	   slot % vict->getName());
+      	      return FALSE;
+       }
+
+
+       if (!vict->roomp) {
```
cizra: Really feels like 90% of this code should be shared with wither limb prayer

code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+
+
+       if (IS_SET_DELETE(rc, DELETE_VICT) || (vict->getHit() < -10))
+         delete vict;
+
+
+       return true; 
+     }
+     // Damage proc - 1d8 fire damage
+     else if(chance >= 2 && chance <= 10)
```
cizra: By now everybody's forgotten what was the `if` for this `else`. Split up

code/code/spec/spec_objs.h
```diff
@@ -37,7 +37,7 @@ const int SPEC_FACTIONSCORE_BOARD = 97;
 const int SPEC_GRAFFITI=135;
 const int SPEC_STOCK_BOARD  = 138;
 const int SPEC_BRICKQUEST = 148;
-const int NUM_OBJ_SPECIALS = 161;
+const int NUM_OBJ_SPECIALS = 163;
```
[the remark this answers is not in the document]
parkscs: Not sure.  I don't know why they chose to use constants, and it's a bit of a pain as it seems to fully recompile everything when these values are changed.  If we can find a more elegant way to manage these, I definitely think that would be good.  But I think it's outside the scope of this branch and adding a couple new weapon procs.

code/code/spec/spec_objs_weapons.cc
parkscs: A lot of this was originally borrowed from the unmaker proc (which removes limbs altogether), but I moved away from that after discussing with Aion.  Our concern was that missing limbs cannot be healed by players and that can cause some problems from a game balance perspective.  What I would propose is that I come back in a separate branch and rework unmaker/molten and perhaps wither limb as well to pull some of this out into a separate method.  That would let us continue to have different flavor text for different procs/spells without duplicating all the checks, and would allow us to more easily add new procs/spells in the future.

code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+
+
+    rc = vict->reconcileDamage(vict, dam, DAMAGE_FROST);
+    if (IS_SET_DELETE(rc, DELETE_VICT) || (vict->getHit() < -10))
+      delete vict;
+
+
+    return true;
+    }
+  else if (roll == 0)
```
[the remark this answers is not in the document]
parkscs: Should be fixed.  I need to find a better editor for working on this stuff; copy pasting from a putty window to the web interface is causing some formatting issues.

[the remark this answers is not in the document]
cizra: Well, the usual way is to clone the repo and hack away locally, in your own computer, then use Git (not web) to get the code back to GitHub.

code/code/spec/spec_objs_weapons.cc
cizra: Or just pass in the flavor texts as a function parameter, packed into a struct or such.

code/code/spec/spec_objs_weapons.cc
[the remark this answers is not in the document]
parkscs: I'll pick your brain next time you're around on the best way to setup my dev environment. :)

code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+       if (found || !ok) 
+	 return false;
+
+
+       if (vict->isImmune(IMMUNE_HEAT, slot) )
+         return false;
+
+
+
+
```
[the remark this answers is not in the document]
parkscs: Removed - they were caused by copy/pasting into the web interface.

code/code/spec/spec_objs_weapons.cc
```diff
@@ -2524,3 +2524,278 @@ int livingVines(TBeing *vict, cmdTypeT cmd, const char *, TObj *o, TObj *)
...
+       for (slot = pickRandomLimb();; slot = pickRandomLimb())  
+       {
+         if (notBreakSlot(slot, true))
+           continue;
+         if (!vict->slotChance(slot))
+           continue;
+         break;
+       }
+       sprintf(limb, "%s", vict->describeBodySlot(slot).c_str());
```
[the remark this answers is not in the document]
parkscs: No particular reason, other than it was borrowed from a diff. weapon proc.

code/code/spec/spec_objs_weapons.cc
parkscs: Done.

code/code/spec/spec_objs_weapons.cc
parkscs: It'd be doable to pass them in.  Generally there's 3 sets of flavor text - to the victim, to the user and to the rest of the room.  Each one can get various parameters, such as the specific limb being hit, although that would likely be determined in the shared function so that'd be fine.  With that said though, I'd want to take a deeper look at these procs to try and put some more thought into it to try and figure out what's the best approach.

PR #241, opened 2019-12-05
Fix how immunity is applied

wershlak:
Currently immunity is buggy when dealing with values above 100 or below -100. 

This PR updates TBeing::addToImmunity to so that the accurate value is stored in immune_arr 

TBeing::getImmunity will still min/max between 100 -100.

### DIFF
--- a/code/code/misc/immunity.cc
+++ b/code/code/misc/immunity.cc
@@ -178,7 +178,7 @@ void TBeing::setImmunity(immuneTypeT type, short amt)
 
 void TBeing::addToImmunity(immuneTypeT type, short amt)
 {
-  immunities.immune_arr[type] = min(max(immunities.immune_arr[type] + amt, -100), 100);
+  immunities.immune_arr[type] = immunities.immune_arr[type] + amt;
 }
 
 bool TBeing::isImmune(immuneTypeT bit, wearSlotT pos, int modifier) const
--- a/code/code/misc/being.h
+++ b/code/code/misc/being.h
@@ -141,7 +141,7 @@ class specialData {
 
 class immunityData {
   public:
-    byte immune_arr[MAX_IMMUNES];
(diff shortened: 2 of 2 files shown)


code/code/misc/immunity.cc
```diff
@@ -178,7 +178,7 @@ void TBeing::setImmunity(immuneTypeT type, short amt)
 
 void TBeing::addToImmunity(immuneTypeT type, short amt)
 {
-  immunities.immune_arr[type] = min(max(immunities.immune_arr[type] + amt, -100), 100);
+  immunities.immune_arr[type] = immunities.immune_arr[type] + amt;
```
cizra: Immunity should still be capped to [-128, 127] range because it's stored in a single byte.
Instead of reintroducing the bug, though, we could skip applying the immunity, or something.

I also cannot find where immunities are saved during renting. Can you find it? Is it in `st` or in DB or just recalculated on login? We should at least have a look if it's compatible with the fix.

wershlak: Ahh. I didn't look at the immune_arr. I just assumed since "amt" was short that it was an array of short. 

In my test I had a couple items totaling 120. In master this is enough to see the problem but with my fix everything works fine. I tested again with items totaling 130 and saw it broken still. 

I think there are really only two solutions here.
1) Store an accurate number in the array and have getImmunity() return a min/maxed value
2) Store a min/maxed value in the array but recalculate that value entirely from scratch each time an affect changes

I don't think this is stored in the database. Once the immunity value is bugged I was able to fix it by logging out and logging back in.

I think all affects are calculated from scratch at login (and some other times). See TBeing::affectTotal()
https://github.com/sneezymud/sneezymud/blob/master/code/code/sys/handler.cc#L502

I've pushed a new commit that updates immune_arr to be an array of short and this seems to work with larger values.

PR #255, opened 2020-03-26
multiclass hitpoint fix

wershlak:
- Remove advanced defense from monks and rebalance their hitpoints. Overall I think this makes leveling a monk better and I'd like to add a shield spec to advanced def later that makes sense for warrior and deikhan but not really for monk. 

- Fixes the logic for classHpPerLevel() to take multiclass into account. 
--

I tested this a bit:

- Monks without pracs spent in advanced defense see more hitpoints when they log in. They can no longer see the disc when they type prac. No fiddling needed.

- Monk who had already spent pracs in advanced defense have even more hitpoints. They probably should be fixed manually. Requires a god to use \@set

- Multiclass hitpoints seem to adjust themselves with no intervention. The adjustment is generally small since the spread of hp per level of different classes is small. After the fix to monks obviously.


code/code/misc/limits.cc
```diff
@@ -96,20 +96,22 @@ int baseHp()
...
     }
   }
 
-  if(!hpgain){
+  if(!hpgain || num_classes == 0){
     vlogf(LOG_BUG, format("No class in classHpPerLevel() for %s") %  tp->getName());
     hpgain=7.0;
   } 
 
+  hpgain = hpgain / num_classes;
```
cizra: You're casting it back to int, thus losing precision. Instead, let's do this:
```cpp
return hpgain / num_classes;
```

cizra:
@Aion2501 please review

code/code/misc/limits.cc
wershlak: I updated it. Looks nicer this way too. Thanks!

wershlak:
@Aion2501 
a \@set disc char_name 65 0
and  \@set skill char_name 419 0

would fix a monk that had pracs in advanced def. I don't think it's a problem if a couple single class monks have extra hitpoints though.

PR #256, opened 2020-04-02
Multiclass fixes

wershlak:
Went through a lot of hasClass() to see where it was misused to not be multiclass safe. 

           - Evaluating magic items should use the best skill
           - Shaman/Mage combinations can now use the spells command
           - Report should now show all relevant resources for multiclass
           - "list fit" at shops should now work correctly
           - Fixed a typo in a previous news entry!
           - Fixed score output to show all relevant resources
           - Monks who wield weapons now have a speed modifier to number of 
             attacks like all other classes
           - Monk/Warrior multiclass now gets extra attacks when berserking

### DIFF
--- a/code/code/cmd/cmd_score.cc
+++ b/code/code/cmd/cmd_score.cc
@@ -15,29 +15,22 @@ void TBeing::doScore()
 {
   time_info_data playing_time;
   sstring Buf, tString;
-  char part[256];
 
-  Buf = "";
-  sprintf(part, "You have %s%d%s/%s%d%s hit points, ", red(), getHit(), norm(), green(), hitLimit(), norm());  
-  Buf.append(part);
+  Buf = format("You have %s%d%s/%s%d%s hit points, ") % red() % getHit() % norm() % green() % hitLimit() % norm();
 
   if (hasClass(CLASS_DEIKHAN) || hasClass(CLASS_CLERIC)) {
-    sprintf(part, "%s%.2f%s %spiety, ", cyan(), getPiety(), "%", norm());
-    Buf.append(part);
+    Buf+=format("%s%.2f%s %spiety, ") % cyan() % getPiety() % "%" % norm();
   }
 
   if (hasClass(CLASS_SHAMAN)) {
-    sprintf(part, "%s%d %slifeforce, ", red(), getLifeforce(), norm());
-    Buf.append(part);
+    Buf+=format("%s%d %slifeforce, ") % red() % getLifeforce() % norm();
   }
 
   if (hasClass(CLASS_MAGE) || hasClass(CLASS_MONK) || hasQuestBit(TOG_PSIONICIST)) {
-    sprintf(part, "%s%d%s/%s%d%s mana, ", orange(),  getMana(), norm(), green(), manaLimit(), norm());
-    Buf.append(part);
+    Buf+=format("%s%d%s/%s%d%s mana, ") % orange() %  getMana() % norm() % green() % manaLimit() % norm();
   }
 
-  sprintf(part, "and %s%d%s/%s%d%s moves.\n\r", purple(), getMove(), norm(), green(), moveLimit(), norm());
-  Buf.append(part);
+  Buf+=format("and %s%d%s/%s%d%s moves.\n\r") % purple() % getMove() % norm() % green() % moveLimit()% norm();
   sendTo(Buf);
 
   sendTo(format("You are %s.\n\r") % DescMoves((((double) getMove()) / ((double) moveLimit()))));
--- a/code/code/cmd/cmd_score.cc
+++ b/code/code/cmd/cmd_score.cc
@@ -15,24 +15,32 @@ void TBeing::doScore()
 {
   time_info_data playing_time;
   sstring Buf, tString;
+  char part[256];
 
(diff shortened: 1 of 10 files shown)


code/code/misc/shop.cc
```diff
@@ -1852,21 +1852,21 @@ void shopping_list(sstring argument, TBeing *ch, TMonster *keeper, int shop_nr)
...
+    if(ch->hasClass(CLASS_WARRIOR) && !(extra_flags & ITEM_ANTI_WARRIOR))
+      fit=true;
+    if(ch->hasClass(CLASS_THIEF) && !(extra_flags & ITEM_ANTI_THIEF))
+      fit=true;
+    if(ch->hasClass(CLASS_SHAMAN) && !(extra_flags & ITEM_ANTI_SHAMAN))
+      fit=true;
+    if(ch->hasClass(CLASS_DEIKHAN) && !(extra_flags & ITEM_ANTI_DEIKHAN))
+      fit=true;
+    if(ch->hasClass(CLASS_MONK) && !(extra_flags & ITEM_ANTI_MONK))
+      fit=true;
```
cizra, line 1869: It'd be more honest if fit = none(anti_bits), instead of !all(anti_bits). It'd be a major nerf for multiclassers, though.
@Aion2501 opinions?

Aion2501: Multi-classes could use a kick in the pants for pracs. 
Depending on the class, they're getting an extra 1-3 pracs per level for the first 30 levels. Single classes have to max basic 200 points before their prac cap is lifted. Usually something like 7 pracs then 9 starting around Lvl 30.
(because Peel ripped out the INT factor and added a flat modifier)

Multi-classes don't have that restriction in the first place (!bug) and even if they did, they can max the first 200 much, much faster. So they're dripping with extra pracs. About 90 I'd say, on average.

Aion2501: This would make it so Mage/Cleric/Warrior couldn't use a piece of armor that was anti-cleric for example?

HUGE downvote from me on that one if that's the case. We're not equipped to deal with that change, because none (or very few) of the objects in the game were designed with that in mind. Tri-classes would go naked in the streets. 

The alternative idea is to implement the additional separation of objects by type. Heavy Armor, Medium Armor, and Light Armor are sub-types already written in and abandoned. 
Heavy - War/Deik
Medium - Cleric/Ranger
Light - Thief/Mage
Cloth - Monk/Shaman

You'd have to force these flags upon everything in the game based on the AC level the object has, because there are way too many objects for me to do it by hand. Or some other ingenious method. Same reason the anti-flags would be so troublesome. But if you accomplish that, then we can add the armor req to the class kits and force multiclassers to use the lowest kit in their setup.

code/code/cmd/cmd_score.cc
```diff
@@ -15,24 +15,32 @@ void TBeing::doScore()
...
   time_info_data playing_time;
   sstring Buf, tString;
+  char part[256];
 
-  sendTo(format("You have %s%d%s/%s%d%s hit points, %s%d%s/%s%d%s moves and ") %	 red() % getHit() % norm() %
-	 green() % hitLimit() % norm() %
-	 purple() % getMove() % norm() %
-	 green() % moveLimit() % norm());
+  Buf = "";
+  sprintf(part, "You have %s%d%s/%s%d%s hit points, ", red(), getHit(), norm(), green(), hitLimit(), norm());
```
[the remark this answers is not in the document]
wershlak: There's a lot of places in the code where sprintf() is used. I don't do c++ and looking at the sneezy code is my only real guide. I lean on your reviews. I'll look to change it.

code/code/misc/shop.cc
wershlak: The change here is to which items are listed in a shop when you type "list fit". It in NO way affects what a multiclass can wear. It's a defect that typing "list fit" doesn't currently display items that a character can wear.
--
If we want to change what multiclasses can wear that's a balance discussion that we should start somewhere else.

Aion2501: Phew. Ok. Approve than. Don't mind me. You know I can only read about 10% of this code.

code/code/cmd/cmd_score.cc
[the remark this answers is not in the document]
cizra: To be honest, I'm not a terribly big fan of Boost either, but there just doesn't seem to be a good way of assembling strings in C++...

code/code/misc/other.cc
```diff
@@ -586,22 +587,22 @@ void TBeing::doReport(const char *argument)
...
-  else
-    sprintf(info, "$n reports '%s%.1f%% H, %.1f%% M. I am %s%s'", 
-           red(), getPercHit(), getPercMana(), 
-           DescMoves((((double) getMove()) / ((double) moveLimit()))), 
-           norm());
 
+  sprintf(info, "%.1f%% H, ", getPercHit());
+  if (hasClass(CLASS_CLERIC) || hasClass(CLASS_DEIKHAN)) {
+    sprintf(inter, "%.2f%% P, ", getPiety());
+    strcat(info, inter);
```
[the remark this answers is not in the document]
wershlak: Converted doReport to use format()

code/code/cmd/cmd_score.cc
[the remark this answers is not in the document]
wershlak: I converted this over to format() and learned some shit in the process. Thanks.

PR #265, opened 2020-05-02
Add skill to advanced defense

wershlak:
adding a skill to advanced defense for warriors and deikhans. This is very similar in concept to counter move for monks. Lets the tank classes have a chance to avoid some debilitating special moves.


code/code/cmd/cmd_bonebreak.cc
```diff
@@ -239,7 +244,15 @@ int bonebreak(TBeing *caster, TBeing *victim)
...
       rc = bonebreakMiss(caster, victim, 1);
       if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
-	return rc;
+        return rc;
+      return TRUE;
+    }
+
+    if (victim->canFocusedAvoidance(bKnown/2)) {
+      SV(SKILL_BONEBREAK);
+      rc = bonebreakMiss(caster, victim, 2);
```
cizra, line 253: Instead of a function call (or perhaps inside that function call), I'd just have an array of string triples. Look up the strings and print them — less code lines and no branch mispredictions in CPU.

cizra: Not saying you should go and refactor everything in the world, though :)

code/code/disc/disc_defense.cc
```diff
@@ -6,16 +6,20 @@
...
 
 #include "disc_defense.h"
+#include "being.h"
+
 
 CDDefense::CDDefense() :
   CDiscipline(),
-  skAdvancedDefense()
+  skAdvancedDefense(),
+  skFocusedAvoidance()
```
cizra, line 15: Chances are that these constructors, destructors, operator= don't do anything useful and should be deleted.

code/code/disc/disc_defense.cc
```diff
@@ -24,9 +28,36 @@ CDDefense & CDDefense::operator=(const CDDefense &a)
...
+  if (!awake() || getPosition() < POSITION_CRAWLING)
+    return FALSE;
+
+  int skill = getSkillValue(SKILL_FOCUSED_AVOIDANCE);
+
+  // Agi seems to be the defense stat so this makes sense
+  skill *= getAgiMod();
+
+  if (eitherLegHurt())
+    skill = (int) (skill * 0.75);
```
cizra, line 54: The cast doesn't do anything here, it's already implicit in the =

code/code/misc/being.h
```diff
@@ -1404,6 +1404,7 @@ class TBeing : public TThing {
   bool checkSmashed(TBeing *, wearSlotT, spellNumT, TThing *, int, const char * = NULL);
   int hit(TBeing *, int pulse = -1);
   bool canCounterMove(int);
+  bool canFocusedAvoidance(int);
```
cizra, line 1407: I wonder why these skills live in TBeing? This causes horrible recompilation times for everything, as well as leaking inner workings of disciplines into generic code of beings.

As most things in TBeing are public, maybe it makes more sense to have that code in disciplines?

code/code/misc/skills.cc
```diff
@@ -1182,6 +1182,8 @@ CSkill *TBeing::getSkill(spellNumT skill) const
 // disc_defense
     case SKILL_ADVANCED_DEFENSE: // 674
       return &((CDDefense *) cd)->skAdvancedDefense;
+    case SKILL_FOCUSED_AVOIDANCE:
+      return &((CDDefense *) cd)->skFocusedAvoidance;
```
cizra, line 1186: Ugh :( I don't know offhand how to make it good, but this thing smells horribly.

cizra: I'm not requiring you to rewrite this, though.

code/code/cmd/cmd_bonebreak.cc
```diff
@@ -70,6 +70,11 @@ int bonebreakMiss(TBeing *c, TBeing *v, int type)
 	act("$n tries grapple you into a bone breaking hold but you avoid it.", 
 	    FALSE, c, 0, v, TO_VICT);    
       } break;
+      case 2:{
```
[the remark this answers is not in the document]
wershlak: Updated this to use an enum. Also it does look cleaner to me without the brackets.

code/code/disc/disc_defense.cc
wershlak: removed this

code/code/misc/skills.cc
wershlak: All the code bits related to skills and disciplines seem a bit ugly. It's a bit beyond my c++ ability to find a clean way to refactor it. It also seems like a big project without much benefit to users.

code/code/misc/being.h
wershlak: Seems like all the combat type code is dealing with two objects that are subclasses of TBeing of some sort. So it kind of makes sense to be able to write 
```
if (victim->canFocusedAvoidance(bKnown/2)) {
```
I guess the other option would just be

```
if (canFocusedAvoidance(victim, bKnown/2)) {
```
Just not object oriented. It seems a little easier to reason about in the first instance but if the performance impact is huge the other way it's not a big deal to change it all.

I feel like "Disciplines" or "Abilities" maybe should be a class that is instantiated per being.

Issue #158, opened 2018-01-30
Allow those with lockpicking skill to relock containers using a lockpick


rmsr:
Skill check, jam on crit fail.

PR #268, opened 2020-05-04
Adding Set/Cleaning Data

Aion2501:
Adding set #158 Flowing Black
Cleaning up the order of objects to match how they are displayed in game.
Adding a more visible start line for future additions.

### DIFF
--- a/lib/objdata/suitsets
+++ b/lib/objdata/suitsets
@@ -33,18 +33,18 @@
 // Paste below this line.
 // ----------------------
 :Set #158
-:#Flowing Black
-:$Human
-:head     = 2500
-:neck     = 2501
-:back     = 2503
-:arm      = 2504
-:hand     = 2505
-:finger   = 2506
-:waist    = 2507
-:leg      = 2508
-:foot     = 2509
-:shield   = 2511
+#flowing black
+$human
+head     = 2500
+neck     = 2501
+back     = 2503
+arm      = 2504
+hand     = 2505
+finger   = 2506
+waist    = 2507
+leg      = 2508
+foot     = 2509
+shield   = 2511
 :Set #157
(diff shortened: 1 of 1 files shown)


lib/objdata/suitsets
```diff
@@ -6,31 +6,45 @@
...
 :leg      = 
 :foot     = 
-:finger   = 
 :shield   = 
-:back     = 
 // Paste below this line.
+// ----------------------
+:Set #158
+:#Flowing Black
+:$Human
```
cizra, line 37: Others use lowercase here. I don't know if it matters, but uniformity would be good too

lib/objdata/suitsets
```diff
@@ -6,31 +6,45 @@
...
-:wrist    = 
 :leg      = 
 :foot     = 
-:finger   = 
 :shield   = 
-:back     = 
 // Paste below this line.
+// ----------------------
+:Set #158
+:#Flowing Black
```
cizra, line 36: These `:` should be removed — see the other suitsets below. Currently, it's commented out.

wershlak:
I'm wondering what happens if someone checks out the code and tries to load this set? I assume it will just spam me with "There is no such object." a bunch?

Can we get a recent sql_data dump committed that includes this set?

cizra:
We can do a new DB dump, that's not a big problem. I don't know any elegant ways to version DB states, though...

Aion2501:
Ah yes I see the comments now. What a nub. And you're right that they're all lowercase in the 'load set' command. I think they should have all been uppercase, but that's lots of work. Uniformity it is!

Lots of suitsets have missing slots. I've seen it return the entire list with a few lines that say "there is no object in this slot" or something to let the immortal know.

PR #269, opened 2020-05-04
fix crash in recalculating attacks

cizra:


wershlak:
Seems like you fixed it but I'll go through this code again and make sure this is right for all cases. Was kind of an annoying change. Sorry about the bug :)

cizra:
No problem. What _really_ would have helped is unittest coverage.

PR #270, opened 2020-05-04
Update Sanct

Aion2501:
Increased level of sanct proc to 50.
Also increased the duration from about 2 mud minutes to 3 mud hours, as if cast by a player.


wershlak:
Not sure I agree with this. TBH I didn't like the change to the surcoat proc either. 

It's hard to balance when you give an important class ability away. If sanc is basically up all the time for free then why be a cleric? If it's up too little then what's the point of putting it on there?

With multiclass it's actually annoying to have it proc cause it is lower level. You've fixed that here but I'd much rather come up with a new ability and switch it out so that it doesn't interfere with sanc. 

Then we don't have to worry about what percentage of uptime is fair to clerics or annoying to the user.

Off the top of my head it could be an immunity to some type of magic? Fire or drain maybe? Even immunity to non-magic would be pretty awesome (and do basically what this does) but harder to balance.

Aion2501:
I don't believe in half-measures. This proc actually causes conflict when getting sanct from other sources; such as a cleric, platinum potion, or object. Additionally, if you manage to get the same spell affected twice simultaneously, they cancel each other out when one expires and you don't have the buff at all, even if it says you do. Not to mention a 2 mud minute duration? Let's be serious.

I'm in favor of removing it entirely as a defensive measure and going on the offensive. It's a holy sword of devastating after all. Perhaps a random short blessing that gives +hitroll or +dmg to demons or something.

Either way. My fix or cut it out entirely. The status quo has to go.

wershlak:
I agree. I think changing it is a better fix but this fix is obviously easier. I'm happy to think up something else befitting a deikhan sword and code it after I update the all the code I have in the pipeline. That said, if you want to go ahead and merge this now with the goal of a different fix later I am cool with that.

code/code/spec/spec_objs_deikhan_sword.cc
```diff
@@ -76,11 +76,11 @@ void doBlind(TBeing *ch, TBeing *vict, TObj *o) {
 
 void doSanc(TBeing *ch, TObj *o) {
   affectedData aff;
-  int level=30;
+  int level=50;
 
   aff.type = SPELL_SANCTUARY;
   aff.level = level;
-  aff.duration = Pulse::UPDATES_PER_MUDHOUR/4;
+  aff.duration = Pulse::ONE_SECOND * 60 * 8;
```
wershlak, line 83: Is this different than the 3 * Pulse::UPDATES_PER_MUDHOUR for sanc?

Aion2501: It's amateur hour over here when I try to fix code. That's probably the normal way and surely much easier to figure. I just grabbed from the list of timers the code has listed and calc'd out real time to mud time the hard way. Backwards now, I realize.

Aion2501: Pulse::UPDATES_PER_MUDHOUR * 3

Is that how you'd do it? Still not quite sure how you'd say 3 mud hours.

wershlak: `aff.duration = 3 * Pulse::UPDATES_PER_MUDHOUR;`
is what the sanc code for mortals works out to.

PR #273, opened 2020-06-20
Restore skill difficulty

wershlak:
Here's an interesting story. So I was chatting about how people feel skills used to do more damage and decided to walk through how this gets applied and I found this code that had been short circuited 19 years ago. 

getSkillDiffModifier in misc/skill_dam.cc returns a number based on the difficulty of the skill. This is primarily used in two places. 
- in determining success or failure of the skill/spell
- in determining the damage of the skill/spell

It was short circuited to just return 100 as if all skills are a trivial difficulty skill. Never fail unless the char is pretty unfocused.

The result damage wise is that most skills (normal difficulty) will do 25% more damage if this is merged. Up to "dangerous skills" that can do up to 66% more damage. Of course they will fail much more frequently too.

The reason listed to make the change was that "too much complaining about skill failure, so screw it  - bat 2/3/00".

I think we should turn merge this. I think it makes the game more interesting. We can tune the difficulty of an individual skill or global adjust skill success rates if we decide things need to be tuned.

I tested that the change compiles, that using skills/spells still worked and that with damage debug turned on I observe the expected change in damage values. I obviously didn't test how annoying increased failure rates are over the long haul. Also, there could be some abilities that need tweaking after this.

If we want to test this more somehow or if there's a specific concern I am happy to spend more time with it.

---
NOTES

Here's everywhere this function is used in the code
```
// Damage modifier in skill Damage
// Due to the change we are always multiplying dam by 1
// Some dangerous skills would get multiplied by 1.66667
./code/code//misc/skill_dam.cc:  fixed_amt *= (100.0 / getSkillDiffModifier(skill));


// bSuccess MAIN skill check for success. Everything runs through here and
// difficult skills will fail much more often if we make this change
./code/code//misc/discipline.cc:  float limit = getSkillDiffModifier(spell);


// Some skills have longer lockout imposed on the victim based on the difficulty of the skill
// This seems dumb as the difficulty of a skill never changes. Perhaps it was meant to be the
// learnedness of the skill? This has always been irrelevant since it always returned 100.
// I'm going to just remove this so as not to change the expected behavior of these skills. 
./code/code//cmd/cmd_spin.cc:  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_SPIN));
./code/code//cmd/cmd_bodyslam.cc:  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BODYSLAM));
./code/code//cmd/cmd_trip.cc:  wait = (wait * 100.0 / getSkillDiffModifier(SKILL_TRIP));
./code/code//cmd/cmd_bash.cc:  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BASH));
./code/code//disc/disc_air.cc:  rounds = (rounds * 100.0 / getSkillDiffModifier(SPELL_IMMOBILIZE));
./code/code//disc/disc_psionics.cc:    wt = (wt * 100.0 / getSkillDiffModifier(SKILL_KINETIC_WAVE));


// Dodge/parry/block work better based on skill difficulty. These were added since this code has been broken. I will remove so we don't have some unexpected results.
./code/code//disc/disc_monk.cc:  int amt = (int) (100 * 100 / getSkillDiffModifier(SKILL_JIRIN));
./code/code//disc/disc_thief.cc:  int amt = (int) (45 * 100 / getSkillDiffModifier(SKILL_DODGE_THIEF));
./code/code//disc/disc_dueling.cc:    amt += (int)(420 * 100 / getSkillDiffModifier(SKILL_TRANCE_OF_BLADES));
./code/code//disc/disc_dueling.cc:    amt += (int)(80 * 100 / getSkillDiffModifier(SKILL_PARRY_WARRIOR));
./code/code//disc/disc_dueling.cc:    amt += (int) (40 * 100 / getSkillDiffModifier(SKILL_PARRY_WARRIOR));
./code/code//disc/disc_dueling.cc:  int amt = (int) (45 * 100 / getSkillDiffModifier(SKILL_PARRY_WARRIOR));
```

### DIFF
--- a/code/code/cmd/cmd_bash.cc
+++ b/code/code/cmd/cmd_bash.cc
@@ -323,8 +323,6 @@ int TBeing::bashSuccess(TBeing *victim, spellNumT skill)
   // the effect here should be strictly to prevent skill-use
   // it should NOT cause loss of attacks, or do damage
   float wt = combatRound(discArray[SKILL_BASH]->lag);
-  // adjust based on bash difficulty
-  // wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BASH));
 
   // Take a few moves from basher
   // since we cost some moves to perform, allow an extra 1/2 round of lag
--- a/code/code/cmd/cmd_bash.cc
+++ b/code/code/cmd/cmd_bash.cc
@@ -324,7 +324,7 @@ int TBeing::bashSuccess(TBeing *victim, spellNumT skill)
   // it should NOT cause loss of attacks, or do damage
   float wt = combatRound(discArray[SKILL_BASH]->lag);
   // adjust based on bash difficulty
-  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BASH));
+  // wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BASH));
 
   // Take a few moves from basher
   // since we cost some moves to perform, allow an extra 1/2 round of lag
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -194,8 +194,6 @@ static int bodyslamHit(TBeing *caster, TBeing *victim)
 
   // see the balance notes for details on what's going on here.
   float wt = combatRound(discArray[SKILL_BODYSLAM]->lag);
-  // adjust based on bash difficulty
-  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BODYSLAM));
 
   // since we cost some moves to perform, allow an extra 1/2 round of lag
   wt += 1.5;
--- a/code/code/cmd/cmd_spin.cc
+++ b/code/code/cmd/cmd_spin.cc
@@ -180,9 +180,6 @@ static int spinHit(TBeing *caster, TBeing *victim)
   // see the balance notes for details on what's going on here.
   float wt = (combatRound(discArray[SKILL_SPIN]->lag) / 3);
 
-  // adjust based on bash difficulty
-  wt = (wt * 100.0 / getSkillDiffModifier(SKILL_SPIN));
-
   // since we cost some moves to perform, allow an extra 1/2 round of lag
   wt += 1.5;
 
--- a/code/code/cmd/cmd_trip.cc
+++ b/code/code/cmd/cmd_trip.cc
@@ -250,7 +250,6 @@ int TBeing::tripSuccess(TBeing *victim, spellNumT skill)
     return DELETE_THIS;
 
   float wait = combatRound(discArray[SKILL_TRIP]->lag);
-  wait = (wait * 100.0 / getSkillDiffModifier(SKILL_TRIP));
   addToMove(-5);
   wait += 0.5;
   // round up
--- a/code/code/disc/disc_air.cc
+++ b/code/code/disc/disc_air.cc
@@ -196,8 +196,7 @@ int immobilize(TBeing * caster, TBeing * victim, int level, short bKnown)
   // some command lock-out.
   // since spell is tasked, add one to the lag
   float rounds = 0.2233 * (discArray[SPELL_IMMOBILIZE]->lag + 1);
-  // adjust for difficulty
-  rounds = (rounds * 100.0 / getSkillDiffModifier(SPELL_IMMOBILIZE));
+
   // adjust for saving-throw penalty
   rounds = (rounds * 4 / 3);
 
--- a/code/code/disc/disc_dueling.cc
+++ b/code/code/disc/disc_dueling.cc
@@ -226,8 +226,8 @@ int TBeing::parryWarrior(TBeing *v, TThing *weapon, int *dam, int w_type, wearSl
   // the higher amt is, the more things get blocked
   int amt = 0;
   if (trance) {
-    amt += (int)(420 * 100 / getSkillDiffModifier(SKILL_TRANCE_OF_BLADES));
-    amt += (int)(80 * 100 / getSkillDiffModifier(SKILL_PARRY_WARRIOR));
+    // 50% base parry chance
+    amt += 500;
     switch(::number(0,2)){
     case 0:
       strcpy(type, "parry");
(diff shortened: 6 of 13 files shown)


wershlak:
```
Overall I think it is a better experience for players to feel like they are succeeding.
```
I agree. In general I don't think we want failures on maxed spells/skills to be the norm but we've basically taken the idea of skill difficulty out of the game. I think that's actually really sucky too.  I think chars should just have to be built around the idea of being a good caster if they want to use difficult skills. If they are then they should be mostly successful. 

Perhaps I've overstated the failures.

here's the snippet that explains what the expectation is:
```
  // Here's the basis of this stuff:
  // At max learning, we desire the following results:
  // trivial    = 100%
  // easy       = 90%
  // normal     = 80%
  // difficult  = 70%
  // dangerous  = 60%
  // for less than maxed learning, scale it up linearly
  // Have focus factor in: high focus = 1.25 * above rates
  // low focus = 0.80 * above rates
```
So focus factors in and can be significant. Most of my chars have high focus cause I assumed based on the helpfiles that focus was important. With Max focus a char would skill succeed at normal tasks 100% of the time. 

There's only a handful of dangerous tasks in the game and someone with high focus should succeed 75% of the time.

Without adding this back in _some_ form Focus should be a dump stat. Anything higher than 105 is a total waste.

I like that if you want to be a successful "casting" shaman you'd need good focus. Today there aren't any casting shaman cause spell damage is sucky. I think this is the first step to restoring that.

---

If you think the above failure rates are too high. I propose we talk about adjusting them rather than scrapping the whole idea of skill difficulty.

Two ideas there.
1) We adjust the rates. bumping them all up 10%
```
  // trivial    = 110%
  // easy       = 100%
  // normal     = 90%
  // difficult  = 80%
  // dangerous  = 70%
```
above 100% makes sense once you factor in Focus. You'd have to be really dumb to fail.

Meanwhile a focused char will always succeed at even difficult skills and even the most dangerous succeed at 87.5%

OR
2) We add Karma as a factor into skill success. We keep the same formulas but factor Karma in to be half as important as focus.

someone with max Focus and Karma will succeed on difficult tasks 98% of the time and dangerous tasks 84% of the time.

Aion2501:
Are those success values calculating mob level or AC or anything else that could lower the final value? 
People already complain about not being able to hit high level mobs. I just didn't want to add to that burden since most people already feel like using skills isn't very important. We want them to use the skills and feel like their skill kit has a real impact.

I agree that having a chance to fail is important, otherwise we're trivializing success and that's not exciting either. 

I think we could improve upon this by adding a chance to crit and to return a crit success message to the player, rather than the 10% hit bonus. Now they can miss and they can crit.

wershlak:
RE: Crits
Most spells have a crit chance. Unfortunately, it's coded individually in each spell. Any extra damage or cool messages are coded per spell. 

Here's slingshot for instance. It is double damage if you crit. 
https://github.com/sneezymud/sneezymud/blob/master/code/code/disc/disc_earth.cc#L26-L59

Someday I'd like to rework this a bit to use int/wis to modify crit chance. I have it as a separate item. It requires touching every spell individually if i want to adjust messages and crit damage.

So, long story short... There is a crit mechanic and it could use some love but, doing so is a separate task. 

---

```
People already complain about not being able to hit high level mobs. I just didn't want to add to that burden since most people already feel like using skills isn't very important. We want them to use the skills and feel like their skill kit has a real impact.
```

So the feedback I got was that using skills isn't important because they don't seem to do that much damage. This change would be an overall increase in damage and make people want to use skills/spells more. I'd like to make this change and then evaluate how things go before making additional changes to further tweak spells/skills. Overall I would like a pure caster to be a viable char at some point. I think we don't have enough diversity in character styles at the moment.

Giving spells more damage while requiring a bigger investment in mental stats would be a good step there. Next steps are maybe:
- add int or wis to modify damage
- add int or wis to modify crit

At that point we might be mostly to the point where a gnome caster char isn't a joke.
--

I think I like the second option better now that I think about it. Add karma as a smaller component of success here.

Aion2501:
Spells have crit. I was thinking about basic melee fighter abilities. Bash, kick, stab, headbutt, kneestrike, etc. 

The player could get a generic crit return message to blanket all of these, such as:
Whoa! They really felt that one..

Though I'm not sure if that's realistic code-wise.

Karma playing a small role sounds good. If the players complain, we can adjust it.

Aion2501:
Is there another mechanic that replaced this? Because even with max skill/focus it seems really easy to miss prayers and spells.

wershlak:
Players fail spells because they get distracted. The idea was that players should have to decide if they want to be a hitter or a caster. At least at a particular time they shouldn't be able to do both. This is when the engage mechanic was added.

Being engaged and casting currently never fails but if you hit you get distracted a lot. Currently nobody would choose to engage and cast because it doesn't do nearly enough damage.

This code gets us part of the way there... where a char could potentially build to be a caster type char. With all the high level EQ out there I think we'll need so additional adjustments but I prefer to do things one step at a time.
- First restore damage and difficulty
- add int/wis damage bonus
- Take a look at crits
- Then finally, if needed we can just boost damage for skills

wershlak:
@cizra I updated this to add focus as Aion and I discussed. It should be mergable if you're happy with it.

cizra:
Hey, there's 1 outstanding comment still standing out. Can we fix that,
then I'll merge.

wershlak:
What's the outstanding comment @cizra? I thought I addressed everything?

cizra:
https://github.com/sneezymud/sneezymud/pull/273/files

These 2:

  // wt = (wt * 100.0 / getSkillDiffModifier(SKILL_BASH));
   let's remove dead code


int amt = 100;
Does this obsolete Jirin?

wershlak:
```
int amt = (int) (100 * 100 / getSkillDiffModifier(SKILL_JIRIN));
```
It doesn't obsolete Jirin. I changed it to ensure the skill didn't behave differently than it always has since getSkillDiffModifier() will now return a different value.

The new code sets amount to 100 (10% dodge chance). What the code above would do is modify the dodge chance based on the "difficulty" of the skill. The difficulty is constant for the skill so I'm not sure why we'd need to look that up.

PR #276, opened 2020-07-07
Add inevitability for hitters

wershlak:
This is the offensive stacking proc buff. Every time you miss you get a +1 to hit. On easy fights it makes little difference. on higher level fights, where you miss a lot, it can stack up pretty high.

It is removed on a successful hit though so at most it gives you an extra hit here and there against high level mobs.


wershlak:
I like the way the mechanic works and the power level of it. having +10hit up more often would be kind of overpowering I think.

PR #277, opened 2020-07-07
tweak syno so as to not result in death on fade

wershlak:
remove APPLY_CURRENT_HIT from syno so that when it fades it doesn't possibly result in death. Adding regen to the buff to sort of make up for it.

### DIFF
--- a/code/code/sys/socket.cc
+++ b/code/code/sys/socket.cc
@@ -1258,8 +1258,8 @@ bool procCharRegen::run(const TPulse &pl, TBeing *tmp_ch) const
   if (tmp_ch->roomp && !tmp_ch->roomp->isRoomFlag(ROOM_NO_HEAL) && 
       tmp_ch->getHit() < tmp_ch->hitLimit() &&
       tmp_ch->getCond(FULL) && tmp_ch->getCond(THIRST) &&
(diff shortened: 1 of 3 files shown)


code/code/sys/socket.cc
```diff
@@ -1256,10 +1256,10 @@ bool procCharRegen::run(const TPulse &pl, TBeing *tmp_ch) const
 {
   // this was in hit(), makes more sense here I think
   if (tmp_ch->roomp && !tmp_ch->roomp->isRoomFlag(ROOM_NO_HEAL) && 
-      tmp_ch->getMyRace()->hasTalent(TALENT_FAST_REGEN) &&
       tmp_ch->getHit() < tmp_ch->hitLimit() &&
       tmp_ch->getCond(FULL) && tmp_ch->getCond(THIRST) &&
-      !::number(0, 10)){
+      !::number(0, 10) && (tmp_ch->getMyRace()->hasTalent(TALENT_FAST_REGEN) || tmp_ch->affectedBySpell(SPELL_SYNOSTODWEOMER)))
+  {
```
cizra, line 1262: indent

wershlak: Updated

PR #281, opened 2020-07-16
Aion2501 enchanting materials

Aion2501:
Attempting to create a new Item Type for "Enchanting Materials".


Aion2501:
Putting this concept on ice for a few.

PR #282, opened 2020-07-19
tweak difficulties

wershlak:
Make everything easier across the board and then tweak individual skill/spells


code/code/misc/skill_dam.cc
```diff
@@ -14,25 +14,25 @@ double getSkillDiffModifier(spellNumT skill)
   int amt = 0;
   switch (discArray[skill]->task) {
     case TASK_TRIVIAL:
-      amt = 100;
+      amt = 110;
```
cizra, line 17: This change affects hard tasks unproportionally more — for easy task, it adds 10%. For impossible tasks it adds 40%. Is that what we want?

PR #286, opened 2020-07-27
Doubled size of promptBuf from 256 to 512. Users with overlong prompts

Hordeking:
were causing the mud to crash. This seemed to happen with mudlet, but
not the connectificator.

This change is only a band-aid, and does not address the root cause of
the buffer overflow. This cause should be looked into.


Hordeking:
I'm actually slightly more concerned about the rest of the line of code (taken from the originial code, doubled for the fix)
> char promptbuf[256] = "\0\0\0", tString[256];
To my eye, this looks like we're overflowing the buffer by 3 characters on initialization by trying to initialize with three nulls _then_ a 256 string.
Unless someone cares to explain to me why this _isn't_ bad code.

wershlak:
```
  char promptbuf[256] = "\0\0\0", tString[256];
```
This thing?

it's the same as this:
```
char promptbuf[256],  tString[256];
promptbuf = "\0\0\0";
```

I'd argue the second one is more readable but maybe because I don't work in C++ a lot and it seems weird to my eyes.

Hordeking:
Hmmm, I misread it. Good point. I generally don't declare multiple following initializers like that.

PR #290, opened 2020-08-03
Mutable dir for holding mutable data

dhtnseri:


code/code/cmd/cmd_message.cc
```diff
@@ -538,7 +538,7 @@ void TMessages::initialize()
         tChar;
   FILE *tFile = NULL;
 
-  sprintf(tString, "player/%c/%s.strings",
+  sprintf(tString, "../mutable/player/%c/%s.strings",
```
wershlak, line 541: I assume the migration path is:

- Stop Sneezy
- Copy /home/sneezy/sneezymud-docker/sneezymud/lib /home/sneezy/sneezymud-docker/mutable/
- Start Sneezy 


Would it be better to make the location of this directory a configuration parameter in sneezy.cfg? data_directory or something?

code/code/cmd/cmd_zonefile.cc
```diff
@@ -524,20 +438,14 @@ void TBeing::doZonefile(const sstring & tStArg)
...
-    sendTo("zonefile modify <field> SEE HELP FILE\n\r");
-#endif
   }
 
   std::string cmd = tStArg.word(0);
   std::string rest = tStArg.substr(tStArg.find(' ') + 1);
 
   if (is_abbrev(cmd, "save")) {
     doSaveZoneFile(this, rest);
-  } else if (is_abbrev(cmd, "load") && getName() == "Lapsos") {
```
wershlak, line 539: Why would someone have done that in the first place?

code/code/cmd/cmd_message.cc
dhtnseri: Yeah, that should work fine for migration — for new setup, you wouldn't even need to copy. Sneezy ought to create required dirs automatically.

Nope — it'd be better to migrate all this into DB.

code/code/cmd/cmd_zonefile.cc
dhtnseri: Perhaps they didn't have a test instance?

code/code/cmd/cmd_message.cc
wershlak: Yeah I assumed this was a "get it working so we can put off migrating it to the DB at least one more day" kind of effort.

I'm fine with the work you've done here as long as I can get my test container to work without losing all my pfiles.

PR #292, opened 2020-08-04
Updating website from .com to .org. Also updated a few version references to 5.4

Hordeking:
This is just a set of general housekeeping changes. Little updates here and there, as per a request by Aion to update things to say 5.4 and so on.


dhtnseri:
It's considered good style to not lump unrelated changes into one commit. Here I can see 3 different kinds of changes.

PR #293, opened 2020-08-04
Fix warnings on G++9

dhtnseri:


code/code/disc/disc_thief.cc
```diff
@@ -466,7 +466,7 @@ void TBeing::doTrack(const char *argument)
   TBeing *scan;
   affectedData aff, *Vaff;
   TThing *t = NULL;
-  char buf[256]="\0\0\0", buf2[256]="\0\0\0";
+  char buf[256]="\0\0\0", buf2[512]="\0\0\0";
```
Hordeking, line 469: Recommend putting each of these declarations on their own line. It's more readable.

dhtnseri: I'm doing here a minimum-effort job of getting it to run on newer compilers (and perhaps squash a crash bug or six).
For beautification, it should be rewritten to modern C++ style, methinks.

PR #294, opened 2020-08-04
Cosmetic changes

Hordeking:
- Amending sneezymud.com to sneezymud.org.
- Amending version number from 5.2 to 5.4 as per Aion's request.

Note: Most of these references were in urls intended for the sneezymud client to access sound files and so on. None of these appear to be available on the .com site, and the .com site itself appears to be down. I do not expect these changes should break anything, as the files were previously inaccessible anyway.


wershlak:
There's nothing wrong or potentially dangerous here but I'm not sure there's much of a point to update it since most of this doesn't work anyway (feedback, support emails, sounds).

PR #298, opened 2020-08-24
Deikhan Changes

wershlak:
Deikhan changes!

@dhtnseri Sorry this is a lot to review but I appreciate your feedback here.

```
            - Renamed all advanced discs so they don't conflict with cleric
            - Removed cleric wrath spells (rain brimstone etc.)
            - Moved around some skills to match up with the new discs
            - Divine Rescue: Advanced form of rescue that heals the target
            - 2h specialization: Additional damage with 2h weapons
            - Divine Grace: Resistence to some afflictions
            - Guardian's light: Protection proc while using a shield
            - Removed kick
            - Smite is reworked: less damage but usable more often
              Includes a +hit buff -protection debuff so it isn't advised
              to use it while tanking
            - Changed deikhan sword sanc proc to Holy Wrath (+hit buff)
              This makes it no longer conflict with spell sanctuary
            - Deikhan Auras : Buffs that affect a deikhans entire group when in
              the same room. See help aura
```


wershlak:
The history here was messy. I fixed it up and squashed this all into a single commit

cizra (approved):

PR #299, opened 2020-09-10
Keepalive

cizra:
Enabled TCP keepalive. Also changed configuration to tone down item damage rate.


Aion2501 (approved):
Keep alives good stuff.

I know the item damage 100 didn't work. 
There should be some left in if possible with whatever new fix you discover.
Just a fraction of whatever it is now. 1/5th or something.  

The players have been getting away with virtually zero armor damage for a long time now.

PR #300, opened 2020-09-14
Skill/spell stat modifiers

wershlak:
```
09-15-20 : Stat updates
            - Spells and skills now have a modification stat that affects 
              damage and duration of buffs/debuffs. See helpfiles
```

This is rebased off the deikhan changes since they both make a lot of changes to code/code/misc/spell_info.cc and the conflicts would be a pain.

Let's get that merged and I'll clean this up if needed.


Aion2501 (approved):

wershlak:
@cizra I think you perhaps squashed and merged the deikhan changes? The commit in master was different than the commit in this branch and caused the conflict somehow. I rebased and updated these three PRs.

cizra:
I rebased.

lib/txt/news
```diff
@@ -1,3 +1,7 @@
+09-15-20 : Stat updates
+            - Spells and skills now have a modification stat that affects 
+              damage and duration of buffs/debuffs. See helpfiles
```
cizra, line 3: Are the helpfiles updated somewhere else?

wershlak: In code/code/cmd/cmd_help.cc I added "modifier stat" and it pulls from the discarray. 

Looks like this:

Screen Shot 2020-09-15 at 11 15 42 AM

Or for skills where we used STAT_EXT it will say "none"

Screen Shot 2020-09-15 at 11 21 45 AM

wershlak: I think it would be better if it told you exactly how the modifier stat affects the skill/spell but I think for most skill it should be pretty obvious.

cizra (approved):

PR #307, opened 2020-11-06
change spell crit

wershlak:
Here we add Wis and Karma into the spell crit formula as part of the effort to expand the affect for lesser used stats. After this change mages will have a use for int, wis and focus. Yay!

### DIFF
--- a/code/code/misc/discipline.cc
+++ b/code/code/misc/discipline.cc
@@ -81,6 +81,7 @@
 #include "spelltask.h"
 #include "obj_symbol.h"
 #include "disc_commoner.h"
+#include "stats.h"
 
 #define DISC_DEBUG  0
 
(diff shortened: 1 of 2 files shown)


code/code/misc/discipline.cc
```diff
@@ -1359,8 +1360,91 @@ byte defaultProficiency(byte uLearned, byte uStart, byte uLearn)
...
+// Wisdom:      Primary
+// Karma:       Secondary
+// Learnedness: Modifier
+// Position:    Modifier
+//
+critSuccT spellCritSuccess(TBeing *caster, spellNumT spell)
+{
+  // Base chance is 100
+  // If we are totally average then this will hold up
+  int chance = 100;
```
cizra, line 1379: As there's a considerable number of floating-point calculates involved, perhaps we'd want to keep this variable a float as well?

cizra (changes requested):

code/code/misc/discipline.cc
wershlak: You right. I changed it.

cizra (approved):

PR #313, opened 2021-01-06
List account upgrade

wershlak:
add race and classes to account List command

--

I reworked this to be a static method that accepts the class code so it can be usable in the account menu list function.

```
-> l
The following characters are in the test account.
--------------------------------------------------
 1) bleric           human      [ M/C   Lev 50 ] Wed Jan  6 22:08:01 2021
 2) bully            bullywug   [ Warr  Lev  1 ] Thu Jan  7 00:11:35 2021
 3) deik             human      [ Deik  Lev 50 ] Tue Jan  5 23:30:11 2021
 4) elfguy           elf        [ C/S/D Lev 50 ] Tue Jan  5 21:29:48 2021
 5) fishy            kalysian   [ Warr  Lev  1 ] Mon Dec 21 01:37:44 2020
 6) magcler          kalysian   [ M/T/D Lev  1 ] Tue Jan  5 23:49:46 2021
 7) orcguy           orc        [ Warr  Lev  1 ] Sat Dec 12 22:11:08 2020
 8) phil             human      [ S/K   Lev 50 ] Thu Dec 31 06:13:00 2020
 9) reallyreallylong troglodyte [ Warr  Lev  1 ] Wed Jan  6 01:11:35 2021
10) reyek            human      [ C/T/S Lev 50 ] Sun Dec 20 19:46:33 2020
11) smartorc         orc        [ Warr  Lev  1 ] Thu Jan  7 00:55:07 2021
12) testguy          aarakocra  [ M/C/K Lev  1 ] Wed Jan  6 01:29:22 2021
13) thief            human      [ Thief Lev 50 ] Tue Dec  8 23:11:52 2020
14) trollguy         troll      [ Warr  Lev  1 ] Mon Aug 10 07:04:22 2020
15) warlike          orc        [ Warr  Lev  1 ] Wed Jan  6 00:29:39 2021

[Press return to continue]
```

who -l still works the same

```
> who -l
Players: (Add -? for online help)
--------
Aion        Level:[ Implementor  ][creator]
Bleric      Level:[ M/C   Lev 50 ]
Smartorc    Level:[ Warr  Lev  1 ]

Total players / Link dead [3/0] ( 0%)
Number Listed: 3  Max since Reboot [3]  Avg Players : [1.0]
```

### DIFF
--- a/code/code/misc/player_data.cc
+++ b/code/code/misc/player_data.cc
@@ -2298,7 +2298,7 @@ int listAccount(sstring name, sstring &buf)
     char * tmstr = (char *) asctime(localtime(&ct));
     *(tmstr + strlen(tmstr) - 1) = '\0';
     
-    buf += format("%d) %s (L%d) %s\n\r") % ++count % chars[iChar].c_str() % int(max_level) % tmstr;
+    buf += format("%2d) %-16s %-10s [ %-5s Lev %2d ] %s\n\r") % ++count % chars[iChar].c_str() % Races[int(st.race)]->getSingularName() % TBeing::getProfAbbrevName(st.Class) % int(max_level) % tmstr;
   }
   return count;
 }
--- a/code/code/cmd/cmd_who.cc
+++ b/code/code/cmd/cmd_who.cc
@@ -116,8 +116,7 @@ static const sstring getWhoLevel(const TBeing *ch, TBeing *p)
     if(p->isPlayerAction(PLR_ANONYMOUS) && !ch->isImmortal()){
       tmpstring = "Anonymous";
     } else {
-      tempbuf=format("%-5s Lev %2d") % 
-	p->getProfAbbrevName() % p->GetMaxLevel();
+      tempbuf=format("%-5s Lev %2d") % TBeing::getProfAbbrevName(p->player.Class) % p->GetMaxLevel();
       tmpstring += tempbuf;
     }
 
--- a/code/code/misc/being.h
+++ b/code/code/misc/being.h
@@ -476,7 +476,7 @@ class TBeing : public TThing {
   unsigned short GetMaxLevel() const;
   void setMaxLevel(unsigned short num);
   sstring const getProfName() const;
-  std::string getProfAbbrevName() const;
+  static sstring getProfAbbrevName(unsigned short code);
   void deityIgnore(silentTypeT = SILENT_NO) const;
   void nothingHappens(silentTypeT = SILENT_NO) const;
   float percModifier() const;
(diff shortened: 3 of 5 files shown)


code/code/misc/player_data.cc
```diff
@@ -2277,6 +2277,39 @@ std::vector<sstring> listAccountCharacters(sstring name)
   return list;
 }
 
+int count_set_bit(int n) {
+   int count = 0;
+   while(n != 0) {
+      if((n & 1) == 1) {
```
cizra, line 2283: `count += n & 1`
.. saves a branch

code/code/misc/player_data.cc
```diff
@@ -2277,6 +2277,39 @@ std::vector<sstring> listAccountCharacters(sstring name)
...
+         count++;
+      }
+      n = n >> 1; //right shift 1 bit
```
cizra, line 2286: `n >>= 1`

Also, it's very rarely worth commenting what you do (it's redundant with the code, and might get out of sync, confusing everybody).
Instead, comments (if any) should say _why_ you do what you do.

code/code/misc/player_data.cc
```diff
@@ -2277,6 +2277,39 @@ std::vector<sstring> listAccountCharacters(sstring name)
...
+   }
+   return count;
+}
+
+sstring parseClass(unsigned short code){
```
cizra, line 2291: Don't we have something similar in the `who` command?

cizra (changes requested):

code/code/misc/player_data.cc
wershlak: Updated this function. Thanks for the review. 

Yeah I use comments as a scratchpad while I working through things and forget to delete them. That was a redundant one.

code/code/misc/player_data.cc
wershlak: We do yes: https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/multiclass.cc#L350

Unfortunately it's a member of TBeing and we don't have any being objects here only codes that were read from the charfile for the account.

cizra: Since you didn't put it into a `::namespace` (nor marked as `static`, perhaps we could then call this code from `TBeing::getProfAbbrevName()`? It feels creepy to have the same thing twice.

wershlak: I don't like it either. Let me take a look and see what we can do.

cizra (approved):

PR #318, opened 2021-02-25
fix crash when a mob spec kills the next mob in character_list

cizra:


code/code/sys/process.cc
```diff
@@ -184,9 +183,7 @@ void TScheduler::runChar(int pulseNum)
 
   count=max((int)((float)mobCount/11.5), 1);
 
-  for (; tmp_ch; tmp_ch = temp) {
-    temp = tmp_ch->next;  // just for safety
-
+  while (tmp_ch) {
```
wershlak, line 186: I see tmp_ch is a member of the TScheduler class but I don't understand why that would be required. In this function it seems preferable to iterate over the entire list each time rather than start in the middle if tmp_ch is pointing to something other than null. Any idea why it's done this way rather than declaring a new locally scoped var pointing to character_list and iterating over the entire list each time?

code/code/sys/process.cc
```diff
@@ -195,22 +192,23 @@ void TScheduler::runChar(int pulseNum)
...
 
     if (tmp_ch->getPosition() == POSITION_DEAD) {
-      assert(tmp_ch); // I suppose this guy exists if we have already dereferenced it
       // even if shortDescr is NULL.
       vlogf(LOG_BUG, format("Error: dead creature (%s at %d) in character_list, removing.") % 
           tmp_ch->getName() % tmp_ch->in_room);
-      delete tmp_ch;
-      tmp_ch = temp;
+      TBeing* dead = tmp_ch;
+      tmp_ch = tmp_ch->next;
```
wershlak, line 207: So we move the tmp_ch pointer to next but don't we need to move the previous items next pointer to remove it from the list? It's been a long time since I thought about linked lists maybe there's some magic here that I don't understand.

wershlak: Screen Shot 2021-02-25 at 9 41 02 PM
I found a pic online in case my last comment didn't make sense.

code/code/sys/process.cc
cizra: We iterate over 1/11th of the list (see the count-- and break).
If we were to suddenly start iterating over the whole list,
1. CPU consumption would go up
2. lag spikes might appear
3. mob specs would trigger 11x more often.

code/code/sys/process.cc
cizra: In the old code, there was a pointer to current (`tmp_ch`) and a backup pointer to the next one, in case we need to delete the current one. That was the bug - in case the current proc deletes the next mob in the list, the next pointer would point to a deleted mob.

In the new code, we just have one pointer to current, `tmp_ch`, and deal with deletions as they come (advance through the list, stop processing this one).

code/code/sys/process.cc
wershlak: I missed the count logic. So we want to do part of the list and save where we are for next time. Got it.

code/code/sys/process.cc
wershlak: But don't we need to maintain the linked list by updating the "next" pointer of the previous item in the list?

cizra: As weird as it sounds - no, as the destructor of TBeing does that.

wershlak: Ahh ok. That makes sense. So it should be safe unless we point to next *before* we delete next as was the case previously.

wershlak (approved):

PR #320, opened 2021-02-26
Reduce Mana Costs Chain/Blur

Aion2501:
Reducing mana cost of Chain Attack from 10>5
Reducing mana cost of Blur from 25>15


code/code/misc/combat.cc
```diff
@@ -2190,7 +2190,7 @@ int TBeing::hit(TBeing *target, int pulse)
      !heldInPrimHand() && !heldInSecHand() &&
      doesKnowSkill(SKILL_BLUR) && 
      ::number(0, 99) < 20 &&  // this makes it happen 20% of the time
-     !isAffected(AFF_ENGAGER) && getMana()>=25 &&
+     !isAffected(AFF_ENGAGER) && getMana()>=15 &&
```
wershlak, line 2193: This is just a check like "do have have enough many to pay for this if we do it?"
You'll need to update the reconcileMana() call as well in order for this to work. It might break things this way.

code/code/misc/combat.cc
```diff
@@ -3976,7 +3976,7 @@ int TBeing::oneHit(TBeing *vict, primaryTypeT isprimary, TThing *weapon, int mod
      if(doesKnowSkill(SKILL_CHAIN_ATTACK) &&   // must know the skill
 	 !weapon &&                             // must be barehanded
 	 (::number(0,99) < 10) &&                // only 10% of the time
-	 getMana()>=10 &&                       // requires 10 mana
+	 getMana()>=5 &&                       // requires 5 mana
```
wershlak, line 3979: I haven't used blur much but I feel like the chain attack mana cost is fine. I was planning to allow chain attack to work with weapons as part of my "make multi classing more fun" changes. If you want to update the cost though you'll need to update the reconcileMana() call below to match this.

wershlak (changes requested):

code/code/misc/combat.cc
Aion2501: Oops. Thanks!

Monks tend to blow through mana REALLY fast. 
Blur doubles their attacks in a given round, which for many is 4-5 x 2. Whereas chain attack just gives 1 more hit. The mana effectiveness is not balanced there imo.

I also feel that in prolonged combat, or even in plow situations, monks run out of mana entirely too quickly and essentially lose their spam powers in the first minute or so. If this is too OP we can roll it back.

If you plan on making weapons work for monks I suggest we should finish the 'Martial Weapon' object type code and create proper martial arts weapons to work this way.

Aion2501 (commented):
Changes made.

wershlak:
This looks good now. Did you test it or need me to spin this up when I have time?

Aion2501:
> This looks good now. Did you test it or need me to spin this up when I have time?

Whenever you have time. ty

Durhamite2118:
Did this ever get pulled?

cizra:
@Durhamite2118 no, as @wershlak is holding it

wershlak:
I just tested this out and I'm merging this in now.

PR #325, opened 2021-03-16
Add travels

wershlak:
- Fixed logic on portals decrementing use counter and fixed it so mounts don't get left behind
- Added ether gate spell that was previously stubbed out and never implemented.
- added dimensional fold skill for psionic.


code/code/disc/disc_alchemy.cc
```diff
@@ -2208,28 +2209,109 @@ int castGalvanize(TBeing *caster, TObj *obj)
...
+    if (tPerson)
+      next_tmp_obj->checkOwnersList(tPerson);
+
+    caster->roomp->playsound(SOUND_SPELL_PORTAL, SOUND_TYPE_MAGIC);
+    rp->playsound(SOUND_SPELL_PORTAL, SOUND_TYPE_MAGIC);
+
+    act("$p suddenly appears out of a swirling mist.", TRUE, caster, tmp_obj, NULL, TO_ROOM);
+    act("$p suddenly appears out of a swirling mist.", TRUE, caster, tmp_obj, NULL, TO_CHAR);
+
+    sprintf(buf, "%s suddenly appears out of a swirling mist.\n\r", (sstring(next_tmp_obj->shortDescr).cap()).c_str());
```
cizra, line 2289: unsafe. Let's remove buf completely, instead use format

code/code/disc/disc_hand_of_god.cc
```diff
@@ -891,27 +891,26 @@ int portal(TBeing * caster, const char * portalroom, int level, short bKnown)
...
-    caster->roomp->playsound(SOUND_SPELL_PORTAL, SOUND_TYPE_MAGIC);
-
     TPortal * next_tmp_obj = new TPortal(caster->roomp);
+    next_tmp_obj->setPortalNumCharges(num_charges);
     *rp += *next_tmp_obj;
 
     if (tPerson)
       next_tmp_obj->checkOwnersList(tPerson);
 
+    caster->roomp->playsound(SOUND_SPELL_PORTAL, SOUND_TYPE_MAGIC);
```
cizra, line 913: I wonder if we should take the trouble to add MSP support for webclient?

code/code/disc/disc_psionics.cc
```diff
@@ -896,6 +900,118 @@ int TBeing::doPsidrain(const char *tString){
...
+    next_tmp_obj->setDescr(sbuf);
+
+    *rp += *next_tmp_obj;
+
+    act("$n winces in concentration.", TRUE, this, NULL, NULL, TO_ROOM);
+    act(format("You wince as you focus on %s.") % rp->name, TRUE, this, NULL, NULL, TO_CHAR);
+    act("$p suddenly appears out of a swirling mist.", TRUE, this, tmp_obj, NULL, TO_ROOM);
+    act("$p suddenly appears out of a swirling mist.", TRUE, this, tmp_obj, NULL, TO_CHAR);
+
+    sprintf(buf, "%s suddenly appears out of a swirling mist.\n\r", (sstring(next_tmp_obj->shortDescr).cap()).c_str());
```
cizra, line 1002: sprintf is unsafe.

cizra (changes requested):

code/code/disc/disc_hand_of_god.cc
wershlak: I've never used sound and wouldn't use it. Seems like it would be a cool thing to add into the web client though

code/code/disc/disc_alchemy.cc
```diff
@@ -2208,28 +2209,109 @@ int castGalvanize(TBeing *caster, TObj *obj)
...
+int etherealGate(TBeing *caster, const char * portalroom, int, short bKnown)
+{
+  char buf[256];
+  TPerson *tPerson = dynamic_cast<TPerson *>(caster);
+  int i, location=0;
+  int num_charges = 20;
+  sstring sbuf;
+
+
+  for(i=0;i<NUM_GATE_ROOMS;++i){
```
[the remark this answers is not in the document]
wershlak: I copy pasta. Yeah just using a map is better.

code/code/disc/disc_alchemy.cc
```diff
@@ -2208,28 +2209,109 @@ int castGalvanize(TBeing *caster, TObj *obj)
...
+    act("Nothing seems to happen.", FALSE, caster, NULL, NULL, TO_ROOM);
+    return SPELL_FAIL;
+  }
+
+  if (caster->bSuccess(bKnown,caster->getPerc(),SPELL_ETHER_GATE)) {
+    TPortal * tmp_obj = new TPortal(rp);
+    tmp_obj->setPortalNumCharges(num_charges);
+    tmp_obj->obj_flags.decay_time = 5;
+    tmp_obj->name = "portal gate";
+    tmp_obj->shortDescr = "<c>A <B>sw<W>ir<1><b>li<C>ng <1><c>ethereal gate<1>";
```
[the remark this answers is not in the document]
wershlak: I thought about that but I already run into issues because the SQL files here aren't in sync with the mud. I think the code is better if it doesn't rely on loading specific vnums.

This code does look kind of ugly to me too though. I feel like it would be better to pass these as parameters to the constructor rather than create the object and then modify a bunch of stuff. 

In go I'd probably create a struct to pass to the constructor but does that sound like proper the proper way in C++?

code/code/disc/disc_hand_of_god.cc
Aion2501: I haven't used sound since the old Brutius client that came with a sound pack.

In this day and age there are TONS of free music and sound effect pack assets that come with the myriad of game-making programs. It probably wouldn't be very difficult to put something together.

It certainly wouldn't hurt to build towards the eventuality of having some sound effects in the web client.

cizra (approved):

PR #326, opened 2021-03-17
WIP: fix shop_buy

wershlak:
@cizra wanted to get your take on this.

```
// delete objects left on keeper, remove objects sold from db
  for(unsigned int i=0;i<objects_p.size();++i) {
    TObj *cleanupObj = objects_p[i];
    if (!cleanupObj)
      continue;
    if (cleanupObj->parent == keeper)
      delete objects_p[i];
    else
      keeper->deleteItem(shop_nr, objects[i]);
  }
}
```
The intent of the above code looks like this:
- If we somehow at the end of this have any items left on the keeper we want to delete them
- otherwise we assume they've been sold and we call deleteItem which updates the DB.

This code moves the deleteItem and eliminates the extra delete. My test instance isn't crashing but can you think of something that might go wrong or a better way to handle this?

### DIFF
--- a/code/code/misc/shop.cc
+++ b/code/code/misc/shop.cc
@@ -552,8 +552,6 @@ void shopping_buy(const char *arg, TBeing *ch, TMonster *keeper, int shop_nr)
       rent_id=convertTo<int>(db["rent_id"]);
       temp1=keeper->loadItem(shop_nr, rent_id);
       *keeper += *temp1;
-      vlogf(LOG_BUG, format("created: %s ") % temp1->getName());
-
       objects.push_back(rent_id);
       objects_p.push_back(temp1);
       ++i;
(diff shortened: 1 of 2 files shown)


code/code/misc/shop.cc
```diff
@@ -552,6 +552,8 @@ void shopping_buy(const char *arg, TBeing *ch, TMonster *keeper, int shop_nr)
       rent_id=convertTo<int>(db["rent_id"]);
       temp1=keeper->loadItem(shop_nr, rent_id);
       *keeper += *temp1;
+      vlogf(LOG_BUG, format("created: %s ") % temp1->getName());
```
cizra, line 555: Is that a bug? Why?

cizra:
@wershlak I don't know the data model of Sneezy well enough to know what is held where. Therefore, let's get it merged and see if it actually works :) Can always revert later.

Given enough spare time, what we should do is:
1. set up a tiny test instance - something with 1 room, 1 shopkeeper, 5 objects
2. log allocations and deallocations of all objects by pointer value
3. eyeball it that when buying stuff, all objects are created/destroyed/moved properly
4. ... and finally pack it all up as an integration test

cizra (approved):

code/code/misc/shop.cc
wershlak: No. I need to remove that.

wershlak:
> @wershlak I don't know the data model of Sneezy well enough to know what is held where. Therefore, let's get it merged and see if it actually works :) Can always revert later.
> 
> Given enough spare time, what we should do is:
> 
> 1. set up a tiny test instance - something with 1 room, 1 shopkeeper, 5 objects
> 2. log allocations and deallocations of all objects by pointer value
> 3. eyeball it that when buying stuff, all objects are created/destroyed/moved properly
> 4. ... and finally pack it all up as an integration test

That sounds pretty awesome. Part of the difficulty is that each object type has its own buyMe method. There’s a lot of different objects to test.

wershlak:
@cizra I updated the logic here. I found that I could still crash it with the previous fix in the case where part of the transaction was successful like running out of money while buying a bunch of comps.

This way iterates all items in the keeper rather than all items we loaded as part of the transaction. So if you give something to a keeper it will get deleted the next time someone tries to buy something. I think this is fine... at least it's less crashy

PR #327, opened 2021-03-21
add new way of iterating over mobs. Hopefully fixes the proc crash

cizra:


code/code/sys/CharacterList.cc
```diff
@@ -0,0 +1,47 @@
...
+
+extern TBeing *character_list;
+
+TCharacterList::iterator::iterator(TBeing* d)
+  : d(d)
+{}
+
+TCharacterList::iterator TCharacterList::iterator::operator++()
+{
+  d = d->next;
```
wershlak, line 13: Won't we see a crash here in the case the being that d points to is deleted?

wershlak:
My thought to how to fix this tmp_ch issue was that in the TBeing destructor we need to move the tmp_ch pointer in the case that is the thing we are deleting.

```
if this == tmp_ch
  tmp_ch = tmp_ch->next
```
Or some such thing.

At least that seems the most straightforward. It doesn't really feel good though

code/code/sys/CharacterList.cc
cizra: We would, but modifying the loop range while looping is a really, really bad idea anyway. I hope we'll never do that again.

cizra:
I agree, it'd be a simple hack - reaching to the other end of the code, poke the private variable inside an unrelated class... thankfully it's a private variable, which exposes the uncleanliness of such a movement.

wershlak:
Alright cool. Glad that makes sense. 

You have this running right? Are you satisfied this fixes it?

wershlak (approved):

cizra:
It's currently running in prod, yeah.
Absolutely not sure! :D I can only hope it works.

PR #328, opened 2021-03-23
allow longer zone reset timers

wershlak:
Pretty sure this works fine. Only done a little testing but it seems simple.

@Aion2501 Currently 998 minutes is the longest reset time that works. This changes it to 9998.

 I know we were experimenting with different numbers. in a few zones. After we merge this you should adjust those again to where you want it.


Aion2501 (approved):
Yes those numbers in the zonefile represent a pre-set tick timer in the world list.
For zone repop I think each tick is 2.5 minutes or something.

So I need around 1700 ticks in order to make a 3 day zone repop viable. 

Very happy for this change. Hope it doesn't break.

cizra (approved):

PR #335, opened 2021-05-21
Ensure the corpses directory exists

diatomic-ge:
This change ensures the `corpses` and `corpses/corrupt` directories exist when starting the server.

This was missing from the other directories that are checked on startup, and this should prevent a crash that can happen when someone dies when the server is started via the docker compose scripts, or the directory isn't manually created, and sneezymud fails to open `corpses/<playername>`.

### DIFF
--- a/code/code/sys/db.cc
+++ b/code/code/sys/db.cc
@@ -402,6 +402,7 @@ void bootDb(void)
   bootPulse("Verifying runtime lib dirs.");
   verify_path("roomdata/saved");
   verify_path("immortals");
+  verify_path("corpses/corrupt");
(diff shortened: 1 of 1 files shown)


cizra (approved):

cizra:
@diatomic-ge I don't think it has anything to do with docker-compose, but the fix looks good, thanks.

diatomic-ge:
@cizra Oh, no, the only thing it has to do with the docker-compose files in https://github.com/sneezymud/sneezymud-docker is that the docker-compose files don't make the `corpses/corrupt` directory on their own, so running sneezymud using them crashes when players die. That's what I meant

PR #342, opened 2021-06-14
Github Actions

wershlak:
@cizra Can you take a look here? I think ideally we'd have the docker file for the mud here with the mud and CI to publish it all here.
- I removed the travis stuff and replaced it with github actions. Nice to see all the output here.
- I created a new docker hub account "sneezymuddeployer" and have it uploading. We need to invite this to the sneezy mud org and I can fix up this PR
- https://hub.docker.com/repository/registry-1.docker.io/wershlak/sneezymud2/tags?page=1&ordering=last_updated
- See my comments: this is working but I feel like the code checkout should just be in the CI and not in the docker file but I didn't get that working yet.


Aion2501 (approved):

.github/workflows/docker.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/checkout@v2
+        name: Check out code
+        with:
+          fetch-depth: 0
+          submodules: 'true'
+
+      - uses: mr-smithers-excellent/docker-build-push@v5
+        name: Build & Push Docker image
+        with:
+          image: wershlak/sneezymud2
```
wershlak, line 21: Need to fix

Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=utc apt-get install --yes --no-install-recommends build-essential libboost-dev libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libboost-system-dev libmariadbclient-dev scons libcurl4-openssl-dev git ca-certificates
+ARG UID=1000
+RUN useradd -m -u $UID sneezy
+USER sneezy
+ARG BRANCH="master"
+# COPY . /home/sneezy/sneezymud
+WORKDIR /home/sneezy
+ARG FORCE_REBUILD=1
+RUN echo Building from branch: $BRANCH; git clone --depth 1 --shallow-submodules --recurse-submodules --single-branch --branch $BRANCH --no-tags https://github.com/sneezymud/sneezymud
```
wershlak, line 12: So I'd really love to remove this but it didn't work locally to just copy the code into the container as on line 9

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+        fetch-depth: 0
+
+    - name: C++ Style Check (compiled)
+      uses: Dal-Papa/actions-uncrustify@v1.2
+
+    - run: sudo apt-get update
+    - run: sudo apt-get install software-properties-common
+    - run: sudo apt-get update
+    - run: sudo apt-get install scons g++-9 libboost-dev libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libmariadbclient-dev libmariadb3 ccache
+    - run: scons -C code -j4 -Q ccache=1 pretty=0 sanitize=1 shared=0 check sneezy
```
cizra, line 21: It'll hang, as Sneezy won't exit. Besides, you don't have the DB set up.
Would it be possible instead to use a compose file, combining DB with a MUD client logging in as admin and shutting down the MUD? Maybe also run a couple of commands. We could use this to build up a suite of blackbox tests.

Dockerfile
cizra: Huh, why? It's good practice to build inside the container, as it'll stabilize your build environment (Ubuntu updates in the pipeline's worker agents won't cause build to break because Boost updated a version, or such.

cizra (changes requested):

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+      with:
+        fetch-depth: 0
+
+    - name: C++ Style Check (compiled)
+      uses: Dal-Papa/actions-uncrustify@v1.2
+
+    - run: sudo apt-get update
+    - run: sudo apt-get install software-properties-common
+    - run: sudo apt-get update
+    - run: sudo apt-get install scons g++-9 libboost-dev libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libmariadbclient-dev libmariadb3 ccache
```
cizra, line 20: Why would you build it outside of container, if it's possible to build inside container?

cizra (changes requested):

.github/workflows/test.yml
cizra: oops, it'll just compile, not run it. Disregard the bit about hanging.

.github/workflows/test.yml
wershlak: Ooh so this was me copying over the travis stuff. Apparently, that's what the travis CI was doing when we created PRs. This is a separate workflow from the container build and is just meant as a test. We could just build containers on every commit in every branch (as a way of testing that the code change is good). Think that's better?

.github/workflows/docker.yml
wershlak: fixed

Dockerfile
wershlak: Yeah exactly. Ok so since there might be some additional bits to work out I'm going to leave it this was and get back to it another time.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,26 @@
+name: Build and Publish
+on:
+  push:
+    branches:    
+      - 'master'
+      - 'add_github_actions'
+jobs:
+  build_image:
+    runs-on: ubuntu-latest
```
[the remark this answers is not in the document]
wershlak: This is for the CI container. I updated this line to specifically use ubuntu-20.04

.github/workflows/test.yml
wershlak: I updated the other things and removed the WIP. I think this is ready to merge unless you think we should do something different with this test workflow.

cizra: Yeah, we could build the container but avoid publishing it, so as not to spew junk into Docker Hub.

cizra:
Why's the Travis build still required by GitHub, if you removed the .travis.yml? It's blocking the merge.

.github/workflows/test.yml
```diff
@@ -0,0 +1,23 @@
...
+  build_and_test:
+    runs-on: ubuntu-latest
+    name: Build and Test
+    steps:
+    - name: Checkout this commit
+      uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: C++ Style Check (compiled)
```
cizra, line 14: Where can I see the output of this?

cizra (changes requested):

.github/workflows/test.yml
wershlak: The action says it only runs on files added in a specific commit. https://github.com/marketplace/actions/c-style-check-compiled

The output is on the "checks" tab of the PR but there's nothing interesting for this one.

I haven't tested it at all though. I'm expecting a style check could be painful for our codebase but we can try it out.

cizra: OK, fair enough

.github/workflows/test.yml
wershlak: I _think_ the travis stuff is still here cause it was here when I started the branch and the PR. I need to test that out.

Other than that I need to fix the test here but I'm super behind after all that time off. I'm going to merge this so we can get things publishing and fix the test when I get a chance. Feel like there's so many PRs in the pipeline 😱

PR #346, opened 2021-06-24
Update Zonefiles/Update DB changes

wershlak:
# Zonefile Updates

Starting an MR to update zone files with some of the changes folks are working on for the sneezy-mud instance. We'll continue to build up this branch as we update more items. Just want to ensure it's preserved somehow for now.

## background

Currently, zone files are in the lib directory which is preserved in a docker volume. So a "new" build will get these changes as we commit. Existing instances will need to manually put these in place. I'll take a crack at fixing this process later unless someone else fixes it first.

## Item Updates

We need to contribute the DB updates back at the same time but aren't able to do that just yet. More to come.


cizra:
Here https://github.com/sneezymud/sneezymud/tree/mutable-dir is my attempt at decoupling mutable from fixed data. I don't quite recall where it went wrong, but it didn't really work. Perhaps you can use it as a starting point in your endeavour.

wershlak:
@cizra I've got all the DB updates here for the latest updates on live. This should be mergeable as far as I know.

Can you think of any reason it wouldn't be?

cizra (approved):

PR #348, opened 2021-06-27
Cleric/deik/shaman buff spells

wershlak:
I had some time to start on this. It's a bit bigger than I thought but seems like everything works.


code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
 #include "obj_magic_item.h"
 #include "obj_player_corpse.h"
 
+void crusade(TBeing *ch)
+{
+  affectedData aff1, aff2, aff3, aff4, aff5, aff6;
+  int level;
```
cizra, line 16: No point in declaring it 3 lines above where it's actually written into.

code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
...
+  TBeing *tmp_victim = NULL;
```
cizra, line 17: nullptr would be slightly more modern

code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
...
+
+  if (!bPassClericChecks(ch,SPELL_CRUSADE))
+    return;
+  
+  level = ch->getSkillLevel(SPELL_CRUSADE);
+
+  if (ch->bSuccess(ch->getSkillValue(SPELL_CRUSADE), ch->getPerc(), SPELL_CRUSADE)) {
+    int found = FALSE;
+    TThing *t=NULL;
```
cizra, line 26: Why declare it a kilometer away from usage?

code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
...
+    aff6.location = APPLY_INT;
+    aff6.modifier = 10;
+    aff6.bitvector = 0;
+
+
+
+    ch->sendTo("A <y>golden ray of light<1> shines down from the sky as you call the crusade.\n\r");
+    act("$n looks to the sky as a <y>golden ray of light<1> shines down on $s group.", TRUE,ch,0,0,TO_ROOM);
+
+    for(StuffIter it=ch->roomp->stuff.begin();it!=ch->roomp->stuff.end() && (t=*it);++it) {
```
cizra, line 80: Badly needs spaces. Also, why's the derefence in the for's condition? Can `roomp->stuff` ever contain nulls? I don't recall other iterations over `roomp->stuff` checking for nulls there.

code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
...
+        act("You feel more focused on your crusade.", TRUE, tmp_victim, NULL, NULL, TO_CHAR);
+
+        tmp_victim->affectJoin(ch, &aff1, AVG_DUR_NO, AVG_EFF_NO, false);
+        tmp_victim->affectJoin(ch, &aff2, AVG_DUR_NO, AVG_EFF_NO, false);
+        tmp_victim->affectJoin(ch, &aff3, AVG_DUR_NO, AVG_EFF_NO, false);
+        tmp_victim->affectJoin(ch, &aff4, AVG_DUR_NO, AVG_EFF_NO, false);
+        tmp_victim->affectJoin(ch, &aff5, AVG_DUR_NO, AVG_EFF_NO, false);
+        tmp_victim->affectJoin(ch, &aff6, AVG_DUR_NO, AVG_EFF_NO, false);
+
+        found = TRUE;
```
cizra, line 97: We don't need to use these pseudo-bools anymore, since C++ invented real bools around 30-odd years ago :)

code/code/disc/disc_aegis.cc
```diff
@@ -10,6 +10,102 @@
...
+      }  
+    }
+    if (!found)
+      ch->sendTo("But, there's nobody in your group.\n\r");
+    
+    return;
+  } else
```
cizra, line 104: By now, I've lost context of what that else is. Perhaps rearrange it, drop out of the function on fail condition?

code/code/disc/disc_aegis.h
```diff
@@ -34,16 +34,19 @@ class CDAegis : public CDiscipline
...
     CSkill skSecondWind;
     CSkill skRelive;
+    CSkill skCrusade;
 
     CDAegis()
       : CDiscipline(),
         skSanctuary(), skCureParalyze(),
-        skSecondWind(), skRelive() {
+        skSecondWind(), skRelive(),
+        skCrusade() {
```
cizra, line 43: You don't need to explicitly call the default constructor

code/code/disc/disc_aegis.h
```diff
@@ -34,16 +34,19 @@ class CDAegis : public CDiscipline
...
     }
     CDAegis(const CDAegis &a)
       : CDiscipline(a),
         skSanctuary(a.skSanctuary), skCureParalyze(a.skCureParalyze),
-        skSecondWind(a.skSecondWind), skRelive(a.skRelive) {
+        skSecondWind(a.skSecondWind), skRelive(a.skRelive),
+        skCrusade(a.skCrusade) {
```
cizra, line 49: Do these copy constructors / assignment operators do anything useful? Why not just delete the whole bunch of them?

code/code/disc/disc_shaman.cc
```diff
@@ -1050,8 +1046,7 @@ int cheval(TBeing *caster, TBeing *victim, int level, short bKnown)
     aff.type = SPELL_CHEVAL;
     aff.level = level;
     aff.duration = caster->durationModify(SPELL_CHEVAL, (aff.level / 3) * Pulse::UPDATES_PER_MUDHOUR);
-    aff.modifier = -50;
-    aff.location = APPLY_ARMOR;
+    aff.location = APPLY_NONE;
```
cizra, line 1049: So what'll it do?

code/code/disc/disc_shaman_armadillo.cc
```diff
@@ -421,8 +429,7 @@ int celerite(TBeing *caster, TBeing *victim, int level, short bKnown)
     aff.type = SPELL_CELERITE;
     aff.level = level;
     aff.duration = caster->durationModify(SPELL_CELERITE, (aff.level / 3) * Pulse::UPDATES_PER_MUDHOUR);
-    aff.modifier = -100;
-    aff.location = APPLY_ARMOR;
+    aff.location = APPLY_NONE;
```
cizra, line 432: so what'll it do now?

code/code/misc/spell_parser.cc
```diff
@@ -2413,6 +2414,9 @@ int TBeing::doDiscipline(spellNumT which, sstring const& n1)
       case SPELL_SANCTUARY:
           sanctuary(this, ch);
         break;
+      case SPELL_CRUSADE:
```
cizra, line 2417: This monstrosity could be changed into a neat little lookup table of spellnum -> function.

cizra (changes requested):

code/code/disc/disc_shaman.cc
wershlak: The main point of cheval is reducing command lockout: https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/utility.cc#L1569

code/code/disc/disc_shaman_armadillo.cc
wershlak: Reduces command lockout and does haste 
https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/utility.cc#L1569
https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/offense.cc#L2260

wershlak: Since Shaman don't wear armor I guess the coder added some extra AC onto their spells. But now we have shiny silver potions and everyone has these spells on at all times. This just brings them in line with mage accelerate and haste.

code/code/disc/disc_aegis.cc
wershlak: Hrmm. Looking through the code it's about half and half... But I'm not sure why one versus the other

code/code/disc/disc_aegis.cc
wershlak: I always wondered why these were in the code!

wershlak:
I'm going on vacation but I'll fix this later.

code/code/disc/disc_aegis.h
wershlak: This is boilerplate that's in every discipline. I'm going to leave this here for now but I'll remove them all in one seperate commit and make sure nothing breaks.

code/code/misc/spell_parser.cc
wershlak: I'm trying to figure this out.  Like it seems like it'd be possible to make a map of spell number to a function but then calling the function that may have different return types is tripping me up. Any help?

Issue #163, opened 2018-01-31
Physical skill success formula audits


drewhershey:
**Tl;Dr:**
I think there's an opportunity to improve gameplay quality and increase the benefit of character specialization if we tweak how success chances for various skills are calculated. By going through all the active skills and updating how various aspects of the attacker/victim/environment contribute to the two different types of success checks we could potentially make combat more dynamic and strategic while still maintaining a healthy level of challenge.

**Full Post**
After spending a lot of time looking through the code off and on over the last couple years, I think the way that skill success is being handled could potentially benefit from some updates. This applies specifically to physical skills, as mental skills like spells/prayers use a different type of secondary success roll (although that could stand to be reevaluated, too).

I'm curious if others feel like this is even remotely necessary or worth the time.

Currently, for physical skills there are two separate success checks. Here's my interpretation of how they work and their underlying purpose:

1) A roll within a range from 0 - 99 that more or less represents if a character knows _how_ to perform the skill.

     In other words, the chance of success based on character's understanding of the skill itself and level of experience actually executing it in practice.

2) A roll within a range from 0 - 999 (0 - 49 being a crit failure and 950 - 999 a crit success) that represents the character's chance of success against a _specific target in a specific situation_. 

     In other words, it assumes the character will execute the mechanics of the skill correctly but then determines if external factors (the surrounding environment, defensive skill of the intended victim, etc.) will still cause the attempt to fail.

     Practical example: A thief could pass the first roll and therefore perfectly execute their backstab mechanically, yet still miss if the intended victim was agile enough or the environment messed up his otherwise solid attempt. Alternatively, a victim that would normally be able to avoid the attack due to high agility might instead be hit if affected negatively enough by their environment (too dark, too loud, etc.).

Those two different checks make sense in theory. However, what appears to have happened over the years is that various factors which seem like they should be accounted for during the second, more dynamic success roll (the one that deals with the specifics of the situation) are being factored into the first roll, during what should be more of a "does the character flat out possess the knowledge/experience to perform this skill at all" check. 

Or, oftentimes, things that should be in that second roll are instead just used as criteria for hard skill failures - "you can't do that now" type responses.

The warrior bash skill is a good example. To me something like the attacker's height and weight would primarily affect their success chance against a specific victim, yet those stats are only accounted for during the first roll. 

[...]

PR #349, opened 2021-06-29
Warrior Revamp

parkscs:
This pull request is related to the warrior gameplay issue thread I recently created.  My plan is to tackle these changes in a number of different phases, starting with granting warriors 2h specialization, making a usability tweak to rescue, and some other minor things.

The next phase will be to revamp deathstroke to be a core ability for the warrior class, which involves making it actually useful while keeping some of the risk involved in using it.  I'd also like to address movement point usage for warriors and potentially take a look at berserking to ensure all the restrictions make sense and the risk remains worth the reward.

From there, I plan to look at all the other offensive warrior abilities like headbutt, kneestrike, bodyslam and so on and look to give players a reason to use these abilities (at least in some situations).


Aion2501 (commented):
Im very happy to see someone else working on improving classes!

That said, it would be nice to share a google doc and get a design plan laid out. Balance is important and I'm sure we could all share ideas.

Are warriors going to look more like Barbarians from D&D? Warriors from WoW? Is the end result a balanced class compared to others? 

A final note: I think some kind of berserking self-heal or lifesteal a % of dmg ability would be nifty. If nothing else perhaps some kind of Medic skill.

parkscs:
Responded in the issues thread, as I don't want to clutter this one up with lengthy balance discussions.  Happy to continue the discussion there and as we go forward.

code/code/misc/combat.cc
```diff
@@ -1992,9 +1992,10 @@ int TBeing::getWeaponDam(const TBeing *v, const TThing *wielded, primaryTypeT is
     // 2h Specialization: should be no additional dam for anyone without specialization
     // extra 10% for 2h is part of the object
     TObj *obj = dynamic_cast<TObj *>(heldInPrimHand());
-    if (obj && obj->isPaired() && doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
+    if (obj && obj->isPaired() && (doesKnowSkill(SKILL_2H_SPEC_DEIKHAN) || doesKnowSkill(SKILL_2H_SPEC_WARRIOR))) {
```
wershlak, line 1995: Check out how SKILL_DUAL_WIELD is handled just above here. Both warriors and thieves have dual wield
- There are two skills SKILL_DUAL_WIELD and SKILL_DUAL_WIELD_THIEF
- getSkillNum() will return the appropriate skill based on learnedness: https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/discipline.cc#L632

wershlak:
Couple general comments on changes.

- In your description, add a bullet point list of things that this change is meant to address. That way we can review not just the code but if your intentions match your implementation. 

- Remove any code that is no longer needed rather than commenting it out. We have the entire history in git so we don't need it for historical record.

- Comments can help explain the reasoning behind a particular implementation but aren't really needed to explain why you are changing a thing. Think of someone looking at the code 20 years from now. They won't have context around your change.

parkscs:
All makes sense and I can incorporate that into future submissions.

parkscs:
Itemized changes are as follows:

- Added 2h specialization to warriors and more specifically to the soldiering discipline.
- Fixed a bug where monk crit was only honing with barehanded attacks (yet it works with all attacks).
- Removed some restrictions to rescue that were bad from a gameplay perspective (can't rescue mounted people, can't rescue NPC's including mounts you got knocked off of).
- Added checks to the order command so that players will benefit from advanced riding skills of the relevant type when ordering their mount; this was needed because some mounts like the gold dragon were resisting commands 15+ times with max skills.  I could go on and happy to elaborate if needed, but this is terrible for gameplay.
- Shortened the lockout on unsuccessful berserk attempts (more to come for berserk later).
- Minor buff to riposte, allowing it to fire on all parries.

code/code/cmd/cmd_rescue.cc
```diff
@@ -31,19 +31,12 @@ static int rescue(TBeing * caster, TBeing * victim, spellNumT skill)
     caster->sendTo("What about fleeing instead?\n\r");
     return FALSE;
   }
-  if (dynamic_cast<TBeing *>(victim->riding)) {
```
wershlak, line 34: Ok cool

code/code/cmd/cmd_rescue.cc
```diff
@@ -31,19 +31,12 @@ static int rescue(TBeing * caster, TBeing * victim, spellNumT skill)
...
-    caster->sendTo("You can't rescue a mounted person!\n\r");
-    return FALSE;
-  }
+
   if (victim->riding) {
     caster->sendTo(COLOR_MOBS, format("You can't rescue %s off of %s!\n\r") % 
          victim->getName() % victim->riding->getName());
     return FALSE;
   }
-  if (!victim->isPc() && !victim->isPet(PETTYPE_PET | PETTYPE_CHARM | PETTYPE_THRALL)) {
```
wershlak, line 43: Sure

code/code/disc/disc_warrior.cc
```diff
@@ -95,7 +95,7 @@ int berserk(TBeing * caster)
 
     af.type = SKILL_BERSERK;
     af.level = level;
-    af.duration = 6 * Pulse::UPDATES_PER_MUDHOUR;
+    af.duration = Pulse::UPDATES_PER_MUDHOUR/2;
```
wershlak, line 98: Yeah the cooldown time was always annoying

code/code/misc/combat.cc
```diff
@@ -1992,9 +1992,9 @@ int TBeing::getWeaponDam(const TBeing *v, const TThing *wielded, primaryTypeT is
     // 2h Specialization: should be no additional dam for anyone without specialization
     // extra 10% for 2h is part of the object
     TObj *obj = dynamic_cast<TObj *>(heldInPrimHand());
-    if (obj && obj->isPaired() && doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
+    if (obj && obj->isPaired()) {
       // Take half the skill value and normalize between 10 and 50
-      int amt = (int) getSkillValue(SKILL_2H_SPEC_DEIKHAN);
+      int amt = max(0, (int) getSkillValue(SKILL_2H_SPEC));
```
wershlak, line 1997: Does this work? Does it need a getskillvalue(getskillnum(THING)) here instead to get the right skill?

code/code/misc/combat.cc
```diff
@@ -2047,6 +2047,9 @@ static void checkLearnFromHit(TBeing * ch, int tarLevel, TThing * o, bool isPrim
         if (obj && ch->doesKnowSkill(SKILL_2H_SPEC_DEIKHAN) && obj->isPaired()) {
           ch->learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_2H_SPEC_DEIKHAN, max(0, (100 - (2* myLevel))));
         }
+        if (obj && ch->doesKnowSkill(SKILL_2H_SPEC) && obj->isPaired()) {
```
wershlak, line 2050: Same. Do we need to call get skill num or is that included in doesknowskill()?

code/code/misc/combat.cc
```diff
@@ -2292,12 +2295,13 @@ int TBeing::hit(TBeing *target, int pulse)
 	 !equipment[WEAR_HAND_R] && !equipment[WEAR_HAND_L])
 	learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_IRON_FIST, 20);
 
-      if(doesKnowSkill(SKILL_CRIT_HIT))
-        learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_CRIT_HIT, 20);
     }
     if (((fx > 0.999) || (fy > 0.999))){
       if(doesKnowSkill(SKILL_POWERMOVE))
 	learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_POWERMOVE, 20);
+
```
wershlak, line 2302: Cool. In the future is probably do this with another PR but this is easy and fine

parkscs:
Slowly figuring out git I think - definitely going to change how I manage some things locally going forward, haha.  In any event, you're absolutely correct on that check in combat.cc - I didn't catch that after reworking it to match the dual wield check.  Should be better now.

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -62,69 +62,113 @@ static int deathstroke(TBeing *caster, TBeing *victim)
...
       act("You hit $N in $S vital organs!", FALSE, caster, 0, victim, 
                 TO_CHAR);
       act("$n hits you in your vital organs!", FALSE, caster, 0, 
                 victim, TO_VICT);
+
+      // Adding a lockout
+      // You successfully landed your deathstroke - assessing a bonus
+      aff.type = SKILL_DEATHSTROKE;
+      aff.duration = Pulse::UPDATES_PER_MUDHOUR / 3;
+      aff.location = APPLY_HITNDAM;
```
wershlak, line 94: I'm not sure APPLY_HITNDAM works. Did you test how this affect works? Also, 15 sounds insane... at least for APPLY_HIT each +1 is like a levels worth of offense.

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -62,69 +62,113 @@ static int deathstroke(TBeing *caster, TBeing *victim)
...
   int dam = caster->getSkillDam(victim, SKILL_DEATHSTROKE, lev, caster->getAdvLearning(SKILL_DEATHSTROKE));
 
-  if ((caster->bSuccess(bKnown+ percent, SKILL_DEATHSTROKE) &&
-         (i = caster->specialAttack(victim,SKILL_DEATHSTROKE)) && 
-         (i != GUARANTEED_FAILURE)) ||
-      !victim->awake()) {
+  if (caster->bSuccess(bKnown+percent, SKILL_DEATHSTROKE) || !victim->awake()) {
     wasSuccess = TRUE;
+
     if (!dam) {
```
wershlak, line 68: How would this happen?

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -62,69 +62,113 @@ static int deathstroke(TBeing *caster, TBeing *victim)
 
   int dam = caster->getSkillDam(victim, SKILL_DEATHSTROKE, lev, caster->getAdvLearning(SKILL_DEATHSTROKE));
 
-  if ((caster->bSuccess(bKnown+ percent, SKILL_DEATHSTROKE) &&
-         (i = caster->specialAttack(victim,SKILL_DEATHSTROKE)) &&
```
wershlak, line 66: What's in specialAttack() here that you are removing?

parkscs:
I did not test to see if it works - just assumed they all worked and that
it was shorthand for hit and damage.  I can adjust the #'s - I was just
basing it off bless being 10 hitroll and that's a low level spell that you
can cast on your entire party, whereas this applies only to the warrior,
may not work and has consequences.  But easy enough to tweak the #'s.
Also, I looked at advanced offense and what that's giving to monks/thieves.
bonus += ((getSkillValue(SKILL_ADVANCED_OFFENSE) / 4.0) * 3.0);

If my math is correct, that's adding a max of 100/4*3... so 75?

Now if we look at hitroll and where that gets applied:

bonus += 5 * (getHitroll() + getSpellHitroll())/3;
So adding 15 hitndam, ignoring the dam for now and assuming it works, this
would be adding 5*15/3 as a bonus... so 25.

Again, I'm happy to tweak the #'s and of course we need to verify hitndam
actually is functional haha.  Just pointing out that I did look at other
skills when coming up with 15.

code/code/misc/combat.cc
wershlak: This still needs fixing

code/code/cmd/cmd_deathstroke.cc
parkscs: I removed it just because I wasn't really sure why all that was added to the skill check in the first place.  Honestly the initial code for the skill check was pretty convoluted (just my opinion) shifting bits in the algorithm and having not just one success check but multiple that had to succeed, which is part of the reason I suspect why it failed frequently in game.  My goal was to simplify the algorithm but to keep it tied to armor and to bring in some stats as well (including the victim's stats).

code/code/misc/combat.cc
```diff
@@ -1992,9 +1994,9 @@ int TBeing::getWeaponDam(const TBeing *v, const TThing *wielded, primaryTypeT is
     // 2h Specialization: should be no additional dam for anyone without specialization
     // extra 10% for 2h is part of the object
     TObj *obj = dynamic_cast<TObj *>(heldInPrimHand());
-    if (obj && obj->isPaired() && doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
+    if (obj && obj->isPaired()) {
       // Take half the skill value and normalize between 10 and 50
-      int amt = (int) getSkillValue(SKILL_2H_SPEC_DEIKHAN);
+      int amt = max(0, (int) getSkillValue(getSkillNum(SKILL_2H_SPEC)));
       amt = 100 + min(50, max(10, (int) (amt / 2)));
```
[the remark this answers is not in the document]
parkscs: Sounds like a blunder on my part and I did not test the skill on characters that didn't know it.  I'll look into this tonight and fix it up.

code/code/cmd/cmd_deathstroke.cc
parkscs: I didn't dig into how they're calculating skill damage deep enough to figure that out, but it seems like there are times when you succeed but the damage is non-existent.  A lot of stuff goes through getSkillDam() and so I didn't really mess with it at this point, but it's odd behavior if you ask me to have a skill succeed and somehow the damage is still 0.

code/code/cmd/cmd_deathstroke.cc
drewkht: Just to jump in here in hopes of helping a bit, as a lurker who wishes I had more spare time so I could contribute to the code, I dug into the specialAttack method a while back to see how it works. For all intents and purposes it's basically a comparison of attacker's offense vs defender's defense, but the formulas for calculating those offense and defense values rely heavily on level. 

Because of that reliance on level, there's a certain point where it's really difficult for a PC to increase their offense value any further, so specialAttack essentially ends up functioning as kind of a convoluted way to greatly reduce the chance physical PC special attacks (think backstab, deathstroke, etc) have of landing successfully on mobs 10-20+ levels higher (depending on PC offense). At around 20 levels higher the chance of the specialAttack check being successful becomes almost nil, usually.

So, without the call to specialAttack() here, what I believe will likely happen is that if the PC passes the bSuccess check (which is really just a check against their learnedness in the skill as a value from 0-100 where 100 = maxed/maxed, modified  +/- 25% by focus, and now I believe the difficulty check that was added back in) they'll be able to land deathstroke against a mob of any level without any test of offense vs defense.

Hopefully this helps. I still have the spreadsheet somewhere that maps out and tests the function so I can try to dig it up if you want more detail.

code/code/misc/skill_dam.cc
```diff
@@ -150,7 +150,7 @@ if (discArray[skill]->disc == discArray[skill]->assDisc) {
 
   // cut area effects in half
   if (IS_SET(discArray[skill]->targets, TAR_AREA)) 
-    fixed_amt *= 0.75;
+    fixed_amt /= 2.0;
```
wershlak, line 153: I just changed this. Why would you change it back?

parkscs: I didn't intentionally - my guess is that this code just didn't get that update.

code/code/cmd/cmd_deathstroke.cc
wershlak: Thanks @drewhershey I thought these were going through some hit check. Let me dig into this.

There should be some check for skill success AND there should be some check that compares user VS victim offense and defense. 

So for this @parkscs I would add that back in and let's start work (in a new branch) on an adjustment to how special attack hits are calculated. I know it sucks that this is hard to land but the problem is bigger than just death stroke and let's go ahead and adjust it for everything.

---

Here's some brainstorming thoughts on this:

skill success:
- skill success chance is based on learnedness and modified by the difficulty of the skill, focus and karma.
- should not be modified by other caster or victim stats or other things
- Is just essentially, do I know how to do this move?

hitting:
- offense takes into account level, skills, position, dex, equipment etc
- defense takes into account level, armor, skills, position
- So those two things really account for all factors. We shouldn't need to factor anything else.
- So it's just a matter of how we compare those two values.

However, I know that it is stupid hard to hit a much higher level mob. We need to adjust this formula so that higher level mobs aren't un-hittable but also still present a greater challenge.

drewhershey: Whoops, just realized I posted on my work account before. I wrote up an issue about this exact topic a few years back actually:

https://github.com/sneezymud/sneezymud/issues/163

Like I mention in that issue, it looks like over the years coders started confusing bSuccess checks (e.g. the first category @wershlak mentioned - whether the PC knows HOW to do something), the secondary check (referred to as "hitting") that determines if the attack succeeds based on factors out of the PC's control (weather, vision, level, whatever), and then a third kind of check that I think is referred to in the code as a "hard fail" check - basically an instafail scenario, like using a blunt weapon to backstab (although one could argue hard fails lead to less dynamic gameplay, rather than letting players deal with the consequences of their actions?).

Also, and you probably already know this, but, for most active skills (including prayers and spells, I'm pretty sure) there's two separate functions - <name> and "do<name>". So "backstab" and "doBackstab" for instance. The "do" function appears to be intended to contain all the hard fail checks, where a condition being true just returns FALSE immediately, or otherwise calls the "non-do" function if all the hard fail checks pass. Yet in many skills those types of hard fail checks have started getting placed in the "non-do" function.

For example: there's a check in doBackstab() to fail if the victim isn't found in the room - which makes sense. But then in backstab(), which gets called by doBackstab(), there's a hard fail if the thief is in a peaceful room. Why isn't that just in doBackstab() with the other hard fails?

It makes things confusing and introduces a lot of room for bugs IMO. So it could make sense to try to improve that separation of concerns at the same time, since it's related.

But I also very much agree that the physical special attack success formula could use some tweaks, based on my memories of hundreds of missed backstabs by my thief during high level groups over the years...

Oh, and if you want to see an example of how powerful a skill becomes when you remove the specialAttack() check, take a look at garrotte. It's super strong against high level casters because it only requires passing a bSuccess check to land.

parkscs:
What led me down this path was the existing code for deathstroke factors
enemy armor, player Dex and enemy Agi into the bSuccess check, which looks
to me like it's not a "does the PC know HOW to do something" but it's
blending that plus attributes of the secondary check like can you hit them
because of their armor+agility.

/* AC makes a big difference here ... */
percent = ((10 - (victim->getArmor() / 50)) << 1);
percent += caster->getDexReaction() * 5;
percent -= victim->getAgiReaction() * 5;
if (!caster->isImmortal())
caster->addToMove(-DEATHSTROKE_MOVE);
int lev = caster->getSkillLevel(SKILL_DEATHSTROKE);
int bKnown= caster->getSkillValue(SKILL_DEATHSTROKE);
int dam = caster->getSkillDam(victim, SKILL_DEATHSTROKE, lev, caster->
getAdvLearning(SKILL_DEATHSTROKE));
if ((caster->bSuccess(bKnown+ *percent*, SKILL_DEATHSTROKE) &&

Essentially what this is doing against high level opponents who have lots
of armor is adding a negative value onto their bKnown value which
represents how well the player knows the skill.  As such, it's not
functioning like garrotte does.  However, the deathstroke code then went on
to ALSO have a separate hit check and then the insta-fail check.
Ultimately I didn't see the need for the secondary check because the first
check already included attributes of the secondary check and just having
multiple separate hit checks is almost certain to reduce the chance of
success and muddy the waters as to what all is going on.  Add in that
there's a failure use-case built in when getSkillDam() returns 0 (which I
left in because I didn't want to mess with getSkillDam() as everything goes
through that function) and there were half a dozen ways for the skill to
fail (which is probably why it always fails in the normal code).

With that said, I'm not trying to suggest that's the "right" way to do it
and what you guys are proposing is much cleaner and more intuitive.  I'm
just pointing out why I ended up removing those checks in the first place,
particularly the 2nd check.

drewhershey:
No, I totally get your thought process. Encountering something similar in the warrior bash code was the exact thing that caused me to write that issue I linked. And deathstroke was actually in an even worse place, like you said, because its success was being modified negatively by the victim's AC in two separate places - in both the bSuccess and specialAttack checks. 

There's quite a few places I've found where previous coders modify the bSuccess check in this way, when they should be modifying the offense value inside specialAttack. Maybe because it was easier than modifying the specialAttack check in a clean way? Though it seems like it wouldn't be too hard to add an overload for specialAttack that accepts an extra modifier parameter used to modify the offense value. That'd probably allow the most flexibility to customize success chance on a skill-by-skill basis.

I just looked through the code again and the way deathstroke success/fail was originally coded is ridiculous, so I can see why you feel it needs changed. Correct me if I'm wrong, but I'm pretty sure it works as follows:
1) Get bKnown value from getSkillValue (range from 0 - 100, 100 meaning maxed skill)
2) Get percent value based on victim's armor or lack thereof (AC of 500 [which I believe is the default for a level 10 mob or level 0 PC per comments in combat.cc] sets this value to 0, AC of 2000 [default for lvl 60 mob] sets it to -60!!)
3) Get value based on attacker's dex (-20 at 5 dex up to 30 at 205 dex)
4) Do same for victim's agi
5) Add dex and agi mods to percent value, then add that to bKnown before passing to bSuccess call

So bKnown gets passed into the bSuccess with a value anywhere from 0 to 180+. The crazy thing is, coded this way a character with like a single prac in deathstroke but very high dex actually stands a reasonable chance of success if they're using it against a mob with bad AC and low agi. Starting with a bKnown of 1, you could potentially end up with something like 61+ after the mods haha. That makes no sense.

So then, inside bSuccess: 

6) The bKnown value gets capped at 100 before almost anything else. This means that low victim AC/high victim agi/low attacker dex can massively penalize an attacker with maxed deathstroke, while high bonuses from high victim AC/low victim agi/high attacker dex can only raise success chance up to the cap of 100.
7) THEN, the bKnown value is further reduced by skill difficulty (-10% for deathstroke as a "normal" difficulty skill)
8) Then modified by anywhere from -20% to 25% of the previous value based on focus
9) Then modified by anywhere from -10% to 12.5% _of the previous value again_ based on karma
10) Finally bKnown is compared to a random roll between 0 and 99 to see if the attempt succeeds

Then after all that, now the attacker has to pass a specialAttack check as well. Deathstroke must've been ridiculously hard to land all this time. 

Also, when skill success is coded this way it kind of turns focus and karma into super-stats - because they're factored in later on and are multiplicative they can compensate quite a bit for penalties that were factored in earlier like with deathstroke. 

My (unsolicited, I know, sorry) opinion is that just completely removing the "percent" modifier to the bKnown value being passed to the bSuccess check is the quickest/easiest/most effective way to immediately fix this specific issue with the skill. That way deathstroke would function similarly to other physical special attacks, and maybe tweaks could be made later on to address how specialAttack itself works?

parkscs:
I think that's where we should start if we want to clean this all up, and
then perhaps we need to look at how special attacks themselves work because
right now no one ever uses them at end-game because they don't really do
much and they drain too much in the way of moves.  When updating the skill
I tried to stick close to the original intent, but I completely agree it's
messy - to the point that I ended up writing a small script to just process
a bunch of AC values with the old formula to see how various armor/dex
values would be applied.  I ended up keeping the same variables
(armor/dex/agi), redid the algorithm and then iteratively tuned the
algorithm until it felt pretty good in game, but I think instead I should
have just scrapped all of that and just made it like all the other skills
(which would have been a lot easier lol).

There was also some weirdness in how the damage was calculated (it scaled
way down if the enemy knew deathstroke, so it was garbage damage against
all warrior NPCs) which I removed, but only after I had spent an hour or
two pouring through getSkillDamage() and I think it was called
genericDamage() and seeing no reason why it was hitting certain mobs for
10% damage; turns out they were calling those functions with different
parameters depending on whether the victim knew the ability (so basically
are they a warrior).

If we are in agreement, I can scrap the new algorithm and see how it
behaves if we just call bSuccess() with just the value returned by
getSkillValue() and add the other two checks back in.

code/code/cmd/cmd_deathstroke.cc
wershlak: Ok so hitndam works. My confusion is that it's not something you can apply to eq but it does work.

Here's how the apply affects a char:
https://github.com/sneezymud/sneezymud/blob/master/code/code/sys/handler.cc#L452-L454
so +15 hit == +150 hitroll

then in combat: https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/combat.cc#L2499

so the 150 hitroll turns into 250 offense. You should be able to verify by stating the PC or mob where this is applied.

I think something like +3 HITNDAM is more appropriate.

wershlak:
So let me just do one comment here regarding balance.

--

I set up warriors in my mind as tanks (along with deikhans). What this means to me is that they should have more ability to tank than any other class. I've added some stuff to help in advanced def but we aren't done. 

However, tanks should be able to deal out damage in groups almost as well as a hitter (monk/thief). They should just have to make choices and not do both at the same time. This, to me makes them more fun and interesting. I want a warrior to be like "Oh you want to tank now? Cool I'll pull out my 2h axe and start berserking!". It's just less fun if they were already doing that while tanking :) 

The end goal in my mind should be a warrior:

### Single Class Warrior
- Best tank for the majority of content.
 - Can take more of a melee pounding than deikhan
 - Has more control options which are useful for tanking
 - Deikhan probably better group utility and tanking vs casters since he has more protection vs avoidance
- Has more offense than a deikhan in general but not as much as a thief/monk. Has to commit to offense
 - Best control abilities should require a shield (like bash)
 - Best offense abilities should either reduce ability to mitigate damage or require 2h or dual wield
 
### Multi-Class Warrior
- Warrior or deikhan required for any kind of tank
- The best tank should somehow be WARRIOR/DEIKHAN/X - We need synergy and overlap in abilities
- Warriors should be a really strong choice for any character that wants to be a good hitter. BUT not required so a little less overlap/synergy here with monk/thief. Power move is a good example here. Monk grit hitting is strictly better than Powermove and they don't stack.
- Adds a ton of toughness and melee ability to a utility character (War/Mage/Cleric or something)


Deikhans aren't perfect and I learned some things and need to adjust some things further but:
- I added a protection proc that works off shield block
- I added a -50% protection debuff to smite to balance the offense buff
- Soloing you have a decision on auras but in a group you really don't so this was less successful than I hoped.

Things I might consider:
- I might change smite to require a 2her
- increase the debuff for smite
- currently auras have a "deikhan only bonus". I might change this to a self only bonus. Means if two deikhans in a group both want the best offense bonus they have to use the same aura. 


--

So that's me sharing my experience and vision. You don't have to agree with me about how warriors fit in.

However, you're changes here to berserk and death stroke look to have very strong offensive upside with no real downside. I would guess that warriors will use both while tanking most of the time.

I would suggest something along the line of:
### Deathstroke
- change the buff so that it is applied every time regardless of hit/miss damage or no damage
- -40% immunity to non-magic (armor reduction is mostly meaningless since warriors are way over the cap nowadays)
- +3 hitndam or +6dam maybe (lots of things have +hit but +dam is more unique and sort of makes more sense)

### Berserk
- You're currently adding +attacks, +offense and random auto-success bash
- remove the bash thing and replace it with another defensive debuff. random auto bash is pretty crazy. 5 berserking warriors can probably take down anything in the game.
  - - defense
  - - immunity non-magic
  - can't parry or block or something

parkscs:
*DEATHSTROKE*
I would disagree a bit on the no downside part to deathstroke.  The buff
only works on a successful hit, the counter attack can happen on a hit or a
miss, and more importantly for the point you're making about tanking, on a
miss there's a sizable penalty to the player's armor which a tank would
definitely not want.

aff.location = APPLY_ARMOR;
aff.modifier = caster->GetMaxLevel()*8 -
caster->getSkillLevel(SKILL_DEATHSTROKE);

My initial idea was that you're taking a risk when using it, and at least
under the previous algorithm, it had a pretty good chance of failing.  Sure
a tank could use it and not take a penalty if it lands, but previously it
was only landing <50% against end-game fights like Staley in my testing and
so there was a real chance that a tank gets punished for being aggressive
with it.  Rather than just making it always land and always do the same
thing, my goal was to make it more dynamic because while it originally
kinda sucked, I did like the fact that it had a real chance to come back
and penalize you and I wanted to keep that same dynamic/risky feel to it.

Based on the hitroll comment earlier, the hitndam number definitely needs
tweaking (and wow, bless is OP as hell lol).  And when we redo how the
skill check works, I think for this design to really work it needs to not
land 100% of the time, as I really want there to be a penalty/risk involved
in using it.

*BERSERK*
As for berserk, that's not complete yet, so I don't know how much you want
me to address it.  But I'll give you some thoughts on it.

For one, I didn't remove any of the penalties for berserking.  You say 5
berserking warriors can take stuff down, but you have to keep in mind you
can't flee while berserking and you can't cast while berserking.  This
doesn't really matter much to single class warriors but it's pretty
punishing for multi-class characters.  You're also set to aggro when
berserking and in some zones, that's a real problem (especially when you
can't heal yourself and you can't flee).

So while the penalties are pretty major, I wanted there to be an upside to
it as well.  I've been playing around with what all to add to it and as I
said it's not done yet, but perhaps the most important thing missing is
that the intention is not to have it always bash.  My plan is to have a
chance to fire off a random warrior ability (not including deathstroke)
including bash, but also headbutt, body spam, spin, grapple, kneestrike,
kick and whatever else I'm forgetting.  I stopped working because while
bash has a function I can easily call for a successful bash, the other
warrior skills either don't have such a function that's available to call
from this part of the code or I just can't figure out the syntax for it.
But if and when I can get this working, your hypo about 5 berserking
warriors will look a lot different.  You'd have 5 players that are set to
aggro, cannot flee, cannot cast, and might just throw out an extra kick or
headbutt or they might get lucky and bash.  I would argue that's not nearly
as OP as what you're suggesting.  However, that does play into the combo
system we previously talked about.  My thought is that the auto-fired
skills could act like a wild card in the combo logic when we get that
working.  I don't want them to mess up a combo you're working towards, but
I did want to create a dynamic play style where if you see a bash land
automatically, you could stop the combo you were going for and switch to a
combo that begins with bash if that's preferred.  This would let players
execute combos more quickly if they're paying attention to what
automatically fires during berserk and would enable them to land combos
more often simply because the auto-fired skill is a guaranteed hit and the
ones you use normally have a chance to fail.

But ultimately, it's worth noting that it's halfway complete at the moment
and it's not really ready for a review in this state.  That gives you an
idea of what I'm planning though.  Some of my major goals with this kit are
to make it dynamic and risky, at least when played offensively.  I'm
purposefully avoiding just having guaranteed buffs that always land or
removing penalties from the abilities.  We have plenty of other
skills/spells that always give you a buff, but abilities like deathstroke
and berserk are unique in the sense that you can absolutely get killed
using them incorrectly.  My goal with advanced berserk is really to just
make players consider using berserk more often, despite all the risks and
limitations involved in doing so.

drewhershey:
I was really impressed with the deikhan changes, personally, when I was messing around with them a while back in my dev copy of Sneezy. Thought they added a lot of cool flavor to the class and really made deikhans feel powerful in a unique way. They also felt really strong, but with how difficult high level mobs are (without cheesing them, at least) it seems like bringing other classes up to where deikhans are now would be a good goal.

One thing I thought of in response to your balance comments was this part: 

>  I would guess that warriors will use both while tanking most of the time.

Back when there was no limit on multiplaying I had a full group of like 8 characters I controlled all at once that I took into some fairly high level areas, with a warrior as the tank. After he got defensive trance he literally never did anything in combat except for that, since it's a task and it offers such a massive survivability increase that you'd be crazy not to use when tanking stuff above level 60 or so. 

So unless you're concerned about lower level warriors using deathstroke/etc. while tanking or are thinking to change how defensive trance works it actually might not be an issue at all. 

---

On a related note, though, I've had some ideas regarding warrior tanking for a while now:

Warrior tanking is super boring and annoying because of the reliance on defensive trance and how trance works. About the only input you make as the tank after typing "trance" is to stop the trance if you need to rescue someone. Plus mobs will constantly try to use things like hurl on you if you're trancing which, from my experience fighting lake elves who apparently start practicing hurl at birth, does not make for a very enjoyable tanking experience.

In that vein, I've always thought turning the brawling discipline into a sort of hybrid support/tank discipline could be really cool and give warriors some very unique flavor. 

Basically the idea would be to make grapple the focal skill, with how it works changing based on whether the warrior is tanking or not. Tank grappling would functionally just be "in close proximity, maneuvering their and the enemy's position to create openings for groupmates and disadvantages for the enemy" type of grappling, whereas non-tank grappling would be more of a straight-up "put the enemy in a headlock from behind while your teammates stab them repeatedly" or "dive at the enemy's legs and go for a takedown" scenario. I think it could open up a ton of potential for the combo stuff you guys have mentioned several times.

At the same time, defensive trance could be turned into a passive that's automatically active when certain conditions are met 
 - maybe add stances the warrior can switch between, each one giving a unique, permanent affect while active that skills can check to see if they're available?. The new passive trance could replace and act as a combination of parry/riposte/dodge/countermove/shield block while also reducing the warrior's # of attacks and/or raw weapon attack damage and maybe increasing command lockout time. That would allow a tanking warrior to still effectively mitigate physical damage while using active, tank-related skills to do things like debuff the enemy or set up openings for groupmates to use group-oriented skills that could be added, and could make the tanking experience much more active and reactionary instead of just sitting there in trance waiting to either be hurled, lose concentration, or for the fight to end.

As far as why the brawling disc - I always envision the way Aragorn fights in the Lord of the Rings movies as pretty emblematic of how a warrior in a fantasy game like Sneezy would probably do it. He's not some wannabe fencer who does everything with fancy sword moves but rather often uses the sword as a tool to create openings for him to get in closer and throw a punch or move to a better position. He's truly brawling more than swordfighting and because he's up in the enemy's face they're definitely going to focus their attention on him. 

Then soldiering could be the "advanced offensive fighting" disc, since soldiers would be well-practiced in attacking enemies as part of a group but would've also drilled extensively in one-on-one combat. Things like advanced berserking could fit in well there too, as berserkers are often portrayed as some sort of shock troop in an army.

Then the stuff in the dueling disc could probably be split up between brawling and soldiering, just to remove clutter.

Anyway, I'll just leave it at that, since I could go on and on about this haha. But I'm definitely going to try to devote some of my sadly very limited spare time at the moment to messing around with these ideas in my dev copy I've got running.

wershlak:
`sizable penalty to the player's armor which a tank would definitely not want.` Tanks are above the armor cap. A fully buffed warrior in diamond plate probably will not lose any defense with a 300 armor penalty. It's not an effective penalty.

Bless is +10 spell hitroll with equals +1 hit

parkscs:
Fair enough.  I assumed this wasn't an issue any more after we nerfed the
buffs recently.  If people are 300 above the cap, seems like buffs are too
high still?  We could either tweak the number (if people are 300 above the
cap, make the penalty 600 for example) or just go a different direction.
Initially it was the same penalty as smite but steeper, but I opted for
armor instead just to give it some more unique flavor.  I don't really care
what values we assign to it, it needs to be balanced, but I do want there
to be a sizable penalty associated with it and it should punish any tanks
who decide to use the ability against a difficult opponent.

Regarding warrior tanking, I agree with everything you've said.  Trance is
good but kinda boring - you just recast it when it fails and it's your
group who has to keep up with maintaining your movement points.  Right now
tanking isn't really a big issue as players aren't taking enough damage at
end-game (provided spellcasters can't cast) but it would be nice to give
warriors a creative tanking spec and then adjust the overall balance to
give it a role in the game.  I know Warluk has thrown out the idea of doing
something with a block value for warriors.  I've been toying with the idea
of an ability that gives you a lot of defense for say 3-5 rounds and then
requires 10-15 rounds worth of time before you can use it again.  You would
just spam it on cooldown right now, but that opens the door to some more
interesting boss battles in the future where they do bursts of damage.  I
have a better idea of where I'd like to go with warriors offensively though
(high risk / high reward and probably a lot of warrior deaths as a result)
than defense though.

parkscs:
Since a lot of changes have been committed to this branch (due to yours truly not fully understanding branches), I wanted to clean some things up today and do an updated post.  I believe this update addresses all of the comments raised above, but please let me know if something is still off and I'll be happy to look at it.  Since the last update, I've added the following tweaks:

- **Deathstroke rework** - I've updated this ability to deal additional damage that scales with the victim's missing HP.  The skill also grants a damroll buff on successful hits and always grants an armor penalty to discourage tanks from using the ability.  The skill now follows the framework we discussed (and that the stab ability currently implements) for a skillCheck, a hitCheck, and a critFail check.  I'm not entirely happy with how this ability lands at end-game (I went 0/5 against staley) but more testing is needed and this hit chance can be remedied in the future when we address other skills failing to land.
- **Bloodlust (new ability)** - This is a new ability in the base warrior kit.  Warriors can gain 1 stack per killing blow up to a maximum of 10 stacks.  Each stack grants +1 damroll and a bonus to regen (up to 166% of base regen) which should help warriors with their complete lack of a way to heal (aside from a bedroll) and can also help with them running out of vitality (which is very easy to do atm with enough speed)
- **Advanced berserking (new ability)** - This is a new ability in the brawling discipline which has a few effects that kick in only while berserking.  First off, the extra attacks granted by berserking are increased by an additional 0.5 (up to a total of +1) with advanced berserking.  Secondly, there's an attack bonus similar to the one granted by the offensive attack mode that is granted while the character is berserking.  And finally, there's a chance to gain a stack of bloodlust each round of combat if you are berserking.  While that means there's a lot of power baked into berserk, I would ask that you keep in mind that you can't flee while berserking, you're aggro while berserking, and you can't cast while berserking (which doesn't matter for single class warriors but is a big deal for multi-class).  I've also submitted a change so that you can no longer enter portals while berserking.  In short, I want there to be risk and limitations involved in using this ability, in exchange for the combat bonuses it provides.

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -35,96 +41,115 @@ static int deathstroke(TBeing *caster, TBeing *victim)
...
+              FALSE, caster, 0, victim, TO_CHAR);
         act("You take advantage of $n's vulnerability for a cheap shot!", 
-               FALSE, caster, 0, victim, TO_VICT);
-
-        dam = 3*victim->GetMaxLevel();
-        dam += 3*victim->plotStat(STAT_CURRENT, STAT_STR, 0, 6, 3);
-        dam = ::number(1, dam);
-        dam = victim->getActualDamage(victim, NULL, dam, SKILL_DEATHSTROKE);
-        dam /= 2;
+             FALSE, caster, 0, victim, TO_VICT);
```
drewhershey, line 152: If you're still adding up the `percent` value here and passing it to `bSuccess` doesn't that mean you're giving the victim bonuses when they attempt their counter-deathstroke? Or am I reading this wrong?

code/code/cmd/cmd_deathstroke.cc
```diff
@@ -35,96 +41,115 @@ static int deathstroke(TBeing *caster, TBeing *victim)
...
-  }
-  if (victim->isFlying() && (victim->fight() != caster)) {
-    caster->sendTo("You can only deathstroke fliers that are fighting you.\n\r");
-    return FALSE;
-  }
   if (!caster->heldInPrimHand() || 
       !(tw=dynamic_cast<TBaseWeapon *>(caster->heldInPrimHand()))) {
     caster->sendTo("You need to hold a weapon in your primary hand to make this a success.\n\r");
     return FALSE;
   }
```
drewhershey, line 48: Not a big deal, but if you wanted to you could move these hard-fail checks (lines 15-50) into the `doDeathstroke` function to separate concerns a little more. I _think_ that was how they intended it to work - `doSkillName` functions perform the hard fail checks then call the `skillName` functions that actually contains the logic for the skills themselves. But it seems like most of the previous coders ended up splitting hard-fail checks into both functions fairly randomly.

parkscs:
I didn't rework the counter-attack at this point, mainly because I want to
think on it a bit more.  My concern with following the same skillSuccess(),
skillHit(), !critFail() algorithm is that I suspect that would make
counter-attacks only land for warrior NPC's.  I want to rethink the counter
attack algorithm a bit more before making changes, and so I just left it
alone for now.

parkscs:
Makes sense.  That's something I can look to refactor in the next iteration
(along with the counter-attack algorithm).

wershlak:
@parkscs Where's this at? I see there's a conflict. Do you have more changes to make? Do you need help with anything?

parkscs:
Thought it was ready for review - will try and look to see about the
conflict, wasn't aware.

parkscs:
Resolved the conflict (I think), let me know what you think.

code/code/misc/equip.cc
```diff
@@ -1058,8 +1058,8 @@ int TBeing::wear(TObj *o, wearKeyT keyword, TBeing *ch)
               --(*o);
               equipChar(o, HOLD_RIGHT);
               // learn 2h spec as with dual wield
-              if (doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
-                learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_2H_SPEC_DEIKHAN, 60);
+              if (doesKnowSkill(SKILL_2H_SPEC)) {
```
wershlak, line 1061: I see you updated getSkillNum() but aren't using it anywhere

code/code/misc/limits.cc
```diff
@@ -467,6 +467,22 @@ int TPerson::hitGain()
   if (affectedBySpell(SPELL_ENLIVEN))
     gain *= 2;
 
+  // Scale regen based on the number of bloodlust stacks 
+  if (affectedBySpell(SKILL_BLOODLUST)) {
+    int mod = 1;
```
wershlak, line 472: Why not zero?

code/code/misc/limits.cc
```diff
@@ -467,6 +467,22 @@ int TPerson::hitGain()
...
+
+    for (ch_affected = affected; ch_affected; ch_affected = ch_affected->next) {
+      if (ch_affected->type == SKILL_BLOODLUST) {
+        // set the mod and remove the affect so we can add it fresh
+	mod += ch_affected->modifier;
+	break;
+      }
+    }
+
+    gain *= 1+(mod/15);
```
wershlak, line 483: Gain and mod are integers so mod/15 is always going to be zero

code/code/misc/combat.cc
```diff
@@ -1992,9 +1994,9 @@ int TBeing::getWeaponDam(const TBeing *v, const TThing *wielded, primaryTypeT is
     // 2h Specialization: should be no additional dam for anyone without specialization
     // extra 10% for 2h is part of the object
     TObj *obj = dynamic_cast<TObj *>(heldInPrimHand());
-    if (obj && obj->isPaired() && doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
+    if (obj && obj->isPaired() && doesKnowSkill(SKILL_2H_SPEC)) {
```
wershlak, line 1997: Need getSkillNum here too

wershlak (changes requested):
We need help files for new skills.

Also, I don't understand bloodlust. When should you be gaining it and what does it do for you and when?

I'm looking forward to trying a new death stroke but here's what I know:
- the game is too easy
- auto attack is probably 90% of warrior damage

You've added an additional attack per round with advanced berserk and a ton of extra auto attack damage. I'd love to see something different besides more auto attacking and more damage/hit for auto attack. Like if bloodlust decreased command lockout or increased bash damage 3x or something.

code/code/misc/equip.cc
parkscs: Updated.

code/code/misc/limits.cc
parkscs: Probably should be in hindsight.  Updated.

code/code/misc/limits.cc
parkscs: Updated again (and tested to ensure it's performing as desired).

code/code/misc/equip.cc
parkscs: And just a note - please confirm this update is using the functions correctly.   I might just be missing something here.

code/code/misc/combat.cc
parkscs: Updated - hope this is the correct usage.  Please let me know if I'm missing something.

parkscs:
> We need help files for new skills.

Happy to write them but need to look into how the help system actually works.  I'll get descriptions written up at a minimum today.

> Also, I don't understand bloodlust. When should you be gaining it and what does it do for you and when?

You gain it in two ways.  The basic way is when killing enemies, each killing blow = 1 stack up to a maximum of 10 stacks.  Each stack gives +1 damroll and a bit of regen with an enliven type effect.  The alternate way to gain stacks is there's a chance once/round while berserking with advanced berserking to gain a stack; this is intended to further encourage the use of berserking and will help berserkers regen movement points (as they cannot cast/flee and drain more stamina attacking).  The goal is to provide a mechanic that provides a bit of a combat bonus and also helps warriors with their complete lack of regen, but it also requires that players find a way to keep the effect going.  After testing last night I upped the duration slightly and removed the scaling with player level (as it was falling off too quickly), but my goal is to make it challenging to keep this active (particularly at max stats).  This is in contrast this with passive, always active abilities that don't really require any player skill to get the most out of them, and goes to your first point of the "game is too easy."

> auto attack is probably 90% of warrior damage
> You've added an additional attack per round with advanced berserk and a ton of extra auto attack damage. I'd love to see > something different besides more auto attacking and more damage/hit for auto attack. Like if bloodlust decreased command > lockout or increased bash damage 3x or something.

Auto-attack is all warriors have for damage at the moment because their kit is screwed up.  This is phase 1 in reworking that kit, and deathstroke alone will change that percentage by quite a bit.  The advanced berserking changes are designed to make berserking more attractive to players; currently it's very rarely used and the only players I've seen use it are twinks with maximum gear in very safe situations.  My goal is to add a bit more rewards to offset the risk/reward equation, but keeping all of the risk (and adding some by nerfing portal access during combat) so that warriors want to use berserk but have to think about when they can use it safely.  I would not look at this skill as achieving the goal of making the class more active to play, as that's not my intent with this skill.  The intent is to make an existing skill more attractive to players to use, keeping in mind that this is a skill that comes with a lot of drawbacks/risks.

As for the suggestions on bloodlust, that's more of a mechanic that I think would be good for the combo system but that's not really what I am wanting to do with bloodlust.  As for things like bash doing 3x damage though, I've talked about this with some folks but I'll just put my thoughts here... I do not want the traditional warrior skills like headbutt, bash, trip, etc. to become the "main damage abilities" for the class because, well, that's fairly silly.  If I'm hitting a mob with a giant 2h axe, why would me tripping them or kneestriking them do more damage than hitting them with a weapon?  I agree with the overall design concept of making skills play much more of a role in combat, but I think there are different ways to achieve this beyond just buffing the existing warrior toolkit.  As it stands, I'm envisioning the existing warriors skills doing a bit more damage but really being spammable abilities that tie into a combo system (likely phase 3).  I'm then looking at adding a handful of additional warrior abilities that would be larger sources of skill-damage but they would come with lockouts and flesh out the overall kit.  However, that's more for phase 2+.  

I'll start writing up my plans for phase 2/phase 3 soon so we can get concrete plans and feedback posted on here, but as for these changes, I'd just say that these aren't intended to address all of the concerns for warriors and it's just phase 1 of likely 3+ phases of changes.

code/code/misc/equip.cc
drewhershey: So I think what you need here is to first get a reference to which version of 2H Spec the character in question uses, by doing something like `auto skill2HSpec = getSkillNum(SKILL_2H_SPEC)`. Then you would pass that result to the `doesKnowSkill` function (`getSkillNum(skill2HSpec)`) instead of passing `SKILL_2H_SPEC`. Or you could just do `doesKnowSkill(getSkillNum(SKILL_2H_SPEC))` if you don't need to reference the skill again anywhere later on in the code.

It's necessary here because with the way you have it coded currently a Deikhan will never hone 2H spec (I think), since for them `doesKnowSkill(SKILL_2H_SPEC)` will always return false as they don't have SKILL_2H_SPEC - they have SKILL_2H_SPEC_DEIKHAN. So `getSkillNum` when called by a Deikhan will return the skill number for SKILL_2H_SPEC_DEIKHAN instead, which will then return `true` for the `doesKnowSkill` call and allow them to hone their version of the skill.

You can tell which skills you need to use `getSkillNum` for by just looking at the code for that function. Sorry if you already knew all this, I just figured it out myself like yesterday so thought I'd share in case it helps.

code/code/misc/equip.cc
```diff
@@ -1058,8 +1058,8 @@ int TBeing::wear(TObj *o, wearKeyT keyword, TBeing *ch)
               --(*o);
               equipChar(o, HOLD_RIGHT);
               // learn 2h spec as with dual wield
-              if (doesKnowSkill(SKILL_2H_SPEC_DEIKHAN)) {
-                learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_2H_SPEC_DEIKHAN, 60);
+              if (doesKnowSkill(getSkillNum(SKILL_2H_SPEC))) {
+                learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_2H_SPEC, 60);
```
drewhershey, line 1062: I think this is another situation where you'll need to use the result of `getSkillNum(SKILL_2H_SPEC)` in the `learnFromDoingUnusual` call instead of just SKILL_2H_SPEC.

The doesKnowSkill call will successfully make sure that the character knows either SKILL_2H_SPEC (the warrior version) or SKILL_2H_SPEC_DEIKHAN, but will then only actually work for chars who have the warrior version, as coded.

code/code/misc/combat.cc
drewhershey: Just another comment to point out that the fix here is the same for the other places I mentioned - replacing SKILL_2H_SPEC with `getSkillNum(SKILL_2H_SPEC)`.

code/code/misc/combat.cc
```diff
@@ -2101,6 +2103,56 @@ int TBeing::hit(TBeing *target, int pulse)
...
+      // This will eventually grant a chance to land a random warrior ability automatically while berserking.
+      // The abilities currently planned are:
+      //   - headbutt (1)
+      //   - kneestrike (2)
+      //   - bash (3)
+      //   - bodyslam (4)
+      //   - spin (5)
+      //   - stomp (situationally - opponent needs to be downed) (6)
+      //   - taunt (not particularly advantageous) (7)
+      /*
```
cizra, line 2116: Could we  have no dead code please

code/code/misc/combat.cc
```diff
@@ -2291,13 +2343,13 @@ int TBeing::hit(TBeing *target, int pulse)
...
 	learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_IRON_FIST, 15);
-
-      if(doesKnowSkill(SKILL_CRIT_HIT))
-        learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_CRIT_HIT, 15);
     }
+
     if (((fx > 0.999) || (fy > 0.999))){
       if(doesKnowSkill(SKILL_POWERMOVE))
 	learnFromDoingUnusual(LEARN_UNUSUAL_NORM_LEARN, SKILL_POWERMOVE, 15);
+      if(doesKnowSkill(SKILL_CRIT_HIT))
```
cizra, line 2351: inconsistent whitespace

cizra (changes requested):

parkscs:
Closing this out in favor of the updated pull request for warrior_changes_2.  I've gone through and incorporated the feedback from this PR in the new branch.

PR #352, opened 2021-07-25
Update SConstruct to fix clang compilation and add new sanitizers for gcc

drewhershey:
Out of curiosity I started testing out compiling in Linux (Ubuntu 21.04/Hirsute) with clang instead of gcc, as I saw the SConstruct file was set up to allow either.

I had to make some changes to the file for compilation with clang to succeed. First, when sanitize=true, moving the code that appends the asan library up above where it appends the ubsan library as Address Sanitizer needs to be the first library in the list, and secondly changing the cxxflags/linkflags to remove a bunch of link errors that occur otherwise.

Clang seems to find more/different errors and warnings than gcc so it might be worth testing. It also appears that it was using extra sanitizers that gcc wasn't: an undefined behavior sanitizer and integer sanitizer. The benefit of using the undefined behavior sanitizer appears to be that it catches some runtime errors that wouldn't be apparent otherwise - these show up in the logs as the game is running. They can also lead to new and interesting seg faults!

Disclaimer: I tested these changes in Ubuntu running inside WSL 2. I haven't tested them with the Sneezy Docker image, but since that (I believe) uses Ubuntu as well it seems like the changes should be compatible.

### DIFF
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -186,9 +186,10 @@ if flag_debug:
             # undefined behavior sanitizer, apple clang unsupported
             env.Append(CXXFLAGS='-fsanitize=undefined')
             env.Append(extra_libs=['ubsan']) 
-            # integer sanitizer, apple clang unsupported
-            env.Append(CXXFLAGS='-fsanitize=integer')
-        if (clang_version and clang_version >= (3, 9) or gcc_version >= (7, 0)):
+            if (clang_version):
+                # integer sanitizer, g++ and apple clang unsupported
+                env.Append(CXXFLAGS='-fsanitize=integer')
+        if (clang_version and clang_version >= (3, 9) or gcc_version and gcc_version >= (7, 0)):
             env.Append(CXXFLAGS='-fsanitize-address-use-after-scope') 
 
 env.Append(extra_libs=[opt_libs])
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -72,7 +72,7 @@ need good crash debugging).
 """)
 
 
-flag_sanitize = get_bool_option('sanitize',      True)
+flag_sanitize = get_bool_option('sanitize',    True)
 flag_ccache =   get_bool_option('ccache',      False)
 flag_debug =    get_bool_option('debug',       True)
 flag_gprof =    get_bool_option('gprof',       False)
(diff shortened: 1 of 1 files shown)


code/SConstruct
```diff
@@ -177,24 +177,27 @@ if flag_debug:
...
     env.Append(CXXFLAGS='-fno-omit-frame-pointer')
-    # undefined behavior sanitizer, apple clang unsupported
-    if clang_version and clang_version >= (3, 9):
-        env.Append(CXXFLAGS='-fsanitize=undefined')
-        env.Append(extra_libs=['ubsan'])
-    # memory error sanitizer
-    if flag_sanitize:
+    if flag_sanitize:   
+        # memory error sanitizer
         env.Append(CXXFLAGS='-fsanitize=address')
```
cizra, line 182: Can you have `-fsanitize=address` without `-lasan`?

drewhershey: This is how it was written before my changes, I just rearranged the structure to reduce some repetition and simplify the `if` flow. Or at least that was my intention.

From the "native clang doesn't need explicit asan lib" comment, I assumed having the `env.Append(CXXFLAGS='-fsanitize=address')` line on its own was related to Apple clang support. I don't use Apple myself so have no way to test it.

My thought process was that if someone were compiling using native Apple clang with flag_sanitize = true, they'd get the `env.Append(CXXFLAGS='-fsanitize=address')` line by default, wouldn't need the `env.Append(extra_libs=['asan'])` line per the comment, and wouldn't be compatible with the rest of the stuff inside that branch anyway.

Anyone compiling from a non-Apple OS would get the same `env.Append(CXXFLAGS='-fsanitize=address')` line by default, and would be guaranteed to have either gcc_version or clang_version and clang_version >= (3, 9) so they'd get the `env.Append(extra_libs=['asan'])` line added in for sure plus whatever other stuff evaluated true.

Any preference on how you'd like me to fix it? Maybe just move the `env.Append(CXXFLAGS='-fsanitize=address')` inside the first `if` statement with the other Appends?

cizra: Nobody's tried to run Sneezy on a Mac for many years (to my knowledge) and whoever would try that would probably use Docker anyway, so maybe it's not very important to preserve Mac compatibility.

cizra (approved):

drewhershey:
Closing this pull request for the moment - even though it should pass the CI tests successfully now I don't think it's ready to merge. UBSan finds some new seg faults at runtime when compiling with g++ and sanitize = true so I'm going to push some commits that address those at the same time.

PR #353, opened 2021-07-26
Fix Plague of Locusts prayer

drewhershey:
Plague of locusts has been broken forever - literally, I remember trying
to use it back in the early 2000s and wondering why it wasn't doing
anything. Feeling personally attacked by this memory, I've fixed it and
it seems to be working as it should now.

Also took it upon myself to refactor it a bit just to clean stuff up and
make the code more descriptive/flow better.

The part that actually fixed the spell was lines 122-126. The swarm was
being summoned successfully but was never actually being made to
attack the victim, and the locusts mob is written to dissipate as soon as
it's no longer involved in a fight - which, before now, was immediately
after being summoned.


code/code/disc/disc_wrath.cc
```diff
@@ -15,103 +15,102 @@
...
   if (Weather::getWeather(*caster->roomp) == Weather::RAINY) {
     act("Nothing seems to happen.", FALSE, caster, NULL, NULL, TO_ROOM);
     caster->sendTo("How do you expect to summon locusts in the rain!\n\r");
     return SPELL_FAIL;
   }
+
+  if ((caster->followers) && (level < GOD_LEVEL1))  {
+    followData *currentFollower, *nextFollower;
+    for (currentFollower = caster->followers; currentFollower; currentFollower = nextFollower) {
+      nextFollower = currentFollower->next;
```
cizra, line 33: Why do you need that variable anyway?

code/code/disc/disc_wrath.cc
```diff
@@ -15,103 +15,102 @@
...
   }
+
   if (Weather::getWeather(*caster->roomp) == Weather::RAINY) {
     act("Nothing seems to happen.", FALSE, caster, NULL, NULL, TO_ROOM);
     caster->sendTo("How do you expect to summon locusts in the rain!\n\r");
     return SPELL_FAIL;
   }
+
+  if ((caster->followers) && (level < GOD_LEVEL1))  {
+    followData *currentFollower, *nextFollower;
```
cizra, line 31: Why not `auto`?

code/code/disc/disc_wrath.cc
```diff
@@ -15,103 +15,102 @@
...
-        delete locusts;
-        locusts=NULL;
-        return SPELL_FAIL;
-      }
-    }
-  }
+  }  
 
-  level = min(level, 75);
+  float normalizedLevel = max(1.0, min(level, 75) / 2.0);
```
cizra, line 63: `min(level, 75)` is always `level`, since our levels are <= 60.

code/code/disc/disc_wrath.cc
```diff
@@ -15,103 +15,102 @@
...
+        normalizedLevel *= 2;
         break;
       case CRIT_S_NONE:
         if (victim->isLucky(caster->spellLuckModifier(SPELL_PLAGUE_LOCUSTS))) {
           SV(SPELL_PLAGUE_LOCUSTS);
-          lvl /= 2;
+          act("Luckily for $N, only a small swarm responds to $n's prayer!", TRUE, caster, 0, victim, TO_NOTVICT);
+          act("Luckily for you, only a small swarm responds to $n's prayer!", TRUE, caster, 0, victim, TO_VICT);
+          act("Unfortunately, only a small swarm responds to your prayer.", TRUE, caster, 0, victim, TO_CHAR);
+          normalizedLevel /= 2;
```
wershlak, line 91: I know you are just preserving the logic in place but we calculate a level for the mob then double it if it's a critical and half it if it's not?

I'd honestly just remove this line. If the plague is too strong we can adjust later :)

code/code/disc/disc_wrath.cc
```diff
@@ -120,21 +119,28 @@ int plagueOfLocusts(TBeing *caster, TBeing *victim, int level, short bKnown)
...
-    locusts->setLevel(WARRIOR_LEVEL_IND, (byte)lvl);
-    locusts->setHPLevel(lvl);
+    locusts->setLevel(WARRIOR_LEVEL_IND, (byte)normalizedLevel);
+    locusts->setHPLevel(normalizedLevel);
     locusts->setHPFromHPLevel();
-    locusts->setACLevel(lvl);
+    locusts->setACLevel(normalizedLevel);
     locusts->setACFromACLevel();
-    locusts->setDamLevel(lvl);
+    locusts->setDamLevel(normalizedLevel);
```
cizra, line 134: Why the duplicate code? Looks like it could be extracted from these branches.

cizra (changes requested):

code/code/disc/disc_wrath.cc
wershlak: `level = caster->getSkillLevel(SPELL_PLAGUE_LOCUSTS);`

It's the learnedness or the level specified on the item if it's a wand or something.

cizra: Ooh. The variable name could be more explicit, then.

drewhershey: I believe this code gets called when a mob uses this prayer as well, but I'd have to confirm that. But if so, is it possible this code is there to limit the strength when used by very high level cleric mobs? 

I wasn't entirely sure why they coded it this way originally - my primary intention behind this specific change was just to avoid mutating the level parameter in that way, since the result was only used in a single place in the code afterwards.

code/code/disc/disc_wrath.cc
drewhershey: Sounds good, I've never been a fan of this sort of crit fail/success code myself.

code/code/disc/disc_wrath.cc
drewhershey: Honestly not sure - I preserved the original logic and just renamed the variables to be more descriptive rather than single letters. I'll look at it further and try to simplify it.

code/code/disc/disc_wrath.cc
drewhershey: I had the same thought then just forgot about it. I'll take another look at it.

code/code/disc/disc_wrath.cc
drewhershey: Good question - fixed in upcoming commit.

code/code/disc/disc_wrath.cc
drewhershey: I remember now why I left it - it was because of the crit success/isLucky saving throw code that @wershlak mentioned above as being unnecessary. On prayer success there was a chance that the locusts' level would be doubled or halved, so we needed to apply those modifiers in each branch.

code/code/disc/disc_wrath.cc
drewhershey: I forced a lvl 89 cleric mob to cast plague of locusts on another mob and confirmed in gdb that this code triggered from the mob prayer and forced its level down to 37.5. Still not really sure what the original intent was here.

@cizra How would you like me to proceed?

drewhershey: At this point I'm just getting rid of this level adjustment. It doesn't seem to serve much purpose except to gimp a spell that's really not that powerful in the first place. If you guys object I'm happy to change it back, though.

wershlak: `min(level, 75)` When this is called as I posted it's the learnedness of the spell. Not the level of the caster. Looks like the intent is for level to be capped at 37.5 but even without the min() statement the maximum level that could be passed into the function is 100 which would be divided by 2.

If you want to rework it then you should probably rework what's passed in so we can have high level mobs summon terrible demon locusts that breath fire on puny mortals.

If you just leave level that's passed in as the level of the mobs we are going to have some bad assed clerics walking around.

drewhershey: After further testing I'm realizing it's probably stupid to allow clerics to easily summon a mob the same level as themselves that also has fairly strong damage resistance. So I'm going to ask for some feedback in #balance on Discord.

code/code/disc/disc_wrath.cc
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/353/commits/1067a974875a7cb81742309e23a76e9ce279cd6b

code/code/disc/disc_wrath.cc
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/353/commits/e678035f48a6a98e660100b84771f378f1a8a648

code/code/disc/disc_wrath.cc
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/353/commits/e678035f48a6a98e660100b84771f378f1a8a648

code/code/disc/disc_wrath.cc
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/353/commits/bb6987a12f3deb50ecb985e9095d18fe80652d58, though I'm not sure why the check is failing for that commit

code/code/disc/disc_wrath.cc
drewhershey: Addressed in https://github.com/sneezymud/sneezymud/pull/353/commits/bb6987a12f3deb50ecb985e9095d18fe80652d58

code/code/disc/disc_wrath.cc
```diff
@@ -13,159 +13,155 @@
...
-        delete locusts;
-        locusts=NULL;
+  if ((caster->followers) && (level < GOD_LEVEL1))  {    
+    for (auto currentFollower = caster->followers; currentFollower; currentFollower = currentFollower->next) {      
+      if (isname("locust", currentFollower->follower->name)) {
+        act("Nothing seems to happen.", false, caster, nullptr, nullptr, TO_ROOM);
+        caster->sendTo("You would have to be a god to summon another locust swarm!\n\r");        
         return SPELL_FAIL;
-      }
+      }
```
cizra, line 35: Let's remove those trailing spaces

cizra (approved):

code/code/disc/disc_wrath.cc
drewhershey: I'm going to run the code through VS Code's C++ auto-formatter that I've been afraid to use before now because it would generate so many file changes. Get rid of all the extraneous whitespace and other hard-to-see formatting things at once.

Let me know if it does anything in a way you'd prefer done differently - I can customize pretty much everything about it.

drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/353/commits/d4b339ee3f5aa092aff3194c561eb00b2806589c.

Glad you caught that, there was a lot more than just that one spot.

PR #354, opened 2021-07-26
Add specialAttack overload and refactor bash skill to utilize it

drewhershey:
Submitting this as a draft pull request just to get feedback and start discussion on whether others feel this is a good direction to go as far as modifying how skill success works. Used warrior bash as the test skill.

First commit adds an overload for the TBeing::specialAttack method that allows a situational modifier to be passed in. The modifier is then added to the result of the attackRound() calculation within to effectively adjust the attacker's level during the context of this individual skill attempt. This is a cleaner/more meaningful/more logical way to adjust the success chances of each individual skill attempt based on its circumstances at the time of execution, compared to modifying the bKnown value that gets passed into the bSuccess call like many skills currently do. See comments on that commit for more detail.

Second commit refactors the warrior bash skill to demonstrate how situational modifiers could be calculated and used with the specialAttack overload. I also just updated the code in various ways that seemed useful or practical, including formatting.

I've tested the code out in my dev build and it all works as expected. Bash was able to land noticeably more often on high level opponents without making it feel easy, which was a huge improvement from before where it often felt almost pointless to even attempt it against higher level enemies. And there are still plenty of further situational considerations that could be coded for to fine-tune the success chance even further.


code/code/cmd/cmd_bash.cc
```diff
@@ -62,15 +63,11 @@ bool TBeing::canBash(TBeing *victim, silentTypeT silent)
...
-    if (!silent)
-      sendTo("You can't bash while flying if your victim isn't flying.\n\r");
-    return FALSE;
-  }
+
   if (checkPeaceful("You feel too peaceful to contemplate violence.\n\r"))
     return FALSE;
 
-  if (eitherLegHurt()) {
+  if (eitherLegHurt() && !isFlying()) {
```
drewkht, line 70: Figured you should still be able to bash into someone even if you have a damaged leg if you're flying

code/code/cmd/cmd_bash.cc
```diff
@@ -94,162 +93,216 @@ bool TBeing::canBash(TBeing *victim, silentTypeT silent)
...
     return FALSE;
   }
+
   if (!victim->hasLegs()) {
     if (!silent)
       sendTo("You can't knock them over, they have no legs.\n\r");
     return FALSE;
   }
 
+  if ((isSwimming() || victim->isSwimming())) {
```
drewkht, line 127: Moved this check into canBash() just to be consistent. I was considering making bashing while swimming for aquatic races (so really only Kalysians, as far as PC races go) but wasn't sure it was worth it, since it's such an edge case and would require updating the crashLanding method as well.

code/code/cmd/cmd_bash.cc
```diff
@@ -94,162 +93,216 @@ bool TBeing::canBash(TBeing *victim, silentTypeT silent)
...
-    c->reconcileDamage(victim, 0, skill);
-    // return true so the lag and stuff
+  /*
+    If attacker is not tanking, pass agility check to successfully get into 
+    position for a bash attempt. Makes bashing while not tanking more difficult
+    but still possible.
+  */
+  if (victim->fight() && victim->fight() != attacker && !attacker->isAgile(0)) {
+    attacker->sendTo("You try to line up a bash but can't find a good angle!\n\r");
+    // Still add skill lag, as with deikhan charge
```
drewkht, line 191: Changed it so that bashing someone who's fighting but that you're not actively tanking requires passing an agility check to "line up" the bash amidst the chaos. Was aiming for something similar to how Deikhan charge works.

I think these sorts of stat checks have a lot of potential throughout the codebase, but they're not used super often as is.

wershlak:
So far I like the approach and it looks clean. I want to look at the formulas a bit more because I think we might want to adjust how these work and I'm not sure of the best way to refactor them.

wershlak (commented):
I'm not sure this goes far enough. where you call hits() at the end of your special attack overload we might need to have a specialAttackHits() that has a formula that is more gentle with level differences.

I think we could merge this and see how it plays. The code is now cleaner and better

code/code/cmd/cmd_bash.cc
```diff
@@ -94,162 +93,216 @@ bool TBeing::canBash(TBeing *victim, silentTypeT silent)
...
-  if ((learning = c->getAdvLearning(skill)) <= 10) {
-    if ((obj = c->heldInSecHand()) &&
-        (tbc = dynamic_cast<TBaseClothing *>(obj)) &&
-        tbc->isShield()) {
-      shieldWtBonus=(tbc->getWeight() / 5) + 1;
-    } else {
-      c->sendTo("You are not skilled enough to bash without a shield!\n\r");
-      return FALSE;
+static int bash(TBeing *attacker, TBeing *victim, spellNumT skill) {   
+  int brawlingLearnedness = attacker->getAdvLearning(skill);
```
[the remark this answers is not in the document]
drewhershey: Addressed your first point in https://github.com/sneezymud/sneezymud/pull/354/commits/fbe18fe09f9c0e56f8014430f5e1cad41217f537. 

From looking at `getAdvLearning()` it does initialize `ret` to -1, but then `ret` gets set to the value of `getLearnedness()` before it's ever returned. So it doesn't seem there's any way it'll return -1, as the property that `getLearnedness` returns gets initialized to 0 when its instance of CDiscipline is instantiated.

wershlak (approved):
Thanks for making the code more deikhan friendly. :P

PR #355, opened 2021-07-27
Fix null deref that can cause seg fault

drewhershey:
Reported as a runtime error by ubsan and caused a seg fault due to
null dereference in my test build.

### DIFF
--- a/code/code/sys/db.cc
+++ b/code/code/sys/db.cc
@@ -3437,14 +3437,14 @@ void runResetCmdZ(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload, 
 
 void runResetCmdY(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload, TMonster *&mob, bool &objload, TObj *&obj, bool &last_cmd)
 {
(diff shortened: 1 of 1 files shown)


code/code/sys/db.cc
```diff
@@ -3437,14 +3437,14 @@ void runResetCmdZ(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload,
 
 void runResetCmdY(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload, TMonster *&mob, bool &objload, TObj *&obj, bool &last_cmd)
 {
+  if (!mob || !mobload)
+    return; // log error here?
```
cizra, line 3441: Yes! Please do! Along with other relevant info, like zone name, etc.

cizra (changes requested):

drewhershey:
This seg fault was found by running UBSan when compiling with g++, so I'm going to close this PR and merge this change with the PR that updates the SConstruct file to use UBSan in the first place. I'll address @cizra 's change request above at the same time.

PR #357, opened 2021-07-28
Add clang-format test

wershlak:
Adding a clang-format test.

I formatted one file as an example. We can change the config to suit our needs and I can format all the files once we agree on the right formatting.

NOTE: unlike the uncrustify check this check is only checking the git diff so only LINES which are updated are checked. IE. this is safe to merge.

### DIFF
--- a/code/code/cmd/cmd_attribute.cc
+++ b/code/code/cmd/cmd_attribute.cc
@@ -8,34 +8,29 @@
 
 #include <stdio.h>
 
-#include "being.h"
 #include "account.h"
 #include "ansi.h"
-#include "extern.h"
-#include "handler.h"
-#include "guild.h"
+#include "being.h"
 #include "configuration.h"
-#include "weather.h"
+#include "extern.h"
 #include "gametime.h"
+#include "guild.h"
+#include "handler.h"
+#include "weather.h"
 
-static void showStatsTo(const Descriptor *d, const TBeing *ch, bool hidden_stuff)
-{
+static void showStatsTo(const Descriptor *d, const TBeing *ch, bool hidden_stuff) {
   time_info_data playing_time;
   char time_buf[160];
   char buffer[512];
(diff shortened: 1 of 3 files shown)


code/code/cmd/cmd_attribute.cc
```diff
@@ -496,36 +541,36 @@ const char *statBonusDescr(int percent)
...
-    return;
+  if (!desc) return;
 
   one_argument(arg, cmdbuf);
   if (cmdbuf.empty()) {
-    sendTo("Syntax: attribute <personal | condition | statistics | reset>\n\r");
+    sendTo(
+        "Syntax: attribute <personal | condition | statistics | "
+        "reset>\n\r");
     return;
```
drewhershey, line 560: This is kinda weird looking - wonder what rule causes it?

wershlak: that's ugly. Added `BreakStringLiterals: false`

wershlak:
@cizra Any opinion on what style we use? Also do you think it's a good idea to format the entire repo once we decide on a style?

cizra:
> @cizra Any opinion on what style we use? Also do you think it's a good idea to format the entire repo once we decide on a style?

I'm 100% sure we should format the whole repo in one go. See https://www.moxio.com/blog/43/ignoring-bulk-change-commits-with-git-blame how to make this commit not end up in `git blame`.

As for which style to use - I haven't read them all. Based on cursory glance, it seems Google / LLVM / Chromium are all similar enough.
In particular, I like:
- tabs
- Egyptian braces
- curly elses
- ... and generally to preserve vertical space

cizra (changes requested):

drewhershey:
Going to close this out, as I have a .clang-format file I've been tweaking for a while now that I'm going to use for this purpose on the entire codebase at once, which will be done in a new PR.

PR #361, opened 2021-07-29
Fix runtime errors

drewhershey:
This PR fixes various, usually small, pieces of code that were causing Undefined Behavior Sanitizer + clang (in Ubuntu) to generate runtime errors and/or seg faults while the game was running.

Tested that it compiles and works with g++ using the default SConstruct file as well. G++ doesn't catch as many runtime errors as clang, I think because clang recognizes some optional UBSan flags that g++ doesn't. 

See comments on individual commits for more info on them. I did a lot of digging/researching/testing for these fixes so I feel pretty good about the quality of them, but if any of them look weird or something I don't have any problem with just scrapping them all, since unless the Docker image starts compiling with clang these were all issues that didn't seem to actually be causing crashes.


code/code/disc/disc_adv_adventuring.h
```diff
@@ -19,8 +19,7 @@ class CDAdvAdventuring : public CDiscipline
     CSkill skForage;
     CSkill skSeekWater;
     CSkill skSkin;
-    CSkill skDivination;
```
cizra, line 22: Can you delete this whole mess, instead, and rely on default constructors?

code/code/disc/disc_murder.h
```diff
@@ -36,23 +36,19 @@
...
 {
 public:
-    CSkill skGarrotte;
     CSkill skThroatSlit;
 
     CDMurder()
-      : CDiscipline(),
-      skGarrotte(),
+      : CDiscipline(),  
       skThroatSlit() {
```
cizra, line 43: Try deleting these constructors and assignment operator?

code/code/disc/disc_thief_fight.h
```diff
@@ -28,22 +28,26 @@ class CDThiefFight : public CDiscipline
...
 public:
     CSkill skDodgeThief;
+    CSkill skGarrotte;
     CSkill skDualWieldThief;
 
     CDThiefFight()
       : CDiscipline(),
       skDodgeThief(),
+      skGarrotte(),
       skDualWieldThief() {
```
cizra, line 38: Try deleting

code/code/misc/monster.h
```diff
@@ -115,7 +115,8 @@ class TMonster : public TBeing {
     // VIRTUAL FUNCTIONS
     virtual double baseDamage() const;
     virtual int grenadeHit(TTrap *);
-    virtual int checkSpec(TBeing *, cmdTypeT, const char *, TThing *);
+    using TThing::checkSpec;
```
cizra, line 118: Which problem is this line fixing?

code/code/misc/periodic.cc
```diff
@@ -2080,8 +2080,10 @@ int TObj::updateBurning(void)
     // suppose we have a wooden staff, 4000 volume, 20 structure and 500
     // flammability, (20/4000)=.005, .005*500 = 2.5, 20/2.5 = 8
     // so it will burn for approx 8 rounds
-    int burnamount = (int)((double)((double)getMaxStructPoints()/getVolume())*
-			       material_nums[getMaterial()].flammability);
+
+    // (getVolume() | 1) avoids division by zero edge case
```
cizra, line 2084: Huh, what a "creative" approach.
is it OK that a volume of 100 now magically starts to be 101?
Why the double cast to `double`? Why the explicit cast to int?

code/code/misc/player_data.cc
```diff
@@ -1084,7 +1084,7 @@ void do_the_player_stuff(const char *name)
     return;
 
   // skip faction data
-  if (strlen(name) > 6 && !strcmp(&name[strlen(name) - 8], ".faction"))
+  if (strlen(name) > 8 && !strcmp(&name[strlen(name) - 8], ".faction"))
```
cizra, line 1087: why not sstring(name).endsWith(".faction")? if it doesn't exist, let's write it, instead of perpetuating this eyebleed.

code/code/misc/weather.cc
```diff
@@ -751,7 +746,7 @@ void Weather::weatherChange()
   // keep in range 1040 - 960
 
   int diff = 0;
-  changeWeatherT change;
+  Weather::changeWeatherT change = Weather::CHANGE_NONE;
```
cizra, line 749: Could you explain why `Weather::` is necessary in `Weather::changeWeatherT`?

code/code/sys/db.cc
```diff
@@ -3437,15 +3437,16 @@ void runResetCmdZ(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload,
 
 void runResetCmdY(zoneData &zone, resetCom &rs, resetFlag flags, bool &mobload, TMonster *&mob, bool &objload, TObj *&obj, bool &last_cmd)
 {
+  if (!mob || !mobload) {    
+    return;
```
cizra, line 3441: should log an error, complete with zone name and any other helpful data.

cizra (changes requested):

code/code/misc/monster.h
drewhershey: Without that line, or if I had two separate overloads there with simply a different parameter type as the final parameter (one being the original TThing* and one being TObj*) I got tons of compiler warnings about hiding an overload whenever thing.h was used in the same file as monster.h (IIRC). From researching it, adding that using statement was one thing that was suggested on places like StackOverflow as a way to avoid those types of warnings.

code/code/misc/periodic.cc
drewhershey: > Huh, what a "creative" approach.

Guessing that's sarcasm, haha. 

> is it OK that a volume of 100 now magically starts to be 101?

Not entirely sure what you mean here, which makes me think that I misunderstood how short circuiting works in C++. From looking it up I was under the impression that doing the `(getVolume() | 1)` there would behave similarly to how I use them in JavaScript, where if `getVolume()` resolves to a truthy value it will ignore the `| 1`, whereas if `getVolume()` is falsy it will replace it with the value on the other side of the `|` operator. But I take it that's not correct?

> Why the double cast to double? Why the explicit cast to int?

I don't honestly know - for a lot of this stuff I'm just trying to get the the code to work within the context of the original intent/methodology. Here I just saw that they were getting runtime errors for division by 0, so I tried to fix that specific part of the problem. Maybe I should've looked into it more deeply, but I'm already a little beyond my depth as you can probably tell.

code/code/misc/weather.cc
drewhershey: Nope haha, I can't explain it as I don't know the answer (though I'm guessing the answer is "it's not necessary"?). That was just how they declared it when they initialized the same variable elsewhere so I wrote it the same way here. Would they have written it that way elsewhere because they weren't inside the Weather namespace in that code?

code/code/misc/player_data.cc
drewhershey: > why not sstring(name).endsWith(".faction")?

Entirely because I haven't learned how sstrings work yet and didn't really understand what this code is doing. 

I just saw what appeared to be an easy fix due to a typo that was causing some runtime errors and was attempting to stabilize more than optimize.

code/code/sys/db.cc
drewhershey: That function gets called with a null pointer passed to mob hundreds of times during boot, so it would be adding a lot of bloat to the logs, but I definitely can if you want. I'm guessing you saw it, but just in case - I tried to explain my thought process on this one in the comments on this commit itself.

drewhershey:
My takeaway from this is that it I'm out of my league when it comes to making these sorts of changes. It's probably better if I just stick to working on adding/updating/fixing game features and leave digging into the guts of the codebase itself to people more experienced than I with the technical aspects of C++. Should I just close this PR out? Like I said previously it's not like any of these things were actually causing issues in the live build of the game, so probably not a big deal to leave them as is.

code/code/misc/monster.h
cizra: Hmmmm, how's the spec runner going to know which overload to call, anyway?

code/code/misc/periodic.cc
cizra: Nah, that'd be `||`. By doing `a | 1` you're setting the last bit to 1, no matter what the value of a is. So if a is 0 or 1, you end up with 1. For any even number, you get number+1.

cizra: Regarding the casts - the usual way is to omit them, unless you actually need an explicit cast. Here, you might need just one cast - so that the division would be fractional, not whole-integer division. The rest look redundant, to me.

Regarding "within the context" - well, the code might have been good, once upon a time, but it's gotten a bit dusty and rusty. Best if we clean up anything we touch.

code/code/misc/player_data.cc
cizra: I'm worried here. We're making a change without understanding the intent of the original code - and we assume it works purely because the compiler doesn't warn anymore.

`strlen()` is rarely a good thing to call - code should be high-level, read like a book. `foo.endsWith(".faction") reads easily. Even if your language (plain old C, suppose) doesn't support strings / methods, you should still have a function  `string_endsWith(foo, ".faction")` or similar.

cizra: I just had a look at the code - turns out that this `do_the_player_stuff` (genius name, huh?) is called for every file in player files. Therefore `name` is a filename, and the only sensible thing you can do with a filename and `.faction` is to check whether the filename ends with `.faction`. Therefore, even without hunting down the bytes and offsets and pointers in that (frankly foul-smelling) `strlen()`, we can assume that it's checking for the ending.

We have a string library in Sneezy, in `code/sys/sstring.h` and `code/sys/sstring.cc`, so what I'd do in this situation is slap on another method into .h, implement it in the cc, and write a unittest in `code/tests/SString.h`, then replace that strlen-monstrosity with a nice and readable boolean check.

code/code/misc/weather.cc
cizra: Class, not namespace - otherwise yes. You don't need that prefix inside the class. Though you added it, so I was assuming that I'm missing something, and thought perhaps there's some reason.

code/code/sys/db.cc
cizra: oof, no, then not.
Tho there's a rule of thumb that says - anything asked in code review deserves explaining in the code itself. Perhaps add a comment about that?

cizra: Commit log is not what one usually reads, when trying to figure out what a given bit of code does :)

cizra (changes requested):

cizra:
> My takeaway from this is that it I'm out of my league when it comes to making these sorts of changes. It's probably better if I just stick to working on adding/updating/fixing game features and leave digging into the guts of the codebase itself to people more experienced than I with the technical aspects of C++. Should I just close this PR out? Like I said previously it's not like any of these things were actually causing issues in the live build of the game, so probably not a big deal to leave them as is.

That depends on if you'd like to learn something new :) My vote is - don't close it, let's figure out how to fix it, together.

drewhershey:
So after coming back to this PR many months later, I'm going to close it out with the intention to do a new runtime error PR in the future. With my C++ skills and knowledge of the codebase greatly improved since I started this one I'm not satisfied with the way I approached some of the fixes so I'd rather just start over fresh, using the most recent versions of g++ and clang and the sanitizers.

PR #368, opened 2021-08-06
Thief quest weapon proc

drewhershey:
This PR repurposes the spec proc currently only used on a single weapon - silver jambiya - for use on a new thief quest weapon, as requested by Vasco. It maintains the same overall functionality as the original proc while making it much more likely to occur, do more damage, and also work with slit throat.

To be able to pass which limb was stabbed from the stab skill when it calls this proc I had to create some new methods to convert the limb type enums to strings and back.

I also had to move the code that calls checkSpec from the backstab and slit functions so that the proc was executing visually right after the attack, and also not executing on missed attacks. The place it was before allowed it to proc on a missed attack, or sometimes it would proc after the mob had already fled the room in reaction to a slit/backstab.


wershlak:
Can't you convert the int to a char to pass to checkSpec()? I'm wondering if all the code is necessary or if that just seemed like the best way to do it?

code/code/disc/disc_murder.cc
```diff
@@ -137,34 +137,33 @@ int TBeing::backstabHit(TBeing *victim, TThing *obj)
...
+    act("$n tried to backstab you, but you avoid $m.", FALSE, this, obj, victim, TO_VICT);
+    act("$n tried to backstab $N, but nearly cut $s own finger.", FALSE, this, obj, victim,
+        TO_NOTVICT);
     d = 0;
   }
 
   if (reconcileDamage(victim, d, SKILL_BACKSTAB) == -1)
     return DELETE_VICT;
 
-  victim->addHated(this);
```
wershlak, line 165: Did this get moved somewhere else or did you find in not needed? What's the purpose in removing it?

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -0,0 +1,56 @@
...
+
+  auto weapon = dynamic_cast<TBaseWeapon *>(object);
+  auto thief = weapon ? dynamic_cast<TBeing *>(weapon->equippedBy) : nullptr;
+  if (!weapon || !thief) return false;
+
+  auto damage = 0;
+  auto damageType = SKILL_STABBING;
+
+  if (command == CMD_STAB) {
+    damage = thief->GetMaxLevel() * 1.5;
```
wershlak, line 33: So an extra 150 damage on stab? up to 300 damage on slit

You have this going off 50% of the time so that could be a lot of damage from this proc. particularly for stab across a fight.

How much have you tested the impact of the damage here? My intuition is that it's a lot of added damage.

Would you consider toning the damage down? Maybe proc 33% and reduce the damage to 75 for stab and maybe up to 200 for slit.

Is it more powerful if you reduce damage and add some life drain or healing component when it procs?

My experience is people will be mad if we decide it's too good later and we want to nerf it :)

wershlak (changes requested):

drewhershey:
> Can't you convert the int to a char to pass to checkSpec()? I'm wondering if all the code is necessary or if that just seemed like the best way to do it?

If that works I'm gonna feel real dumb for not thinking of it, lol, and I'm guessing it will. I'll try it out.

code/code/disc/disc_murder.cc
drewhershey: Yeah, it actually gets called almost immediately after this function returns, at line 354, so this call was redundant.

code/code/spec/spec_objs_thief_quest.cc
drewhershey: No, you're right, it looks super strong. Vasco intentionally wanted it pretty strong, since he's planning to use it on a lvl 45-50 thief quest weapon, but he's also going to do playtesting of it himself and we'll tweak the power of it from there. Currently it's only used on one weapon in the game that no one even uses, and Vasco could just remove the proc from it temporarily if he needed to.

Also, unless I'm doing my math really wrong here, I intended the damage to be 75 on stab, 125 on backstab, and slit to be 150. Where are you getting the 150 for stab/300 for slit?

I'm up for any changes you guys think make sense - the original proc was super low chance and fairly low damage, so we just ramped it up quite a bit.

I tested it out quite a bit myself and it's definitely strong, but there's not really a good way to let Vasco test it otherwise, at this point. I really need to set up a dev server I can let others connect to.

wershlak: My brain isn't working. I think I was thinking skill level or something.

wershlak:
> > Can't you convert the int to a char to pass to checkSpec()? I'm wondering if all the code is necessary or if that just seemed like the best way to do it?
> 
> If that works I'm gonna feel real dumb for not thinking of it, lol, and I'm guessing it will. I'll try it out.

Want to let me know when you try that and I'll merge this?

Eventually we'll get that test server up and can test from a branch which will be better.

--

People have quest items or things Aion built for them via coins that could end up with a proc on it even if it's meant to be unique or special so we should ensure things are somewhat balanced because you never know where it will end up.

But I really like the flavor and the fact that this adds damage to stab backstab and slit rather than just proccing off auto attack.

drewhershey:
> Can't you convert the int to a char to pass to checkSpec()? I'm wondering if all the code is necessary or if that just seemed like the best way to do it?

Fixed in https://github.com/sneezymud/sneezymud/pull/368/commits/ba0818e5e1194a9b5e6795547673c7f56b7bb3fd btw

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -0,0 +1,62 @@
...
+#include "being.h"
+#include "comm.h"
+#include "extern.h"
+#include "obj_base_weapon.h"
+
+int thiefQuestWeapon(TBeing *victim, cmdTypeT command, const char *arg, TObj *object, TObj *) {
+  wearSlotT limb = WEAR_NOWHERE;
+
+  if (command != CMD_STAB && command != CMD_BACKSTAB && command != CMD_SLIT) return false;
+  if (command == CMD_STAB) limb = static_cast<wearSlotT>(reinterpret_cast<uintptr_t>(arg));
```
wershlak, line 12: So when we get to this conversion we know that this is coming in with the CMD_STAB and you've coded CMD_STAB to pass in the limb. There should be no way that something else could get here that's not what you expect. 

So it looks kinda scary but I think it should be safe.

Let's see if @cizra has an opinion on this bit.

wershlak (approved):
Approved pending feedback from @cizra

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Right - not sure if adding the extra code to convert cleanly between enum/string or doing these ugly casts is more preferable here. The code seems to work fine either way though, based on my fairly extensive in-game testing/debugging of it.

Until we tackle the job of rewriting the checkSpec function to accept typed unions or variants or whatever as parameters, and therefore also update the function signatures of every single spec proc that gets called by checkSpec, we'll have to find less favorable workarounds like this, I think. :\

cizra: I don't know of a better way of doing it, except for https://en.cppreference.com/w/cpp/utility/variant - and if we're not ready to switch to varianst, well, we gotta reinterpret_cast them :(

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -0,0 +1,62 @@
...
+  if (command != CMD_STAB && strcmp(arg, "-special-")) return false;
+  if (!victim || !object) return false;
+
+  limb = command == CMD_BACKSTAB ? WEAR_BACK
```
cizra, line 16: I find it looks better when the ? is split into the next line.
```cpp
limb = command == CMD_BACKSTAB
    ? WEAR_BACK
    : command == CMD_SLIT
    ? WEAR_NECK
    : limb;
```

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -0,0 +1,62 @@
...
+  if (command == CMD_STAB) {
+    if (::number(0, 9)) return false;
+  } else if (::number(0, 1))
+    return false;
+
+  auto weapon = dynamic_cast<TBaseWeapon *>(object);
+  auto thief = weapon ? dynamic_cast<TBeing *>(weapon->equippedBy) : nullptr;
+  if (!weapon || !thief) return false;
+
+  auto damage = 0;
```
cizra, line 35: int is 1 less character to type :stuck_out_tongue:

cizra (changes requested):

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Yeah, this was before I learned that stuff about auto from you the other day. I'm gonna change them back to explicit types and just use auto from now on where it actually does save some typing/adds clarity.

code/code/spec/spec_objs_thief_quest.cc
drewhershey: I agree - I formatted this using the clang-format rules @wershlak set up in the repo so that might be one we need to change. I'll just reformat them manually for now.

drewhershey: Personally I also favor having the statement in a single-statement `if` on its own line too, even when it's pretty short. Feels like it adds readability. Maybe that's something we could change in clang-format too.

code/code/disc/disc_murder.h
```diff
@@ -32,6 +32,7 @@
 
 #include "discipline.h"
 #include "skills.h"
+#include "obj.h"
```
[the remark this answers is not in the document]
drewhershey: I just removed the obj.h include and compiled and this time it compiled fine. I might've been using clang still at the point that I got this error - I'm back to using g++ now just to be consistent with everyone else. Either way, resolved in https://github.com/sneezymud/sneezymud/pull/368/commits/1f6c058d899d53f65470619c7ef8118014f73944.

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Resolved in https://github.com/sneezymud/sneezymud/pull/368/commits/7cd35cc3ce8966bdb70498453540f0af242e6778

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Resolved in https://github.com/sneezymud/sneezymud/pull/368/commits/7cd35cc3ce8966bdb70498453540f0af242e6778

PR #369, opened 2021-08-06
Pierce crit rework

parkscs:
- Updated the pierce crit table, added damage bonuses for special crit attacks and removed weapon dislodging.
- Adjusted NPC crits so that high level NPCs do not always land the same crit types in the crit table, buffed high level NPC crit rate to make up for this reduction in crit severity.
- Fixed a bug in retreat that makes success rate scale inversely with skill level and overall buffed retreat.

The new pierce crit table is as follows:

====Pierce crit table====
0-33: double damage
34-66: triple damage
67,68,69,70,71: pierced larynx (4-5x damage)
72,73,74: gouged out eye (5x damage)
75,76,77,78: sever tendon (4x damage)
79,80,81,82: stab back (impale) (4-5x damage)
83,84: pierce cranium (4-5x damage)
85,86,87: shatter elbow (4x damage)
88,89,90: sever hand (4x damage)
91,92,93: punctured lung (impale) (4-5x damage)
94,95,96: punctured kidney infect (impale) (4-5x damage)
97,98,99: punctured stomach (4-5x damage)
100: crit kill (death)

total:
  double damage: 33%
  triple damage: 33%
  limb mutilation (leg/elbow/hand): 10%
  body mutilation (lung/kidney/stomach): 9%
  debilitate (larynx/eye): 8%
  stun: 4%
  death: 3% (1% if helmet)

### DIFF
--- a/code/code/misc/offense.cc
+++ b/code/code/misc/offense.cc
@@ -952,8 +952,8 @@ int TBeing::doFlee(const char *arg)
   for (i = 0; i < 20; i++) {
     dirTypeT attempt = dirTypeT(::number(MIN_DIR, MAX_DIR-1));        // Select a random direction 
     
-    // fighting, so give slight chance of letting PC direct the direction
-    if (chosenDir != DIR_NONE && !panic && !(::number(0,99) < (30+getSkillValue(skill)/2)))
+    // fighting, so give PC a skill-based chance to specify direction with a small failure chance at max skill
+    if (chosenDir != DIR_NONE && !panic && ::number(0,100) < 45+(getSkillValue(skill)/2))
       attempt = chosenDir;
     
     if (canFleeThisWay(this, attempt)) {
--- a/code/code/misc/crit_combat.cc
+++ b/code/code/misc/crit_combat.cc
@@ -2109,6 +2109,223 @@ buf=format("$n's %s rips into $N, tearing the tendon in $S lower leg.") %
 	return ONEHIT_MESS_CRIT_S;
       case 79:
       case 80:
+      case 81:
+	// Strike shatters elbow in weapon arm. Arm broken 
+	new_slot = v->getSecondaryArm();
+	if (!v->hasPart(new_slot))
+	  return 0;
+	if (!v->isHumanoid())
+	  return FALSE;
+buf=format("$N blocks your %s with $S arm.  However the force shatters $S elbow!") %
+		limbStr;
+	act(buf, FALSE, this, 0, v, TO_CHAR, ANSI_ORANGE);
+buf=format("$n's %s is blocked by your arm.  Unfortunately your elbow is shattered!") %
+		limbStr;
+	act(buf, FALSE, this, 0, v, TO_VICT, ANSI_RED);
+buf=format("$n's %s shatters $N's elbow!") %
+		limbStr;
+	act(buf, FALSE, this, 0, v, TO_NOTVICT, ANSI_BLUE);
+	v->damageArm(FALSE,PART_BROKEN);
+	if (desc)
+	  desc->career.crit_broken_bones++;
+	if (v->desc)
+	  v->desc->career.crit_broken_bones_suff++;
+	*part_hit = new_slot;
+	// triple damage
+        *dam *= 3;
+
+	rc = damageLimb(v,new_slot,weapon,dam);
+	if (IS_SET_DELETE(rc, DELETE_VICT))
+	  return DELETE_VICT;
+	return ONEHIT_MESS_CRIT_S;
+      case 82:
+	new_slot = v->getPrimaryHand();
+	if (!v->hasPart(new_slot))
+	  return 0;
+	if (!v->isHumanoid())
+	  return 0;
+
+	// triple damage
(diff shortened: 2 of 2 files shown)


wershlak:
Looks like stuff is pasted in twice in crit_combat.cc and there's a change to flee in this as well?

code/code/misc/offense.cc
```diff
@@ -952,8 +952,8 @@ int TBeing::doFlee(const char *arg)
   for (i = 0; i < 20; i++) {
     dirTypeT attempt = dirTypeT(::number(MIN_DIR, MAX_DIR-1));        // Select a random direction 
     
-    // fighting, so give slight chance of letting PC direct the direction
-    if (chosenDir != DIR_NONE && !panic && !(::number(0,99) < (30+getSkillValue(skill)/2)))
+    // fighting, so give PC a skill-based chance to specify direction with a small failure chance at max skill
+    if (chosenDir != DIR_NONE && !panic && ::number(0,100) < 45+(getSkillValue(skill)/2))
```
wershlak, line 956: 45+skillValue/2 seems arbitrary and doesn't use the methods in place for skill checks

Why not just call bsuccess() here? 

bool TBeing::bSuccess(spellNumT spell)

The rate of success is determined generally by the difficulty of the skill. Any easy task has a success chance of 90%. By using this method you can adjust the success chance by changing the difficulty in the spell info table and get things like hones for the skill without having to implement that yourself.

parkscs:
The idea with 45+skillValue/2 is simply that we were discussing having a
chance of success even for characters that don't have retreat.
Namely, a 45% chance of success with a max of 95%.
I suspect if we rely on bSuccess instead, we'd have to separately add a way
for people without retreat to have a chance at success, as I would expect
bSuccess to always fail if they don't have the skill.

code/code/misc/offense.cc
drewhershey: This is just a guess, but is it possible they didn't use bSuccess there because it's inside a `for` loop that runs up to 20 times looking for possible flee directions, and every time bSuccess executes it potentially gives a skill hone?

I just remember seeing comments elsewhere in the code about not wanting to use bSuccess in a part of the code that gets called really frequently/repetitively and wondered if it was relevant here.

wershlak:
```
for (i = 0; i < 20; i++) {
    dirTypeT attempt = dirTypeT(::number(MIN_DIR, MAX_DIR-1));        // Select a random direction 
    
    // fighting, so give slight chance of letting PC direct the direction
    if (chosenDir != DIR_NONE && !panic && !(::number(0,99) < (30+getSkillValue(skill)/2)))
      attempt = chosenDir;
    
    if (canFleeThisWay(this, attempt)) {
      if (panic || !doesKnowSkill(skill) || 
          !bSuccess(skill)) {
        act("$n panics, and attempts to flee.", TRUE, this, 0, 0, TO_ROOM);
        panic = TRUE;
      } else {
        act("$n skillfully retreats from battle", TRUE, this, 0, 0, TO_ROOM);
        panic = FALSE;
      }
      ```
      You're right Drew I didn't look at the code closely. But this code is pretty terrible. There's a for loop we iterate through 20 times with a variable i that isn't used for anything.
We determine what direction to try and then there's a skill chance later. What a mess

wershlak: So here's the dilemma we have with a lot of code. There's an undesired behavior or, as in this case, a bug. Fixing the bug improves gameplay experience but we're touching code that is terrible and should be refactored. Do we prioritize the gameplay experience and just merge the code or decide that someone that is touching it has the duty to improve the code?

I've tried to improve some code and I've left some shitty code that @cizra probably hates me for.

I feel bad asking you to refactor this and making your simple bug fix into a larger project @parkscs. I'll let you decide if you want to:
- merge this flee fix and let someone else worry about the code.
- pull it out of this PR and spend time to refactor.

Let me know.

parkscs:
Given that this is a bug fix and a single line of code, I'd be inclined to
propose we consider adding it in for now and also making a plan to
completely refactor this function going forward to improve the code.
Realistically I don't know when I'll get around to redo'ing the flee
function and I think we all have things we're wanting to work on in the
near future, but in the meantime we have a pretty wonky bug that's
affecting players.  I would agree we should look to improve the code we
touch, but just my personal opinion, we should balance that with making
timely improvements to the gameplay and here we could easily fix a bug
that's annoying players today and I don't think this change makes the code
worse (it just doesn't improve it beyond a bug fix, haha).

code/code/misc/offense.cc
drewhershey: I actually already started completely refactoring that doFlee function on my own a couple days ago because retreat going the wrong way was annoying me so badly when I was testing thief stuff. So if you want, we can merge this PR just to get the retreat fix in and I'll just finish my refactor of flee that I've already started, @parkscs? That way you can keep going on your warrior stuff.

parkscs:
I'd definitely be in favor of that - and let me know if I can help in any
way.

parkscs:
Updated all values to at max deal 3x damage.

PR #370, opened 2021-08-06
Misc. QoL improvements

parkscs:
- Moving cloth-like materials (silk, wool, etc.) under monk repair, as they're under an non-implemented skill atm (mend)
- Quality of life tweaks for riders, particularly deikhans
- Allowing sacrifice while the shaman is flying


code/code/misc/riding.cc
```diff
@@ -1,4 +1,4 @@
-#include "handler.h"
+nclude "handler.h"
```
drewhershey, line 1: Looks like a couple characters got deleted here, probably going to cause an error

code/code/disc/disc_shaman.cc
```diff
@@ -724,7 +724,9 @@ TTool *TBeing::getHeldTotem() {
 void TBeing::doSacrifice(const char *arg) {
   for (; isspace(*arg); arg++);
 
-  if (getPosition() != POSITION_STANDING) {
+  positionTypeT position = getPosition();
+
+  if (position != POSITION_STANDING || position != POSITION_FLYING) {
```
cizra, line 729: Did you test it? How?
We really, really should get into a habit of writing unit tests.

cizra (changes requested):

code/code/misc/riding.cc
```diff
@@ -799,7 +805,22 @@ int MountEgoCheck(TBeing *ch, TBeing *horse)
     check =  horse->GetMaxLevel();
     check += dynamic_cast<TMonster *>(horse)->anger() / 10;
     check -= ch->GetMaxLevel();
+
+    if (ch->doesKnowSkill(SKILL_RIDE))
     check -= ch->getSkillValue(SKILL_RIDE)/10;
+    // Boz Edit 6-23-2021 - At least making this significantly less of a hurdle for Deikhans
```
cizra, line 811: Redundant with Git commit history. That's OK, though.

cizra (changes requested):

parkscs:
I tested it on my local instance previously, but I refactored this slightly
when repackaging things in github and I'm pretty sure there's an error in
how I did it.  Testing again and resubmitting momentarily.

parkscs:
That's a habit of mine from when I first started making these changes,
haha, and I've mostly stopped it at this point.  I thought I cleaned out
most of those notes in the code but I guess not all of them. :)

code/code/disc/disc_shaman.cc
wershlak: You can do greater than and less than comparisons with these enums too. https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/enum.h#L157-L171

`if (getPosition() < POSITION_STANDING)` for instance would allow you to address everything besides standing, flying and mounted. Just FYI in case it's useful.

parkscs:
Good to know.  I feel like preventing mounted sacrifice is acceptable... I
just used OR when I needed to use AND, lol.  Resubmitted.

cizra (changes requested):

code/code/misc/riding.cc
```diff
@@ -823,8 +841,24 @@ int MountEgoCheck(TBeing *ch, TBeing *horse)
...
+
+    // Bonus for proficiency in type-specific mount ability
+    if (ch->doesKnowSkill(SKILL_RIDE_WINGED) && horse->mountSkillType() == SKILL_RIDE_WINGED)
+      check -= ch->getSkillValue(SKILL_RIDE_WINGED)/6;
+    else if (ch->doesKnowSkill(SKILL_RIDE_DOMESTIC) && horse->mountSkillType() == SKILL_RIDE_DOMESTIC)
+      check -= ch->getSkillValue(SKILL_RIDE_DOMESTIC)/6;
+    else if (ch->doesKnowSkill(SKILL_RIDE_NONDOMESTIC) && horse->mountSkillType() == SKILL_RIDE_NONDOMESTIC)
+      check -= ch->getSkillValue(SKILL_RIDE_NONDOMESTIC)/6;
+    else if (ch->doesKnowSkill(SKILL_RIDE_EXOTIC) && horse->mountSkillType() == SKILL_RIDE_EXOTIC)
+      check -= ch->getSkillValue(SKILL_RIDE_EXOTIC)/6;
```
cizra, line 860: This code is fine, but it's not how I'd write it. I'd write something like this:

```cpp
for (const auto skill : {SKILL_RIDE_WINGED, SKILL_RIDE_DOMESTIC, SKILL_RIDE_NONDOMESTIC, SKILL_RIDE_EXOTIC})
  check -= ch->doesKnowSkill(SKILL_RIDE_WINGED) ? ch->getSkillValue(SKILL_RIDE_WINGED)/6 : 0;
```

Or else I'd see if there's a getSkillValue variant that returns 0 if you don't know the skill, and get rid of the ternary.

code/code/misc/riding.cc
```diff
@@ -344,10 +344,6 @@ int TBeing::doMount(const char *arg, cmdTypeT cmd, TBeing *h, silentTypeT silent
       sendTo("You can't ride that!\n\r");
       return FALSE;
     }
-    if (horse->fight()) {
-      sendTo("Trying to ride something involved in a life-or-death fight is not advisable.\n\r");
-      return FALSE;
-    }
```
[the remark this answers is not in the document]
parkscs: I believe this is already covered above in the following block of code:

```
if (!isImmortal() && (horse->isTanking() || (horse->fight() && !hasClass(CLASS_DEIKHAN)))) {
      sendTo("You do not have the skill to mount something that is fighting!\n\r");
      return FALSE;
    }
```

code/code/misc/riding.cc
parkscs: Pretty nifty.  I haven't used this before (and it'll probably be a while before I think to use it again just based on comfort level), but I reworked the code just to get more familiar with it.  Tested and it seems to function the same way as before (properly).  

I did re-order some of checks just because they make a bit more sense to me this way in the game, namely, we perform the check on if you can ride a flying mount first (because this can be a hard fail) and then we later attempt the ride check for being in combat/tanking.  In the original order, you would get a message about failing to ride because it's hard while tanking... and then if you passed the check, it would just say you can't ride a flying mount.  In short, I moved the hard check before the soft check.

code/code/misc/riding.cc
```diff
@@ -432,12 +428,6 @@ int TBeing::doMount(const char *arg, cmdTypeT cmd, TBeing *h, silentTypeT silent
       }
     }
     if (roomp && !roomp->isFlyingSector()) {
-      if (isFlying()) {
-        if (!horse->isFlying()) {
-          sendTo("Riding a grounded mount while flying is impossible.\n\r");
```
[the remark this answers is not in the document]
parkscs: I reworked the if statements here and added some comments to hopefully make a bit more sense about the scenarios we're checking for (including non-deikhans).

code/code/misc/riding.cc
```diff
@@ -823,8 +847,18 @@ int MountEgoCheck(TBeing *ch, TBeing *horse)
...
     if (ch->doesKnowSkill(SKILL_RIDE))
       check -= ch->getSkillValue(SKILL_RIDE)/10;
+
+    // Bonus for proficiency in advanced riding disc.
+    if (ch->doesKnowSkill(SKILL_ADVANCED_RIDING))
+      check -= ch->getSkillValue(SKILL_ADVANCED_RIDING)/8;
+
+    // Bonus for proficiency in type-specific mount ability
+    for (const auto skill : {SKILL_RIDE_WINGED, SKILL_RIDE_DOMESTIC, SKILL_RIDE_NONDOMESTIC, SKILL_RIDE_EXOTIC })
+      check -= horse->mountSkillType() == skill && ch->doesKnowSkill(skill) ? ch->getSkillValue(skill)/6 : 0;
```
cizra, line 860: Hey, truly sorry, but the previous version only checked as many items as necessary to find a match. Now we loop over them all - even if the mount is WINGED, we still try them all until EXOTIC.

We could add `{ check -= ... ; break}`.

parkscs: Reworked this (and learned a bit more about ternary operators).  I moved away from the ternary operator since I don't think I can include the break within one as it's not an expression.  Let me know if this new format looks acceptable.  Thanks!

drewhershey: I think your ternary statement would've worked fine if you just put the break on the next line after it, instead. But the way you have it now actually reads nicer anyway, I think. Ternarys can be super useful but can also impact legibility a lot of times.

cizra (approved):

cizra (changes requested):
I find these commit messages could use some improvement.
"Update riding.cc" isn't really something I want to see when browsing history, or trying to figure out why/how a certain bug came into being.

Would you like a crash course on Git commit editing?

parkscs:
I wouldn’t mind it but I can also just work on making these more
informative all the time.  On larger changes I tend to try and say more
(although perhaps still not enough?), but I suspect the one you’re
mentioning was a very minor change and I just got lazy with it and used the
default message.

cizra (approved):

PR #371, opened 2021-08-07
Garble tweaks

parkscs:
Two major tweaks to garble:

- Added a new skill to advanced adventuring - Common.  Allows players to spend practices and hone a skill that lets them improve their accent to a substantial degree.  This won't erase their accent at max skill but will heavily mitigate it, allowing players to use their resources to better communicate with other players.  
- Remove garble from psionic messages - this is pretty straightforward, but pick your favorite D&D series with a mindflayer and their psychic message were not "gurble gurble garbleh wrist rglgle" but instead were clear thoughts.  This gives a niche to psionic communications and is intended as a nice quality of life improvement.


code/code/disc/disc_adv_adventuring.h
```diff
@@ -39,7 +40,8 @@ class CDAdvAdventuring : public CDiscipline
       skTrollish(),
       skBullywug(),
       skAvian(),
-      skKalysian()
+      skKalysian(),
+      skCommon()
```
cizra, line 44: I suspect that these constructors don't do anything smart. Try removing, see if the autogenerated version works fine?

code/code/disc/disc_psionics.cc
```diff
@@ -166,7 +165,7 @@ int TBeing::doPTell(const char *arg, bool visible){
 
   mud_str_copy(capbuf, vict->pers(this), 256);  // Use Someone for tells (invis gods, etc)
 
-  char garbedBuf[256];
+  char messageBuf[256];
```
cizra, line 168: Could we drop that `messageBuf` and instead do the string concatenation using `+` or perhaps `.format()`?

code/code/disc/disc_psionics.cc
```diff
@@ -148,9 +147,9 @@ int TBeing::doPTell(const char *arg, bool visible){
     return FALSE;
   }
 
-  sstring garbed = garble(vict, message, Garble::SPEECH_TELL);
```
cizra, line 151: Did you just remove _all_ garbles from _everywhere_?

code/code/disc/disc_psionics.cc
```diff
@@ -366,19 +365,17 @@ void TBeing::doPShout(const char *msg){
   } else {
     learnFromDoing(SKILL_PSITELEPATHY, SILENT_NO, 0);
 
-    mud_str_copy(garbed, garble(NULL, msg, Garble::SPEECH_SHOUT, Garble::SCOPE_EVERYONE), 256);
+    mud_str_copy(message, msg, 256);
```
cizra, line 368: What's the point of this?

code/code/disc/disc_psionics.cc
```diff
@@ -251,9 +250,9 @@ int TBeing::doPSay(const char *arg){
 
     learnFromDoing(SKILL_PSITELEPATHY, SILENT_NO, 0);
 
-    mud_str_copy(garbed, garble(NULL, arg, Garble::SPEECH_SAY), 256);
+    mud_str_copy(message, arg, 256);
```
cizra, line 253: What good does this line achieve?

code/code/misc/garble.cc
```diff
@@ -224,6 +224,8 @@ sstring TBeing::garble(TBeing *to, const sstring &arg, Garble::SPEECHTYPE speech
 // The speakers ability to "talk good". Since this is an accent and not a language per se this makes sense.
 int getLanguageChance(const TBeing *from, TBeing *to, spellNumT language)
 {
+  TBeing * tFrom = const_cast<TBeing *>(from); // copying this from another function within garble.cc - need to check skill
```
cizra, line 227: Why const_cast? Sounds really, really dangerous...

cizra (changes requested):

code/code/disc/disc_adv_adventuring.h
parkscs: I'll give it a go this weekend and provide feedback.

code/code/disc/disc_psionics.cc
parkscs: Same as above - I used these because they were already used in the code, but I'll try moving away from them and update this weekend.

code/code/disc/disc_psionics.cc
parkscs: Negative.  This is only the garble for **psionic** messages.  We have plenty of examples in fantasy of telepathic communications bypassing language barriers - from my perspective, it follows that they should also bypass accents as well.

code/code/disc/disc_psionics.cc
parkscs: I was trying to stick with the existing code as much as possible.  If we move from character arrays as suggested above, I suspect this will also change/go away.

code/code/misc/garble.cc
parkscs: The short of it is these functions pass in the TBeing the message is from as a const, but since we're going to call functions like bSuccess (which have a chance to hone skills and thus need to be able to write to the TBeing data), we need it not to be a const.  I pulled this const_cast code from elsewhere in the Garble code, as this situation has come up previously and this is how they worked around it (while acknowledging it's not clean).

Now with that said... probably the best way to do this is to refactor the relevant Garble functions not to pass in TBeing *from as a const.  I don't know if that would have other repercussions, but I'm guessing since this is the 2nd time it's being recast, that's probably the way we should go.  I don't mind attempting to tackle that, but please let me know if you see any problems going that route before I do.

parkscs: On deeper inspection, the rabbit hole goes deeper than I want to refactor right now.  To remove the const_casts from garble (and mine is only 1 of several), we would need to refactor not just all of garble but also several core info.cc functions.  I think before making that many changes, we should first consider whether it may be easier to just overhaul the communication framework to follow a more modern design - in which case we'd just be rewriting a lot of it anyways.  As such, I'd like to hold off removing the const_cast for now.

code/code/misc/spells.h
```diff
@@ -591,6 +591,7 @@ enum spellNumT {
      SKILL_BULLYWUGCROAK,
      SKILL_AVIAN,
      SKILL_FISHBURBLE,
+     SKILL_COMMON,
```
cizra, line 594: Will injecting an item into the middle of an enum break learnedness data in player files?

drewhershey:
I have a branch in progress to deal with all those unnecessary constructors in various classes, so need to deal with them in this PR.

cizra (dismissed):

code/code/disc/disc_adv_adventuring.h
```diff
@@ -35,7 +36,9 @@ class CDAdvAdventuring : public CDiscipline {
       skTrollish(),
       skBullywug(),
       skAvian(),
-      skKalysian() {}
+      skKalysian(),
+      skCommon() {}
```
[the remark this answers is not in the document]
drewhershey: Oh yeah, I can see why that'd be a bit nicer. I'll keep that in mind for the future. All these pointless constructors will be gone at some point, too.

code/code/disc/disc_psionics.cc
```diff
@@ -98,12 +93,21 @@ int TBeing::doPTell(const char* arg, bool visible) {
     sendTo("What a dumb master you have, charmed mobiles can't tell.\n\r");
     return FALSE;
   }
-  half_chop(arg, name, message);
 
-  if (!*name || !*message) {
+  char n[100];
+  char m[MAX_INPUT_LENGTH + 40];
+  half_chop(arg, n, m);
```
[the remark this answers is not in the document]
drewhershey: Good call. I figured there was probably something in sstring that would do what half_chop was doing, but decided I'd rather refactor half-chop out of existence from everywhere all at once instead, in a future PR.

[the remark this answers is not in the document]
drewhershey: And yeah, using single character variable names is dumb haha, but I figured since the scope of their use was so brief and it was so obvious what they were doing that in this case it was probably acceptable. Plus they'll be gone once half-chop isn't used any longer.

wershlak (approved):

PR #373, opened 2021-08-10
Refactor the doFlee function to improve logic/remove redundancy

drewhershey:
This PR completely refactors the doFlee function that handles flee/retreat. It removes the gratuitous `for` loops used throughout previously and streamlines everything while using more modern syntax/features of C++.

The overall behavior of flee is pretty much identical to before. One new feature added was the ability for a Deikhan to resist falling off their mount if the mount panics and tries to flee. 

As coded this PR allows someone without retreat or who fails retreat a 50% chance to flee in the direction they choose. This can easily be adjusted if others feel 50% is too high of a chance.

I also refactored the lookForEngaged function since it's called in doFlee. The only real change there was updated syntax plus the removal of code that made unintelligent mobs not attack someone who was only engaged after the tank fled, which would leave the engaged person still "fighting" the mob but the mob not fighting them back.

### DIFF
--- a/code/code/misc/monster.cc
+++ b/code/code/misc/monster.cc
@@ -514,10 +514,7 @@ int TMonster::getMobDamage() const
 }
 
 int TMonster::lookForEngaged() {
-  if (isPc())
-    return false;
-
-  if (attackers > 0)
+  if (isPc() || attackers > 0)
     return false;
 
   auto isEngagedOpponent = [this](TThing *thingInRoom) {
@@ -525,24 +522,23 @@ int TMonster::lookForEngaged() {
     return thingAsTBeing && thingAsTBeing->fight() == this;
   };
 
-  auto found = std::find_if(roomp->stuff.begin(), roomp->stuff.end(), isEngagedOpponent);
+  auto* newOpponent =
+    dynamic_cast<TBeing*>(roomp->findInRoom(isEngagedOpponent));
 
-  TBeing *newOpponent = found != roomp->stuff.end() ? dynamic_cast<TBeing *>(*found) : nullptr;
+  if (!newOpponent)
+    return false;
 
-  if (newOpponent) {
-    int rc = takeFirstHit(*newOpponent);
+  int rc = takeFirstHit(*newOpponent);
 
-    if (IS_SET_DELETE(rc, DELETE_VICT)) {
-      delete newOpponent;
-      newOpponent = nullptr;
-    }
-
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return DELETE_THIS;
-
-    return true;
+  if (IS_SET_DELETE(rc, DELETE_VICT)) {
+    delete newOpponent;
+    newOpponent = nullptr;
   }
-  return false;
+
+  if (IS_SET_DELETE(rc, DELETE_THIS))
+    return DELETE_THIS;
+
+  return true;
 }
 
 bool TMonster::isShopkeeper() const {
--- a/code/code/misc/monster.cc
+++ b/code/code/misc/monster.cc
@@ -514,12 +514,14 @@ int TMonster::getMobDamage() const
 }
 
 int TMonster::lookForEngaged() {
-  if (isPc()) return false;
+  if (isPc())
+    return false;
 
-  if (attackers > 0) return false;
+  if (attackers > 0)
+    return false;
 
   auto isEngagedOpponent = [this](TThing *thingInRoom) {
-    auto thingAsTBeing = dynamic_cast<TBeing *>(thingInRoom);
+    TBeing *thingAsTBeing = dynamic_cast<TBeing *>(thingInRoom);
     return thingAsTBeing && thingAsTBeing->fight() == this;
   };
 
@@ -529,15 +531,15 @@ int TMonster::lookForEngaged() {
   TBeing *newOpponent = found != roomp->stuff.end() ? dynamic_cast<TBeing *>(*found) : nullptr;
 
   if (newOpponent) {
-    auto rc = takeFirstHit(*newOpponent);
+    int rc = takeFirstHit(*newOpponent);
 
     if (IS_SET_DELETE(rc, DELETE_VICT)) {
       delete newOpponent;
       newOpponent = nullptr;
     }
 
-    if (IS_SET_DELETE(rc, DELETE_THIS)) 
-      return DELETE_THIS;   
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_THIS;
 
     return true;
   }
--- a/code/code/misc/monster.cc
+++ b/code/code/misc/monster.cc
@@ -513,33 +513,32 @@ int TMonster::getMobDamage() const
   return max(1, idamrnd);
 }
 
-int TMonster::lookForEngaged(const TBeing *previousAttacker) {
+int TMonster::lookForEngaged() {
   if (isPc()) return false;
 
-  // if someone is attacking me directly, bypass
   if (attackers > 0) return false;
 
-  for (auto thing : roomp->stuff) {
-    if (thing == this ||
-        (previousAttacker && dynamic_cast<const TBeing *>(thing) == previousAttacker))
-      continue;
+  auto isEngagedOpponent = [this](TThing *thingInRoom) {
+    auto thingAsTBeing = dynamic_cast<TBeing *>(thingInRoom);
+    return thingAsTBeing && thingAsTBeing->fight() == this;
+  };
 
-    auto thingAsTBeing = dynamic_cast<TBeing *>(thing);
-    if (!thingAsTBeing) continue;
+  auto stuff = roomp->stuff;
+  auto found = std::find_if(stuff.begin(), stuff.end(), isEngagedOpponent);
+  auto newOpponent = found != stuff.end() ? dynamic_cast<TBeing *>(*found) : nullptr;
 
-    if (thingAsTBeing->fight() == this) {
-      auto rc = takeFirstHit(*thingAsTBeing);
+  if (newOpponent) {
+    auto rc = takeFirstHit(*newOpponent);
 
-      if (IS_SET_DELETE(rc, DELETE_VICT)) {
-        delete thingAsTBeing;
-        thingAsTBeing = nullptr;
-      }
-
-      if (IS_SET_DELETE(rc, DELETE_THIS))
-        return DELETE_THIS;
-      
-      return true;
+    if (IS_SET_DELETE(rc, DELETE_VICT)) {
+      delete newOpponent;
+      newOpponent = nullptr;
     }
+
+    if (IS_SET_DELETE(rc, DELETE_THIS)) 
+      return DELETE_THIS;   
+
+    return true;
   }
   return false;
 }
--- a/code/code/misc/monster.cc
+++ b/code/code/misc/monster.cc
@@ -523,9 +523,10 @@ int TMonster::lookForEngaged() {
     return thingAsTBeing && thingAsTBeing->fight() == this;
   };
 
-  auto stuff = roomp->stuff;
-  auto found = std::find_if(stuff.begin(), stuff.end(), isEngagedOpponent);
-  auto newOpponent = found != stuff.end() ? dynamic_cast<TBeing *>(*found) : nullptr;
+  StuffList::iterator found =
+      std::find_if(roomp->stuff.begin(), roomp->stuff.end(), isEngagedOpponent);
+
+  TBeing *newOpponent = found != roomp->stuff.end() ? dynamic_cast<TBeing *>(*found) : nullptr;
 
   if (newOpponent) {
     auto rc = takeFirstHit(*newOpponent);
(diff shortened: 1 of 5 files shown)


cizra (changes requested):

code/code/misc/monster.cc
```diff
@@ -513,45 +513,37 @@ int TMonster::getMobDamage() const
...
+
+  for (auto thing : roomp->stuff) {
+    if (thing == this ||
+        (previousAttacker && dynamic_cast<const TBeing *>(thing) == previousAttacker))
       continue;
-    if (tbt->fight() == this) {
-      // I'm a smart mob, and, oh look, tbt is engaged with me...
-      int rc = takeFirstHit(*tbt);
+
+    auto thingAsTBeing = dynamic_cast<TBeing *>(thing);
```
cizra, line 527: Perhaps we can avoid the double cast by using `thingAsTBeing` on line  524

code/code/misc/monster.cc
```diff
@@ -513,45 +513,37 @@ int TMonster::getMobDamage() const
...
+    if (!thingAsTBeing) continue;
+
+    if (thingAsTBeing->fight() == this) {
+      auto rc = takeFirstHit(*thingAsTBeing);
+
       if (IS_SET_DELETE(rc, DELETE_VICT)) {
-        delete tbt;
-        tbt = NULL;
+        delete thingAsTBeing;
```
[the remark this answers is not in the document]
cizra: lookForEngaged is doing 2 things. Instead, it should just look through the room and find a mob attacking this, and return it. Then code somewhere else could hit that mob (possibly another method, or possibly just `doFlee`)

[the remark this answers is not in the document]
drewhershey: After working on this some more, I realized that it was best for lookForEngaged to remain its own method in the TMonster class, because - though it's not coded that way now nor was it before - lookForEngaged _should_ also be called after a mob's opponent dies. 

Currently when that happens, if the mob has no active attackers when their opponent dies but there's someone engaged with them, the mob won't attack the engager but the engager will still be in combat.

I'm not familiar with the code that executes on mob/player death yet, so I'll try to address calling lookForEngaged there in a separate PR and just leave the on-death code unchanged for now.

As for the lookForEngaged code - I rewrote the function to use std::find_if with a lambda unary predicate to search for a potential opponent, which eliminated the need for two (nested) `for` loops. Everything still worked fine during my testing, so this should be a solid upgrade as far as syntax goes.

I also turned the code that handled the troglodyte racial perk into a lambda and was able to eliminate some code repetition.

code/code/misc/monster.cc
drewhershey: Resolved in https://github.com/sneezymud/sneezymud/pull/373/commits/a9d0e2528b68faa35f2ebc84115e123950d13d6f

cizra (dismissed):

code/code/misc/offense.cc
```diff
@@ -766,332 +766,281 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
-    if (canFleeThisWay(this, attempt)) {
-      if (panic || !doesKnowSkill(skill) || 
-          !bSuccess(skill)) {
-        act("$n panics, and attempts to flee.", TRUE, this, 0, 0, TO_ROOM);
-        panic = TRUE;
-      } else {
-        act("$n skillfully retreats from battle", TRUE, this, 0, 0, TO_ROOM);
-        panic = FALSE;
-      }
+    handleTrogRacial();
```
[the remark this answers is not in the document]
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/373/commits/164366be0c1df5205f7908c1c301c57ec071d589

code/code/misc/offense.cc
```diff
@@ -766,332 +766,283 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
+      addToWait(combatRound(1));
+      panic = true;
+    }
+    // If <this> is a mount being ridden
+  } else if (riderAsTBeing) {    
+    act("Your $O panics and attempts to flee!", true, riderAsTBeing, nullptr, this, TO_CHAR);
+
+    // Allow deikhans to pass checks and prevent getting bucked
+    if (
+      riderAsTBeing->doesKnowSkill(SKILL_CHIVALRY) &&
```
parkscs: Gameplay tweak/question - but would it make sense to also include any advanced riding and/or mount-specific mastery skill in the calculation?  We're bucking off the rider if they fail basic riding and chivalry checks, but it seems to me that someone who knows not just riding and chivalry but also has mastered ride mount_type (e.g., ride domestic) as well as advanced riding should get some bonuses to these checks.

code/code/misc/offense.cc
```diff
@@ -766,332 +766,283 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
+  }
+
+  // If there are no valid flee directions, simply return with the panic
+  // message.
+  if (validDirections.size() == 0) {
     sendTo("PANIC! You couldn't escape!\n\r");
-    return TRUE;
+
+    // Handle troglodyte racial perk
+    if (!::number(0, 1) && getMyRace()->hasTalent(TALENT_MUSK) && getCond(FULL) > 5) {
```
parkscs: Mark this resolved - just a note that I had NO IDEA musk was a thing.

code/code/misc/monster.cc
```diff
@@ -513,44 +513,37 @@ int TMonster::getMobDamage() const
...
-      }
-      return TRUE;
+    return false;
+
+  auto isEngagedOpponent = [this](TThing *thingInRoom) {
+    auto *thingAsTBeing = dynamic_cast<TBeing *>(thingInRoom);
+    return thingAsTBeing && thingAsTBeing->fight() == this;
+  };
+
+  auto found = std::find_if(roomp->stuff.begin(), roomp->stuff.end(), isEngagedOpponent);
```
cizra, line 528: might as well teach TRoom to search by lambda

code/code/misc/offense.cc
```diff
@@ -766,332 +766,301 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
+    }
   }
+
   if (isAffected(AFF_CHARM) && master && sameRoom(*master)) {
     if (!::number(0, 5))
-      act("$n bursts into tears.", TRUE, this, 0, 0, TO_ROOM);
-    act("You burst into tears at the thought of leaving $N.", 
-            FALSE, this, 0, master, TO_CHAR);
-    return FALSE;
+      act("$n bursts into tears at the thought of fleeing without $N!", true, this, nullptr, master,
```
cizra, line 825: true, this, nullptr, master, TO_ROOM - no idea what these are, except maybe for master and TO_ROOM.
I'd look for ways to create a helper around act() that reduces the number of things passed.

code/code/misc/offense.cc
```diff
@@ -766,332 +766,301 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
+      addToWait(combatRound(1));
+      panic = true;
+    }
+    // If <this> is a mount being ridden
+  } else if (riderAsTBeing) {
+    act("Your $O panics and attempts to flee!", true, riderAsTBeing, nullptr, this, TO_CHAR);
+    act("$n's $O panics and attempts to flee!", true, riderAsTBeing, nullptr, this, TO_ROOM);
+
+    // Allow deikhans to pass checks and prevent getting bucked
+    if (riderAsTBeing->doesKnowSkill(SKILL_CHIVALRY) && riderAsTBeing->bSuccess(SKILL_CHIVALRY) &&
```
cizra, line 891: why doesn't bSuccess automatically check for skill knowledge?

code/code/misc/offense.cc
```diff
@@ -766,332 +766,301 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
-            return TRUE;
-          }
-        }
+      // Handle troglodyte racial
+      if (panic && !::number(0, 1) && getMyRace()->hasTalent(TALENT_MUSK) && getCond(FULL) > 5) {
+        act("Your fear causes you to release some musk scent to cover your tracks.", false, this, nullptr,
+            nullptr, TO_CHAR);
+        act("$n releases some musk into the room!", false, this, nullptr, nullptr, TO_ROOM);
+        dropGas(::number(1, 3), GAS_MUSK);
+        setCond(FULL, getCond(FULL) - 5);
```
cizra, line 983: could have a consumeFullness or something. Not sure it'd be worth the extra lines tho.

code/code/misc/offense.cc
```diff
@@ -766,332 +766,301 @@ static bool canFleeThisWay(TBeing *ch, dirTypeT dir)
...
+  if (enemy->fight() == this)
+    enemy->stopFighting();
+
+  // Make mobs fear enemy who forced them to flee
+  // Check IsPc() in case of disguised/polymorphed/otherwise transformed PC
+  TMonster *thisAsTMonster = isPc() 
+      ? nullptr 
+      : dynamic_cast<TMonster *>(this);
+
+  if (thisAsTMonster)
```
cizra, line 1036: why not `if (!isPc()) then dynamic cast and proceed?

cizra (dismissed):

code/code/misc/offense.cc
drewhershey: Definitely makes sense. I don't know how common being bucked off actually is, so initially I was just thinking "don't let Deikhans be easily bucked off because that sucks for gameplay" but it probably should be more specific to the type of mount. I'll look into it.

code/code/misc/offense.cc
drewhershey: Not sure it actually does anything useful, even, but it was used this way previously so I figured we should keep it around. 👍

code/code/misc/offense.cc
drewhershey: So after spending a little time thinking on this one, I'm not sure it's really within the scope of this PR. `act` is used hundreds (if not thousands) of times throughout the codebase in this way, so improving its usage here would only be addressing a small fraction of the problem.

It honestly is a very kludgy function - so much so that it has [its own page in the repo wiki](https://github.com/sneezymud/sneezymud/wiki/act()-format-guide) explaining how to use it. It needs a complete update at some point, instead of just a band-aid fix, IMO. If it were re-written to accept a variant in a couple places it could remove the need for a helper function at all.

If you really do want it addressed to make this PR mergeable any suggestions for how you'd go about writing the helper function would be welcome, as I'm struggling to come up with an idea that doesn't turn into just rewriting the function itself.

code/code/misc/monster.cc
```diff
@@ -513,44 +513,37 @@ int TMonster::getMobDamage() const
...
-int TMonster::lookForEngaged(const TBeing *ch)
-{
-  // here's some AI
-  // tank and healer fight mob, tank flees, mob should now beat on
-  // healer (regardless of whether he engaged or not?)
+int TMonster::lookForEngaged() {
   if (isPc())
-    return FALSE;
-  // if someone is attacking me directly, bypass
+    return false;
```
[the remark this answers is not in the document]
drewhershey: I was following the convention used throughout the codebase - tons of functions rely on implicit boolean conversions to int when the function could possibly return a DELETE_X type value. I've never been clearly exactly why they decided to do it that way. `doFlee` itself does the same.

How would you prefer I handle this?

code/code/misc/offense.cc
cizra: No, I'm definitely not blocking the PR behind that.

code/code/misc/monster.cc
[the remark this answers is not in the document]
cizra: Perhaps... perhaps an enum class would be more appropriate? It'd also give the compiler opportunity to verify calling code covers the cases.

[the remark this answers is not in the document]
drewhershey: That seems like a much better way to do it. It would be a big project, as we'd be refactoring a lot of code. I opened https://github.com/sneezymud/sneezymud/issues/396 to track this, as it should definitely be its own project, and I'll resolve this conversation here.

code/code/misc/offense.cc
drewhershey: Just added https://github.com/sneezymud/sneezymud/issues/397 to track this. It's definitely a massive project.

code/code/misc/offense.cc
drewhershey: In interest of getting this PR done with I'm going to just agree with you and say it's not worth the extra lines. 😁

code/code/misc/offense.cc
drewhershey: It does, actually, and I'm not sure why we don't rely on it more in the code. Addressed this in https://github.com/sneezymud/sneezymud/pull/373/commits/9eafa770ec48ccd42c3287b72176a0950acf293e.

code/code/misc/offense.cc
drewhershey: This now instead relies on passing bSuccess checks for both the advanced riding skill for the type of mount being ridden and the calm mount skill. See https://github.com/sneezymud/sneezymud/pull/373/commits/9eafa770ec48ccd42c3287b72176a0950acf293e.

cizra (approved):

PR #378, opened 2021-08-22
WIP: Add holy

wershlak:
FYI I don't intend to merge this yet. I just want to get feedback on a particular point.

@drewhershey @parkscs check out the comment here:

https://github.com/sneezymud/sneezymud/blob/47f2fb9c9560640a6a7504538f36e1aeda84c19c/code/code/misc/immunity.h#L18-L50

TO me, holy feels like a missing damage type. My intent was to use this for a new harm line replacement but it also works for things like smite and deikhan swords which are currently drain and thus do like no damage to undead (lol).

Since plus2 and plus3 aren't used now this seems fine to convert one to holy.

Since we can't expand the number of immunity types beyond 28 this change will leave us with only a couple extra types left. Anyone think it's a problem?


Durhamite2118:
Looks good from my view, I like the holy addition.

PR #385, opened 2021-08-31
allow bash to do damage

wershlak:
I think bash should do at least some small wimpy damage rather than the current zero.

We can tune the damage later but this at least puts it through the standard formula.


wershlak:
@drewhershey I fixed the damage type here

drewhershey (approved):

Issue #374, opened 2021-08-15
Tweak DR


wershlak:
```
> gamestats combat
Total damage taken  : 2134724
        Mob damage taken    : 2086610   (97.75%)
        PC  damage taken    : 48114   ( 2.25%)
Combat damage only  : 1778370
        Mob combat damage taken    : 1735138   (97.57%)
        PC combat damage taken     : 43232   ( 2.43%)
AC  damage absorbed : 38113
        Mob AC absorb       : 3457   ( 9.07% of all abs) ( 0.20% of mob ComDam)
        PC AC absorb        : 34656   (90.93% of all abs) (44.49% of PC ComDam)
Current absorbtion constants : TBeing 2, PC 4

Total Combat blows        : 182959
        Mob combat blows           : 128617   (70.30%)
        PC combat blows            : 54342   (29.70%)
Total Combat hits         : 109857  (60.04%)
        Mob combat hits            : 75468   (68.70%)  (58.68% hit rate)
        PC combat hits             : 34389   (31.30%)  (63.28% hit rate)
Average Combat level      : 41.91
        Mob combat level           : 42.98
        PC combat level            : 39.36
Average Combat damage     : 16.19
        Mob avg. combat dam.       : 22.99
        PC avg. combat dam.        :  1.26
Total crit-success checks : 12186  (11.09% of hits)
Total crit-fail checks    : 8973  (12.27% of misses)
Total crit-success passes : 515  ( 0.47% of hits)
Total crit-fail passes    : 134  ( 0.18% of misses)

Mobiles have tried to aggro : 531 times.
Mobiles have aggro'd        : 185 times.
```

Here's the function for AC based damage absorption:
https://github.com/sneezymud/sneezymud/blob/master/code/code/misc/combat.cc#L1334

It is tweakable via a code change.

Based on the numbers above and looking at the formulas I would guess we could reduce this by half without impacting gameplay all that much.

I would also contend we could include two additional modifiers here.

1. determine the type of item in the slot and reduce the absorption if it is anything other than armor. This seems both realistic and fair and will further differentiate armor from clothing and make for some more interesting choices in gearing up.
  - cloth becomes 50-70% of the absorption it's AC would grant
  - Jewelry, bags, spell bags or anything else that is worn but not armor or worn 10-30% of absorption.

2. Add a small factor for advanced defense here. The skill should convey that an expertly trained armor wearer gets more out of the armor. We will do something similar with defend_round but I think it works well here too.

[...]

PR #386, opened 2021-09-03
Adjust DR absorption 

wershlak:
Adjust DR dependent on item type and advanced defense
- ARMOR 100% or 120% with adv defense
- WORN 50% or 60% with adv defense
- other 0%

### DIFF
--- a/code/code/misc/combat.cc
+++ b/code/code/misc/combat.cc
@@ -1333,17 +1333,34 @@ bool TObj::makeScraps()
 
 static void absorb_damage(TBeing *v, wearSlotT part_hit, int *dam)
 {
-  int armor;
-  int abs;
+  TObj *equip = nullptr;
+  TThing *t;
+  int armor = v->acForPos(part_hit) / stats.absorb_damage_divisor[v->isPc() ? PC_STAT : MOB_STAT];
+  int absorb = min(armor/10, *dam);
 
-  armor = v->acForPos(part_hit) / stats.absorb_damage_divisor[v->isPc() ? PC_STAT : MOB_STAT];
-  abs = min(armor/10, *dam);
-  stats.ac_absorb[v->isPc() ? PC_STAT : MOB_STAT] += abs;
+  if (v->isPc()) {
+    if (!(t = v->equipment[part_hit]) || !(equip = dynamic_cast<TObj *>(t)))
+      absorb = 0;
+    else {
+      if (equip->itemType() == ITEM_WORN) {
+        absorb /= 2;
+      } else if (equip->itemType() != ITEM_ARMOR) {
+        absorb = 0; 
+      }
+    }
+  }
+
+  // 20% bonus for advanced defense
+  if (v->doesKnowSkill(SKILL_ADVANCED_DEFENSE)) {
+    absorb *= (1 + (0.002 * v->getSkillValue(SKILL_ADVANCED_DEFENSE)));
+  }
+
+  stats.ac_absorb[v->isPc() ? PC_STAT : MOB_STAT] += absorb;
 
 #if DEBUG
   vlogf(LOG_COMBAT, format("Debug : Damage went from %d to %d with armor absorption!") %  *dam % (*dam - abs));
 #endif
-  *dam -= abs;
+  *dam -= absorb;
 }
 
 // this = v, ch = hitter
--- a/code/code/misc/statistics.cc
+++ b/code/code/misc/statistics.cc
@@ -62,7 +62,7 @@ int init_game_stats(void)
   char buf[256];
 
   stats.absorb_damage_divisor[MOB_STAT] = 2;
-  stats.absorb_damage_divisor[PC_STAT] = 4;
+  stats.absorb_damage_divisor[PC_STAT] = 5;
 
   //stats.equip = 0.7;   // this affects the global load potential of things
   //stats.global_lp_target = stats.equip;  // When we want to change the load potential, this is what we change.


code/code/misc/combat.cc
```diff
@@ -1333,17 +1333,34 @@ bool TObj::makeScraps()
...
 {
-  int armor;
-  int abs;
+  TObj *equip = nullptr;
+  TThing *t;
+  int armor = v->acForPos(part_hit) / stats.absorb_damage_divisor[v->isPc() ? PC_STAT : MOB_STAT];
+  int absorb = min(armor/10, *dam);
+
+  if (v->isPc()) {
+    if (!(t = v->equipment[part_hit]) || !(equip = dynamic_cast<TObj *>(t)))
```
cizra, line 1342: Is it possible to wear something other than objects?

code/code/misc/combat.cc
```diff
@@ -1333,17 +1333,34 @@ bool TObj::makeScraps()
...
+      absorb = 0;
+    else {
+      if (equip->itemType() == ITEM_WORN) {
+        absorb /= 2;
```
cizra, line 1346: You're later on doing float math on absorb, perhaps we want float division here?

code/code/misc/combat.cc
```diff
@@ -1333,17 +1333,34 @@ bool TObj::makeScraps()
...
-  armor = v->acForPos(part_hit) / stats.absorb_damage_divisor[v->isPc() ? PC_STAT : MOB_STAT];
-  abs = min(armor/10, *dam);
-  stats.ac_absorb[v->isPc() ? PC_STAT : MOB_STAT] += abs;
+  stats.ac_absorb[v->isPc() ? PC_STAT : MOB_STAT] += absorb;
 
 #if DEBUG
   vlogf(LOG_COMBAT, format("Debug : Damage went from %d to %d with armor absorption!") %  *dam % (*dam - abs));
 #endif
-  *dam -= abs;
+  *dam -= absorb;
```
cizra, line 1363: What conceivable reason would there be to prefer modifying out-params, instead of returning a value?

code/code/misc/combat.cc
```diff
@@ -1333,17 +1333,34 @@ bool TObj::makeScraps()
 
 static void absorb_damage(TBeing *v, wearSlotT part_hit, int *dam)
```
cizra, line 1334: I don't see this code modifying `v`, let's constify it.

cizra (changes requested):

drewhershey:
Here's the results of some testing I did by compiling this branch and checking values in gdb:

(I equipped a set of gleaming and had Slojup (lvl 117) attack me)
- *dam = 11 (first hit of the fight, so damage is at its lowest at this point)
- part_hit was WEAR_BODY
- acForPos(part_hit) = 263 (gleaming suit)
- armor = 52 after line 1338
- absorb = 5 after line 1339
- absorb = 6 after line 1355 (with 100% advanced defense/skill maxed)

One issue I see is that the body part being hit has to have AC of at least 250 to actually receive any benefit from advanced defense, even when it's maxed. 

Because absorb is an `int`, any value < 5 will end up with the bonus from advanced defense being a decimal and therefore will get dropped. (e.g. 4 * 1.2 = 4.8 = 4 when cast to int). An absorb of 5 will end up becoming 6 exactly with maxed adv. defense.

To determine the minimum AC needed to end up with an absorb of 5 before applying the advanced defense bonus, solve:
```
AC / DAMAGE_DIVISOR / 10 >= 5

which becomes:
AC / 5 / 10 >= 5
AC >= 5 * 5 * 10 = 250
```
Even if the bonus from adv. defense were applied to the AC value before other calculations it only lowers the minimum AC threshold to 208:
```
AC * 1.2 / 5 / 10 >= 5

AC >= 5 * 5 * 10 / 1.2 = 208
```

I'm not really sure what to take from these numbers but figured I'd just share them to help with refining the formula.

parkscs:
This is more of a gameplay comment than a comment on the specific code.

I have real concerns about adding absorbs in for tanks.  They are incredibly powerful in a lot of fights and while we may not think absorbing just a few damage is a lot, it really is.  This is especially true against monk NPCs who hit lots of times for just a few damage per hit.  To give you an idea, I recently died fighting against 3 guards in Gnomish Wonders on two of my characters due to the NOFLEE room.... I took the same characters back but with an absorb shield that blocks 1-3 damage per hit and I was able to fight and win 10v1 against the same guards.  Absorbs are incredibly powerful because most enemies only hit for a few damage after DR is applied.  Tanks are already too tanky in the endgame IMO (I can give plenty of other examples as needed) and this would just further compound that issue.

If it was just up to me, I'd propose removing advanced defense as a skill and just letting the other skills in the defensive discipline justify the practices.  But if we're determined to keep it, one thought I had from talking with drew would be to factor advanced defense into the defendRound calculation much like defense is factored in.  This would still result in making warriors tankier, but the true "end game" NPCs would still be able to hit them reliably and unless it's a really massive bonus, it wouldn't be as powerful as an absorb can be.

code/code/misc/combat.cc
wershlak: You are right on all your points. I focused on updating the business logic and forgot to ask myself if the code sucks. Would this be better as:
- a member function of Tbeing that accepts part_hit and returns an absorb amount
- an anonymously namespaced function in combat.cc that accepts a const victim and part hit and returns an absorb amount?

wershlak:
> Here's the results of some testing I did by compiling this branch and checking values in gdb:
> 
> (I equipped a set of gleaming and had Slojup (lvl 117) attack me)
> 
> * *dam = 11 (first hit of the fight, so damage is at its lowest at this point)
> * part_hit was WEAR_BODY
> * acForPos(part_hit) = 263 (gleaming suit)
> * armor = 52 after line 1338
> * absorb = 5 after line 1339
> * absorb = 6 after line 1355 (with 100% advanced defense/skill maxed)
> 
> One issue I see is that the body part being hit has to have AC of at least 250 to actually receive any benefit from advanced defense, even when it's maxed.
> 
> Because absorb is an `int`, any value < 5 will end up with the bonus from advanced defense being a decimal and therefore will get dropped. (e.g. 4 * 1.2 = 4.8 = 4 when cast to int). An absorb of 5 will end up becoming 6 exactly with maxed adv. defense.
> 
> To determine the minimum AC needed to end up with an absorb of 5 before applying the advanced defense bonus, solve:
> 
> ```
> AC / DAMAGE_DIVISOR / 10 >= 5
> 
> which becomes:
> AC / 5 / 10 >= 5
> AC >= 5 * 5 * 10 = 250
> ```
> 
> Even if the bonus from adv. defense were applied to the AC value before other calculations it only lowers the minimum AC threshold to 208:
> 
> ```
> AC * 1.2 / 5 / 10 >= 5
> 
> AC >= 5 * 5 * 10 / 1.2 = 208
> ```
> 
> I'm not really sure what to take from these numbers but figured I'd just share them to help with refining the formula.

Yeah I need to go back to the drawing board here. I'm unsure if I should reduce DR for armor vs clothing vs other OR let advanced defense play here. Seems like I can do one or the other. Probably, armor?

One thing that bugs me is the DR is based on the slot and I kind of feel like it should be based on total armor and be a consistent DR for each and every hit. Why, reduce body damage more than leg damage because the slot is more important?

Keep in mind none of this needs to be realistic. It's a game. So what is interesting and fair?

wershlak:
> This is more of a gameplay comment than a comment on the specific code.
> 
> I have real concerns about adding absorbs in for tanks. They are incredibly powerful in a lot of fights and while we may not think absorbing just a few damage is a lot, it really is. This is especially true against monk NPCs who hit lots of times for just a few damage per hit. To give you an idea, I recently died fighting against 3 guards in Gnomish Wonders on two of my characters due to the NOFLEE room.... I took the same characters back but with an absorb shield that blocks 1-3 damage per hit and I was able to fight and win 10v1 against the same guards. Absorbs are incredibly powerful because most enemies only hit for a few damage after DR is applied. Tanks are already too tanky in the endgame IMO (I can give plenty of other examples as needed) and this would just further compound that issue.
> 
> If it was just up to me, I'd propose removing advanced defense as a skill and just letting the other skills in the defensive discipline justify the practices. But if we're determined to keep it, one thought I had from talking with drew would be to factor advanced defense into the defendRound calculation much like defense is factored in. This would still result in making warriors tankier, but the true "end game" NPCs would still be able to hit them reliably and unless it's a really massive bonus, it wouldn't be as powerful as an absorb can be. Another option would be to apply the absorb before any damage reduction is applied.

I don't think you actually read the code? I didn't add any DR. It exists today and this PR reduces the amount that players will absorb in total.

That said, I'm going to completely rework it so just hold on.

parkscs:
I didn't look at the code - just talked to Drew about his testing of it and the overall design.  I guess the point I was trying to make which seems to have not come across is that damage reduction, even by 1, is very powerful if it's applied in the same way that plasma shield does it (which is after other forms of damage reduction).  If we're reducing raw damage by this amount, that's certainly less impactful (although even 1 point per attack is significant), but my point is just to caution about giving tanks any sort of raw damage absorb similar to plasma shield because it's incredibly powerful and they don't really take enough damage from melee NPCs in the first place.

Aion2501:
@wershlak 

When we first discussed this didn't we decide to just add flat % DR for total AC values instead of chance to miss?
That would make it fair whether facing a monk, warrior, or any class hitting.

If you want to make it more complex you could add diminishing returns for higher values of total armor and a cap.

wershlak:
Ok you all make some good points. 

@parkscs - Yeah raw damage reduction is powerful. It's in the game now though - See issue https://github.com/sneezymud/sneezymud/issues/374

This absorption function is called in combat where attack damage is applied. Check out the current constant and formula and see that if you take a diamond suit for example: 

300AC / 4 / 10 = 7.5 absorption on every body hit for a player wearing this armor.

That's the best case scenario (for the player) but based on gamestats you can see players are absorbing a LOT of damage. It's too much and we should do something about it.

my intention with the PR  (which I should have been more clear about) is:

1. Reduce the amount of DR that players receive
    - To accomplish this I increased the stats.absorb_damage_divisor[PC_STAT] 
2. Differentiate more between classes/roles in Sneezy
    - To accomplish this I added differentiation between armor and cloth and advanced def

I no longer think this is a great formula but I still have the same goals so it warrants further discussion

--

@Aion2501 - a percent based would be more fair to monks and anything that hits for less damage in general. But the percent needs to be small.

--

### Background
DR sort of makes sense to me and better armor should provide better DR but from a gameplay perspective I don't think this works well for a few reasons.

- A flat number reduced means a few large hits are better than many small hits and limits viable gameplay styles. (monk damage is less impactful for instance)
- The number is so small (between 0-7.5) that we can't differentiate it in different ways. Example: skills would typically apply a percentage of their effect based on learnedness but when dealing with small integer amounts the math gets silly
- The per slot style of this bothers me for some reason.

### Options to fix it:

1. Remove this DR entirely. Players will take more damage for sure so maybe this path requires some testing.
2. reduce the amount by increasing stats.absorb_damage_divisor[PC_STAT]
3. Completely change how the formula works
  - Maybe just a flat number based on item type (1 for cloth 2 for armor)
  - Maybe just a flat number based on skills (1 for defense and 2 for advanced def) <- (this already sounds dumb)
  - Maybe a percent based on a combination

This last option is interesting and could be something like:
- If wearing cloth DR = 0-10% absorption based on defensive learnedness 
- If wearing armor DR = 0-15% absorption based on defensive learnedness
- An additional 0-5% absorption based on advanced defensive learnedness

*Of note: * we only care about the class of armor and not the actual quality
So a tank in armor will get a max of 20% absorption. Much less than now for smaller hits
Because it's percent based at least the small differences might work out to more damage absorbed over the length of fights and will be better as we scale up mob damage in other ways.

parkscs:
A flat % reduction would definitely avoid the issues that flat absorbs
create.  Without any buffs like sanctuary active, single hits from a level
~70 or so NPC in Gnomish Wonders were doing about 5 damage each to my level
50.  If you have a flat reduction of even 1 point of damage post-DR, that's
a 20% reduction in incoming damage.  Something like plasma shield that
reduce 1-3 damage are reducing incoming damage from that NPC by 20-60%,
which is pretty insane to me.  I certainly don't want to derail the
discussion here talking about plasma shield, but that spell and thornflesh
(which is just a restrung copy pretty much) are the only flat reduction
mechanics in the game (that I'm aware of) and are incredibly strong,
especially against monks (too strong imo).

code/code/misc/combat.cc
cizra: I'd choose the 2nd.

cizra:
Closing as obsolete.

PR #388, opened 2021-09-07
Crit changes rework (update)

parkscs:
Updating this to reflect some changes from before.  The main changes here are:
- Update crit severity calculations based on level to help encourage diversity of crits from enemies (as currently high level opponents land the same crit over and over again because they cap out on the crit severity chart).  To make up for the lesser severity, I'm proposing increasing the NPC crit crate based on level, so ultimately high level opponents will crit more often but will not constantly roll the most severe instakill crit over and over again.
- Update the pierce crit tree, with the goals being to a) overall improve the tree as currently pierce is significantly worse than all other damage types and b) move the most severe crits to the end of the tree in anticipation of a future crit rework.  Damage multipliers higher than 3 have been removed.

The new table for pierce crits is as follows:

0-33: double damage
34-66: triple damage
67,68,69,70,71: pierced larynx (3x damage)
72,73,74: gouged out eye (3x damage)
75,76,77,78: sever tendon (3x damage)
79,80,81: shatter elbow (3x damage)
82,83,84: sever hand (3x damage)
85,86,87: punctured lung (impale) (3x damage)
88,89,90: punctured kidney infect (impale) (3x damage)
91,92,93: punctured stomach (3x damage)
94,95,96,97: impale stun (3x damage)
98,99: pierce cranium (death/3x damage)
100: crit kill (death)

total:
  double damage: 33%
  triple damage: 67%
  limb mutilation (leg/elbow/hand): 10%
  body mutilation (lung/kidney/stomach): 9%
  debilitate (larynx/eye): 8%
  stun: 4%
  death: 3% (1% if helmet)

### DIFF
--- a/code/code/misc/crit_combat.cc
+++ b/code/code/misc/crit_combat.cc
@@ -1660,7 +1660,8 @@ buf=format("$n thrusts $s %s deep into $N's torso, impaling $M!") %
 	  desc->career.crit_impale++;
 	if (v->desc)
 	  v->desc->career.crit_impale_suff++;
-	rc = dislodgeWeapon(v, weapon, WEAR_BODY);
+
+	rc = damageLimb(v,WEAR_BODY,weapon,dam);
 	if (IS_SET_DELETE(rc, DELETE_VICT))
 	  return DELETE_VICT;
 	rc = applyDamage(v, v->hitLimit()/2,DAMAGE_IMPALE);
@@ -1886,25 +1887,26 @@ buf=format("$n attempts to decapitate $N with $s %s!  Luckily, $p saves $M!") %
 
 0-33: double damage
 34-66: triple damage
-67,68: pierced larynx
-69,70: gouged out eye
-71,72: sever tendon
-73,74: stab back (impale)
-75,76: pierce cranium
-77,78: shatter elbow
-79,80: sever hand
-81,82: punctured lung (impale)
-83,84,85: punctured kindey infect (impale)
-86,87: punctured stomach
+67,68,69,70,71: pierced larynx (3x damage)
+72,73,74: gouged out eye (3x damage)
+75,76,77,78: sever tendon (3x damage)
+79,80,81: shatter elbow (3x damage)
+82,83,84: sever hand (3x damage)
+85,86,87: punctured lung (impale) (3x damage)
+88,89,90: punctured kidney infect (impale) (3x damage)
+91,92,93: punctured stomach (3x damage)
+94,95,96,97: impale stun (3x damage)
+98,99: pierce cranium (death/3x damage)
 100: crit kill (death)
 
 total:
   double damage: 33%
-  triple damage: 33%
-  impale: 7%
-  minor sever/ailment: 11%
-  major sever/ailment: 8%
-  death: 3%
+  triple damage: 67%
+  limb mutilation (leg/elbow/hand): 10% 
+  body mutilation (lung/kidney/stomach): 9%
+  debilitate (larynx/eye): 8%
+  stun: 4%
+  death: 3% (1% if helmet)
 
 ------------------------------------------------------------ */
 int TBeing::critPierce(TBeing *v, TThing *weapon, wearSlotT *part_hit,
@@ -1958,8 +1960,14 @@ int TBeing::critPierce(TBeing *v, TThing *weapon, wearSlotT *part_hit,
 
     switch (crit_num) {
       case 67:
+      case 68:
+      case 69:
 	if (!v->hasPart(WEAR_NECK) || IS_SET(v->specials.act, ACT_GHOST) || IS_SET(v->specials.act, ACT_SKELETON))
 	  return 0;
+
+	// triple damage
+        *dam *= 3;
+
 	*part_hit = WEAR_NECK;
 	if ((obj = v->equipment[WEAR_NECK])) {
 	  v->sendTo(COLOR_OBJECTS, format("Your %s saves you from a punctured larynx!\n\r") %
@@ -1970,7 +1978,8 @@ int TBeing::critPierce(TBeing *v, TThing *weapon, wearSlotT *part_hit,
 	  return ONEHIT_MESS_CRIT_S;
 	}
 	// intentional drop through
-      case 68:
+      case 70:
+      case 71:
 	// Punctured Larnyx, can't speak 
 	if (!v->hasPart(WEAR_NECK) || IS_SET(v->specials.act, ACT_GHOST) || IS_SET(v->specials.act, ACT_SKELETON))
 	  return 0;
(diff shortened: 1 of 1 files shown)


code/code/misc/crit_combat.cc
```diff
@@ -559,6 +559,12 @@ int TBeing::critSuccessChance(TBeing *v, TThing *weapon, wearSlotT *part_hit, sp
   } else {
     double absdiff = abs(diff);
     double level_mod = 50.0 + ((double)diff * log(absdiff/20.0+1) / absdiff) * 75.0;
+
+    // Increasing scaling based on level difference for NPCs
+    if(dynamic_cast<TMonster *>(this)){
```
cizra, line 564: Instead of dynamic_cast, we could have a method `virtual getSkillLevelMod()` for `TMonster`, with `TPlayer` overriding it.

code/code/misc/crit_combat.cc
```diff
@@ -1654,7 +1660,8 @@ buf=format("$n thrusts $s %s deep into $N's torso, impaling $M!") %
 	  desc->career.crit_impale++;
 	if (v->desc)
 	  v->desc->career.crit_impale_suff++;
-	rc = dislodgeWeapon(v, weapon, WEAR_BODY);
+
+	rc = damageLimb(v,WEAR_BODY,weapon,dam);
```
cizra, line 1664: What's the story behind removing `dislodgeWeapon`?

code/code/misc/crit_combat.cc
```diff
@@ -1952,8 +1960,14 @@ int TBeing::critPierce(TBeing *v, TThing *weapon, wearSlotT *part_hit,
 
     switch (crit_num) {
       case 67:
+      case 68:
+      case 69:
 	if (!v->hasPart(WEAR_NECK) || IS_SET(v->specials.act, ACT_GHOST) || IS_SET(v->specials.act, ACT_SKELETON))
 	  return 0;
+
+	// triple damage
+        *dam *= 3;
```
cizra, line 1969: indent here and below

code/code/misc/crit_combat.cc
```diff
@@ -2344,24 +2320,113 @@ buf=format("$n's %s tears into $N's stomach exposing intestines!") %
...
+	  rc = damageLimb(v,WEAR_HEAD,weapon,dam);
+	  if (IS_SET_DELETE(rc, DELETE_VICT))
+	    return DELETE_VICT;
+	  *part_hit = WEAR_HEAD;
+	  rc = damageLimb(v,WEAR_HEAD,weapon,dam);
+	  if (IS_SET_DELETE(rc, DELETE_VICT))
+	    return DELETE_VICT;
+	  return ONEHIT_MESS_CRIT_S;
+	} else {
+buf=format("You thrust your %s into the back of $N's head causing an immediate death.") %
```
cizra, line 2406: indent

cizra (changes requested):

code/code/misc/crit_combat.cc
parkscs: It's for game balance reasons.  Pierce weapons are objectively worse than all other weapon types in part because you just lose your weapon during combat.  It's a "feature" no one likes as far as I'm aware and from a realism perspective it's just kind of stupid that knife fights would repeatedly lose their knives during a lengthy fight by impaling them inside their opponent.

code/code/misc/crit_combat.cc
```diff
@@ -2344,24 +2320,113 @@ buf=format("$n's %s tears into $N's stomach exposing intestines!") %
...
+	  *part_hit = WEAR_BACK;
+	  return ONEHIT_MESS_CRIT_S;
+	}
       case 96:
       case 97:
+	// Side wound, vict stunned 6 rounds. 
+	if (!v->hasPart(WEAR_BACK))
+	  return 0;
+	buf=format("You plunge your %s deep into $N's side, stunning $M!") %
+		limbStr;
```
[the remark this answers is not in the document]
parkscs: From 1959 - limbStr=(weapon ? fname(weapon->name) : getMyRace()->getBodyLimbPierce(this));

Pretty sure even though the existing code refers to it as limbStr throughout this file, it can be a weapon as well.

code/code/misc/crit_combat.cc
cizra: I'm not buying the bit with realism (I've yet to see a computer game of any kind that would be realistic in the slightest).
However, I do empathize with player irritation, so I'm open to killing it.
Seeing as it's a very user-visible change, have you run this by the design department yet?
Is anything else calling dislodgeWeapon now, or should we delete that code?

code/code/misc/crit_combat.cc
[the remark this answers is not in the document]
cizra: Yep, let's fix it to read `weaponStr`, as it's clearly meant to be a weapon (even if some races/classes use natural weapons).

code/code/misc/crit_combat.cc
parkscs: I think the gameplay concerns are reason enough to remove the dislodge effect from pierce crits; the comment on realism (or lack thereof) for constantly losing pierce weapons in a fight was simply the icing on the cake.

I've run these changes by others before submitting but I'd welcome feedback here if people have it, and I've asked for a few to chime in here with their feedback from a gameplay perspective.

As far as whether other functions are using it, I'll have to take a look and get back to you on that.

Durhamite2118: @cizra we've had several discussions on the removal of dislodge weapon.  The design department is for it and sees this as a minor tweak to address a major irritation.

code/code/misc/crit_combat.cc
parkscs: Before I attempt refactoring this - how big of an issue is this?  If it's a deal breaker I can give it a go, but looking at the file again it looks like dynamic_cast is called in this way a number of different times throughout the file to determine whether the individual in question is an NPC or a Player.  If it's not going to cause problems, I'd be inclined to suggest we leave it despite it being non-optimal.  If it may cause some issues, I can take a look into refactoring it though.  Let me know.

cizra: It's not any worse than surrounding code, so I'm not demanding you fix it. Not that the surrounding code is anything beautiful to behold.
If you want to know if somebody is a player, it'd make sense to ask them if they are - have a isPlayer method in TBeing, with default return false, return true for TPerson.
Please note that there are situations where TBeings are also players - for example if an immort switches into a mob.

cizra (approved):

parkscs:
Makes sense.  I'm inclined to leave it as is for now, simply because I'm
worried that in trying to clean things up, I will end up introducing bugs
into what is (as far as I'm aware) stable code.

PR #389, opened 2021-09-07
Retreat fix

parkscs:
Pulling this out separately from the crit changes, as it didn't really belong.  This change fixes a bug (as currently retreat chances go down as your skill improves) and reworks the algorithm for determining whether retreat has failed.  With retreat skill at max, the player will have a 95% chance to flee in their specified direction.

### DIFF
--- a/code/code/misc/offense.cc
+++ b/code/code/misc/offense.cc
@@ -952,8 +952,8 @@ int TBeing::doFlee(const char *arg)
   for (i = 0; i < 20; i++) {
     dirTypeT attempt = dirTypeT(::number(MIN_DIR, MAX_DIR-1));        // Select a random direction 
     
-    // fighting, so give slight chance of letting PC direct the direction
-    if (chosenDir != DIR_NONE && !panic && !(::number(0,99) < (30+getSkillValue(skill)/2)))
+    // fighting, so the chance of success depends on retreat skill (with at least 5% chance to fail)
+    if (chosenDir != DIR_NONE && !panic && (::number(0,99) < (45+getSkillValue(skill)/2)))
       attempt = chosenDir;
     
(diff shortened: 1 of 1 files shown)


cizra (approved):

code/code/misc/offense.cc
```diff
@@ -952,8 +952,8 @@ int TBeing::doFlee(const char *arg)
   for (i = 0; i < 20; i++) {
     dirTypeT attempt = dirTypeT(::number(MIN_DIR, MAX_DIR-1));        // Select a random direction 
     
-    // fighting, so give slight chance of letting PC direct the direction
-    if (chosenDir != DIR_NONE && !panic && !(::number(0,99) < (30+getSkillValue(skill)/2)))
+    // fighting, so the chance of success depends on retreat skill (with at least 5% chance to fail)
+    if (chosenDir != DIR_NONE && !panic && (::number(0,99) < (45+getSkillValue(skill)/2)))
```
cizra, line 956: The PR description says there's a bug where chances to succeed go down as the skill increases - shouldn't the comparison then get flipped, or something?

cizra (changes requested):
Could you double-check this?

code/code/misc/offense.cc
parkscs: The biggest thing that got changed was the missing "!" operator in front of ::num(0,99), which was causing the chances to go down as skills go up.  A more subtle change is to just reduce the overall chance of failure slightly by changing the constant value of 30 to 45.  But that's why the comparison operator did not change.

cizra: Ohh right. Thanks.

cizra (approved):

PR #391, opened 2022-05-01
Fix heap-use-after-free crash in casino valet proc

drewkht:
Crash was caused by the casinoElevatorGuard function calling delete on a
pointer to an object it didn't "own", rather than returning `DELETE_ITEM`
and allowing the object to be deleted further up the stack. The object
was being dereferenced in the `doGive` function in `inventory.cc` after
delete had already been called on it.


drewkht:
Going to close this out and resubmit it from my non-work Github account. Bleh.

PR #398, opened 2022-05-24
Fix logic to prevent crash when weapon in inv

drewhershey:
This proc had some bad logic that allowed it to attempt to trigger
whenever a weapon with the proc was in a character's inventory
when the character used the thief stab skill.

Because the proc was triggering when it shouldn't, it was passing
a garbage value to another call within that was causing a segfault
and crashing the game.

This commit also fixes that the proc was being executed twice
every time stab was used - once from the stab function itself, and
again from the normal check-proc-on-pulse.

### DIFF
--- a/code/code/spec/spec_objs_thief_quest.cc
+++ b/code/code/spec/spec_objs_thief_quest.cc
@@ -5,31 +5,30 @@
 #include "extern.h"
 #include "obj_base_weapon.h"
 
-int thiefQuestWeapon(TBeing *victim, cmdTypeT command, const char *arg, TObj *object, TObj *) {
-  wearSlotT limb = WEAR_NOWHERE;
-
-  if (command != CMD_STAB && command != CMD_BACKSTAB && command != CMD_SLIT)
+int thiefQuestWeapon(TBeing *victim, cmdTypeT command, const char *arg, TObj *object, TObj *thief) {
+  // Thief will only == nullptr when proc is called on the generic command proc
+  // check. Don't want proc to execute in that instance - should only happen in
+  // response to skill use. Otherwise it'll execute twice as often as it should.
+  if ((command != CMD_STAB && command != CMD_BACKSTAB && command != CMD_SLIT) || !thief || !victim || !object)
     return false;
+
+  auto *weapon = dynamic_cast<TBaseWeapon *>(object);
+  auto *wielder = weapon ? dynamic_cast<TBeing *>(weapon->equippedBy) : nullptr;
+
+  // Cast both thief and wielder to base class and check if both pointing to same
+  // memory location
+  if (!weapon || !wielder || !(dynamic_cast<TThing *>(thief) == dynamic_cast<TThing *>(wielder)))
+    return false;
+
+  wearSlotT limb = WEAR_NOWHERE;
 
   if (command == CMD_STAB)
     limb = static_cast<wearSlotT>(reinterpret_cast<uintptr_t>(arg));
 
-  if (command != CMD_STAB && strcmp(arg, "-special-"))
-    return false;
-
-  if (!victim || !object)
-    return false;
-
-  limb = command == CMD_BACKSTAB
-      ? WEAR_BACK
-      : command == CMD_SLIT
-      ? WEAR_NECK
-      : limb;
+  limb = command == CMD_BACKSTAB ? WEAR_BACK : command == CMD_SLIT ? WEAR_NECK : limb;
 
   if (limb == WEAR_NOWHERE || limb == HOLD_RIGHT || limb == HOLD_LEFT || limb == MAX_WEAR)
     return false;
-
-  sstring limbDescription = victim->describeBodySlot(limb);
 
   // 10% chance on stab, 50% on backstab/slit
   if (command == CMD_STAB) {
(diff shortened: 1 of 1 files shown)


drewhershey:
This PR will fix the recent crashes the game has been seeing when a character has the silver jambiya weapon in their inventory (which has the thief quest weapon proc on it) and then stabs while wielding a different weapon.

As a reminder about the ugly `reinterpret_cast` - this was discussed when this proc was merged in last year and we ended up deciding it was the best option to allow the proc to work without doing a large rewrite of the proc handling code.

It's necessary to allow passing a `wearFlagT` enum in through the `const char*` param.

parkscs (dismissed):
Seems like a straightforward tweak - would be good to get this fixed ASAP as it's caused a few crashes recently on prod.

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -5,73 +5,63 @@
...
+  // Thief will only == nullptr when proc is called on the generic command proc
+  // check. Don't want proc to execute in that instance - should only happen in
+  // response to skill use. Otherwise it'll execute twice as often as it should.
+  if ((command != CMD_STAB && command != CMD_BACKSTAB && command != CMD_SLIT) || !thief || !victim || !object)
     return false;
 
-  if (command == CMD_STAB)
-    limb = static_cast<wearSlotT>(reinterpret_cast<uintptr_t>(arg));
+  auto *weapon = dynamic_cast<TBaseWeapon *>(object);
+  auto *wielder = weapon ? dynamic_cast<TBeing *>(weapon->equippedBy) : nullptr;
```
cizra, line 16: I wonder if dynamic_casting a nullptr is UB?

code/code/spec/spec_objs_thief_quest.cc
```diff
@@ -5,73 +5,63 @@
...
 
-  if (command != CMD_STAB && strcmp(arg, "-special-"))
+  // Cast both thief and wielder to base class and check if both pointing to same
+  // memory location
```
cizra, line 19: Yeah, I see what you're doing - but it's absolutely _not_ clear why you would do this. Could we have a "why", not "what" here?

cizra (changes requested):

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Appears it's defined and acceptable:
>A null pointer value is converted to the null pointer value of the destination type by dynamic_cast.

https://docs.microsoft.com/en-us/cpp/cpp/dynamic-cast-operator

code/code/spec/spec_objs_thief_quest.cc
drewhershey: This is a little convoluted, so bear with me. I'll explain here then push a commit to update the comment as well.

As currently designed, every spec proc function (e.g. `thiefQuestWeapon` in this case) has to have an identical function signature because they all get stored in an array of structs of type `TObjSpecs`:

```c++
struct TObjSpecs {
  bool assignable;
  const char *name;
  int (*proc)(TBeing *, cmdTypeT, const char *, TObj *, TObj *);
};
```

This obviously isn't optimal design and it'd be great to find a better way to do it, but that would need to be its own project. In this PR I'm just working within the confines of the existing design.

The way object spec procs get executed is every time a pulse happens in the game loop or a command is executed by a character, `TObj::checkSpec()` is called on _every item_:

```c++
int TObj::checkSpec(TBeing *t, cmdTypeT cmd, const char *arg, TThing *t2)
{
  int rc;

  // we use static_cast here because we don't ALWAYS pass the proper kind of
  // parameter through the pointer fields.
  if (spec) {
    rc = (objSpecials[GET_OBJ_SPE_INDEX(spec)].proc) 
          (t, cmd, arg, this, static_cast<TObj *>(t2));
    return rc;
  }
  return FALSE;
}
```

So it uses the object's `TObj::spec` property to find the matching `TObjSpecs` object in the array then dynamically calls the function referenced in its `TObjSpecs::proc` property.

Therefore, if you want to write a spec proc which requires data in a form not expected in that function signature, you have to do two things:

- First, pass it in to the checkSpec function through an unused parameter - which often requires a `reinterpret_cast`. In this case, the `wearSlotT` enum (an `int`) representing which body part is being stabbed is needed for the proc, so it gets cast to a `char*` and passed through the otherwise unused `arg` param in `checkSpec`.

- THEN, it needs to be passed into the object's proc function as well, which can often (though not in this case) require _another_ cast before passing. It's a big mess.

However, the reason this specific proc was crashing the game was that I hadn't realized when I repurposed it previously (as it used to be the jambiya weapon proc) that object procs get called no matter where the item is located on the character. Just due to lacking a deep enough understanding of the proc code at the time I'd assumed it needed to be wielded for a weapon proc to get called.

So when a character had a weapon with this proc in their inventory then used the `stab` command with a different weapon, this proc was being executed but being passed some garbage values that were used to index an array further down the stack, causing a crash.

To fix this, the proc execution needed to be gated behind the weapon actually being _wielded_. To test for this I needed a reference to the `TBeing` that actually executed the stab command. From `cmd_stab.cc`, where the stab is actually executed and a call to `checkSpec` is made to see if any spec proc needs to activate on weapon stab, a `TBeing*` for the attacker was already being passed to `checkSpec` and then on to the `thiefQuestWeapon` function - but the original proc I repurposed here never used it, so the param wasn't named. I renamed that param to `thief` within the function, which gave me a reference to the person who sent the `stab` command.

Then I needed to confirm that the `TObj* object` being passed in was actually being **wielded** by the `TBeing` who sent the stab command. One way to do this would be to test if `thief->heldInPrimHand() == object`, which in retrospect is probably the better option as it ensures `object` is actually the one being stabbed with when `thief` is dual wielding.

In my case, I saw that the proc code was already referencing `object->equippedBy`, so I chose to test if `thief == wielder`, which would confirm that the person who sent the `stab` command also had `object` equipped. I did the `dynamic_cast<TThing*>` on them both because the compiler didn't like me comparing objects of two different types, and `TThing` is their common base class. 

(`thief` comes in having been previously implicitly cast from `TBeing*` to `TThing*` in the call to `checkSpec`, then explicitly cast to `TObj*` in the call to the spec proc function, so casting them both to `TThing*` seemed like the smartest option if I just wanted to compare memory location.)

I'm going to go ahead and change this to test via `heldInPrimHand` though, as it prevents the proc from executing if the weapon is in the offhand (as stab only uses the primary-hand weapon). So, good catch. 😁

code/code/spec/spec_objs_thief_quest.cc
cizra: In that case, you could write `dynamic_cast<TBeing*>(weapon && weapon->equippedBy)` - though please don't, it'd look abominable.

code/code/spec/spec_objs_thief_quest.cc
cizra: Pheew, long but clear. Thanks for the explanation!

code/code/spec/spec_objs_thief_quest.cc
drewhershey: This line was removed in https://github.com/sneezymud/sneezymud/pull/398/commits/950e22304888a94fcc672a6f3418adf574d64baa

code/code/spec/spec_objs_thief_quest.cc
drewhershey: Made changes as described in https://github.com/sneezymud/sneezymud/pull/398/commits/950e22304888a94fcc672a6f3418adf574d64baa

cizra (approved):

PR #403, opened 2022-06-23
Fix arch vampire and energy drain touch mob combat procs

drewhershey:
Due to popular demand, this PR fixes two mob combat special procs that have been in a broken state for a very long time. The proc code is almost entirely rewritten, but maintains the intended function/spirit of the original code.

There's a good chance this will make mobs with these procs much harder, but the overwhelming feeling on Discord was that it would be a positive change. The procs can easily be made less punishing with a few simple tweaks.

See news file for list of mobs affected by this fix.

### DIFF
--- a/code/code/spec/spec_mobs_combat.cc
+++ b/code/code/spec/spec_mobs_combat.cc
@@ -497,7 +497,7 @@ int arch_vampire(TBeing* vampire, cmdTypeT cmd, const char*, TMonster*, TObj*) {
   // energyDrain function, as that just introduces problems when non-mage mobs attempt it.
   // The bite is an innate ability, anyway.
   int dam = vampire->getSkillDam(victim, SPELL_ENERGY_DRAIN, vampire->GetMaxLevel(), 100);
-  int vit = dice(number(1, vampire->GetMaxLevel()), 4);
+  int vit = dice(::number(1, vampire->GetMaxLevel()), 4);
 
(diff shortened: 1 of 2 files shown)


code/code/spec/spec_mobs_combat.cc
```diff
@@ -10,27 +10,44 @@
...
 
-    return energyDrain(ch,ch->fight());
+  if (!dam || victim->getImmunity(IMMUNE_DRAIN) >= 100 ||
+      victim->isLucky(levelLuckModifier(level))) {
+    act("Nothing happens.", false, ch, nullptr, nullptr, TO_ROOM);
+  } else {
+    act("$N screams in agony as energy pours from $S body!", false, ch, nullptr, victim,
+      TO_NOTVICT);
+    act("You scream in agony as energy pours from your body!", false, ch, nullptr, victim, TO_VICT);
+    if (ch->reconcileDamage(victim, dam, SPELL_ENERGY_DRAIN) == -1) {
```
drewhershey, line 38: Just noticed this is a potential crash waiting to happen from a null dereference. Will fix this shortly.

drewhershey: Resolved in https://github.com/sneezymud/sneezymud/pull/403/commits/0b09ce281205e09697261c4b31b6055b6c35e057

cizra (approved):

cizra (approved):

PR #404, opened 2022-06-23
Bug fix to 2h spec - wasn't honing

parkscs:
Refactored this function to hopefully clean it up slightly, but the main reason for this PR is just to fix a single bug I introduced in the warrior changes code.  Any feedback is welcome, but this is a pretty minor change.  With that said, I'd like to get it in quickly if possible because right now the skill is impossible to hone for players (which is a problem).


drewhershey (approved):
Looks good to me. 👍

cizra (approved):

PR #405, opened 2022-07-12
add basic functional tests

cizra:
There are 3 of them:
- the first sets up a character in a new account,
- the 2nd (this is where the real-life test cases should go) performs an action and watches the output
- the last one cleans up by deleting the account.

Suggest to start reviewing from the test_002 - that's how the tests should actually look. The 1st looks very ugly with all those sleeps in it. Sneezy appears to clear some buffers, or something, without completely processing all data, thence the need for sleeps.

### DIFF
--- a/func-test/run_test.sh
+++ b/func-test/run_test.sh
@@ -1,0 +1,9 @@
+#!/bin/sh
+
+set -eu
+
+rm -f output/*
+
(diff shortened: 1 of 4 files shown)


drewhershey (approved):
Approved with two different accounts just because it's so good

PR #410, opened 2022-12-08
Use ubuntu-20.04 base image in Build and Test Github Action

drewhershey:
libmariadbclient-dev appears to have been deprecated in favor of libmariadb-dev

Should fix this error in our "Build and Test" Github Action:

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,7 +3,7 @@ on: [ pull_request ]
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     name: Build and Test
     steps:
     - name: Checkout this commit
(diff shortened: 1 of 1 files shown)


drewhershey:
Turned out that the original problem - libmariadbclient-dev package name being deprecated - is only an issue in Ubuntu 22.04. Switching the action to use the same image as our production image removes the need to make this change, so I reversed it - hence the force push. The only actual fix necessary was to use the ubuntu-20.04 base image.

I also went ahead and disabled the running of the cxxtest suite until we figure out why the tests aren't able to connect to the database when they run, as it makes the Build and Test action take over 30 minutes due to waiting for the db to time out repeatedly.

drewhershey:
Also rebased onto master, which is why it looks like there was an unrelated change in that second force push.

cizra (approved):

PR #427, opened 2022-12-20
Fix heap-use-after-free crash in `processAllInput()`

### LINKED ISSUE #426: Possible heap-use-after-free in connect.cc after player rents
```c++
2022|1219|21:48:36 :: Player I/O: Renting Chekov [1593011 talens/0 bank/699822238.22 xps/155 items/3 age-mod]
=================================================================
==1914==ERROR: AddressSanitizer: heap-use-after-free on address 0x629000231800 at pc 0x5595d52789a0 bp 0x7ffed83189b0 sp 0x7ffed83189a0
WRITE of size 8 at 0x629000231800 thread T0
    #0 0x5595d527899f in processAllInput() code/sys/connect.cc:2717
    #1 0x5595d53fd352 in TMainSocket::handleTimeAndSockets() code/sys/socket.cc:491
    #2 0x5595d540dfa6 in procHandleTimeAndSockets::run(TPulse const&) const code/sys/socket.cc:1664
    #3 0x5595d53ef66a in TScheduler::run(int) code/sys/process.cc:251
    #4 0x5595d5410139 in TMainSocket::gameLoop() code/sys/socket.cc:1843
    #5 0x5595d51d9f05 in run_the_game() code/sys/comm.cc:86
    #6 0x5595d51cd2fa in main code/main.cc:62
    #7 0x7f24b91f3082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)
    #8 0x5595d51cc70d in _start (/sneezymud-docker/sneezymud/code/sneezy+0x3cee70d)

0x629000231800 is located 5632 bytes inside of 16664-byte region [0x629000230200,0x629000234318)
freed by thread T0 here:
    #0 0x7f24b9a5ec65 in operator delete(void*, unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cc:177
    #1 0x5595d5d98db2 in TBeing::~TBeing() code/misc/structs.cc:316
    #2 0x5595d5b6bc6c in TPerson::~TPerson() code/misc/person.cc:85
    #3 0x5595d5b6bcfb in TPerson::~TPerson() code/misc/person.cc:120
    #4 0x5595d5278975 in processAllInput() code/sys/connect.cc:2716
    #5 0x5595d53fd352 in TMainSocket::handleTimeAndSockets() code/sys/socket.cc:491
    #6 0x5595d540dfa6 in procHandleTimeAndSockets::run(TPulse const&) const code/sys/socket.cc:1664
    #7 0x5595d53ef66a in TScheduler::run(int) code/sys/process.cc:251
    #8 0x5595d5410139 in TMainSocket::gameLoop() code/sys/socket.cc:1843
    #9 0x5595d51d9f05 in run_the_game() code/sys/comm.cc:86
    #10 0x5595d51cd2fa in main code/main.cc:62
    #11 0x7f24b91f3082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)

previously allocated by thread T0 here:
    #0 0x7f24b9a5d587 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cc:104
    #1 0x5595d54113f3 in TMainSocket::newDescriptor(int) code/sys/socket.cc:1914
    #2 0x5595d53fcd17 in TMainSocket::handleTimeAndSockets() code/sys/socket.cc:454
    #3 0x5595d540dfa6 in procHandleTimeAndSockets::run(TPulse const&) const code/sys/socket.cc:1664
    #4 0x5595d53ef66a in TScheduler::run(int) code/sys/process.cc:251
    #5 0x5595d5410139 in TMainSocket::gameLoop() code/sys/socket.cc:1843
    #6 0x5595d51d9f05 in run_the_game() code/sys/comm.cc:86
    #7 0x5595d51cd2fa in main code/main.cc:62
    #8 0x7f24b91f3082 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x24082)

SUMMARY: AddressSanitizer: heap-use-after-free code/sys/connect.cc:2717 in processAllInput()
Shadow bytes around the buggy address:
  0x0c528003e2b0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e2c0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e2d0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e2e0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e2f0: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
=>0x0c528003e300:[fd]fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e310: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e320: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e330: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e340: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x0c528003e350: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==1914==ABORTING
/sneezymud-docker/sneezymud/code/objs/tests/DescriptorListTest
```

Guilty code:
```c++
      if (is_client_sstring(comm)) {
        rc = d->read_client(comm);
        if (IS_SET_DELETE(rc, DELETE_THIS)) {
          delete d;
          d = NULL;
          continue;
        } 
        if (IS_SET_DELETE(rc, DELETE_VICT)) {
          delete d->character;
          d->character = NULL;
          continue;
        }
      }
```

drewhershey:
Should fix a heap-use-after-free that's caused some crashes recently. Must be happening when `d` somehow is equal to `d->character->desc` when a player rents out of the game but haven't been able to reproduce it myself.

Fixes #426 (see that issue for stack trace)

### DIFF
--- a/code/code/sys/connect.cc
+++ b/code/code/sys/connect.cc
@@ -2713,8 +2713,19 @@ void processAllInput()
           continue;
         } 
         if (IS_SET_DELETE(rc, DELETE_VICT)) {
-          delete d->character;
-          d->character = NULL;
+          if (!d->character) continue;
+          
+          if (d == d->character->desc) {
+            vlogf(LOG_BUG, format("d == d->character->desc inside "
+                                  "processAllInput() for %s (Account: %s)") %
+                             d->character->getName() %
+                             (d->account ? d->account->name : "unknown"));
+            delete d->character;
+            d = nullptr;
+          } else {
+            delete d->character;
+            d->character = nullptr;
+          }
           continue;
         }
       } else if (d->str) 


code/code/sys/connect.cc
```diff
@@ -2713,8 +2713,19 @@ void processAllInput()
           continue;
         } 
         if (IS_SET_DELETE(rc, DELETE_VICT)) {
-          delete d->character;
-          d->character = NULL;
+          if (!d->character) continue;
```
cizra, line 2716: Add a warning to log, perhaps?

code/code/sys/connect.cc
```diff
@@ -2713,8 +2713,19 @@ void processAllInput()
...
+          
+          if (d == d->character->desc) {
+            vlogf(LOG_BUG, format("d == d->character->desc inside "
+                                  "processAllInput() for %s (Account: %s)") %
+                             d->character->getName() %
+                             (d->account ? d->account->name : "unknown"));
+            delete d->character;
+            d = nullptr;
+          } else {
+            delete d->character;
```
cizra, line 2726:
```
if (condition) {
  do a
  do b
} else {
  do b
}
```
looks much, much better like:
```
if (condition)
  do a
do b
```

cizra (changes requested):

code/code/sys/connect.cc
drewhershey: I thought about that but wasn't sure how to accomplish it in this case. I need to compare `d` and `d->character->desc` to know whether to set `d` itself to null or only `d->character` to null.

If I factor out the common `delete` call, doing it first, and it turns out `d->character->desc` _was_ pointing to the same desc as `d` then my comparison would cause the same crash I'm trying to fix.

Am I just missing an obvious way around this?

code/code/sys/connect.cc
drewhershey: Are you meaning a warning for when `rc` ends up with the DELETE_VICT flag here but `d->character` is already null somehow?

cizra: Yes. Is this supposed to happen? Probably not?

code/code/sys/connect.cc
cizra: Uh, I'm probably missing something - why don't you delete it _afterwards_, unconditionally? You're doing it in both branches already.

drewhershey: If `d != d->character->desc` I only want to set `d->character` to null, as `d` should continue to exist but needs to have the pointer to some soon-to-be deleted other object contained in its character property cleared out. 

Whereas if they're equivalent, meaning `d->character` is actually pointing to `d`'s own parent object, I want to set `d` itself to null because the desc `d` is pointing to will be deleted as well as soon as `d->character` is deleted.

I can't call delete until after I've tested for equivalency, as in one case deleting `d->character` will actually delete `d` as well, so how do I factor the delete out while still knowing which needs to be set to null afterwards?

Very possible I'm just confused and/or not aware of something obvious here, though.

cizra: Aaaaah, so that's what i was missing :D Sorry for the noise.

cizra (approved):

code/code/sys/connect.cc
drewhershey: All good, thanks for sanity checking me. 😁 It's important for code like this.

PR #434, opened 2023-01-14
Run clang-format on all .h and .cc files in code/code/**

drewhershey:
This PR sets the repo up to have a nicely formatted codebase using easily configurable rules.

- Runs clang-format on all .h and .cc files in code/code/** using the rules in `.clang-format` file in root directory.
- Adds `.git-blame-ignore-revs` file and adds the format commit to it to avoid polluting `git blame` with the format commit. Each developer needs to update the git config in their local repo to make use of this. Instructions are in the config file.
- Adds a `pre-commit-config.yaml` file to include support for `pre-commit` library, which makes it easy to run clang-format on files included in a commit as a pre-commit git hook. Each developer will still need to install it manually in their local repo, however (which is extremely simple). Instructions for use are in the config file.
- Adds an `.editorconfig` file to help with adding proper file endings and removing extraneous whitespace 
- Adds a bash script that can be used, if desired, to manually run clang-format on the .cc/.h files. This requires having clang-format 14 or higher installed in the user's local environment, however, while the `pre-commit` tool doesn't.

Take a glance through the formatting changes and let me know if anything looks weird or you'd prefer it formatted differently. The rules can be configured to handle *almost* any desired formatting style.

I intentionally set clang-format to not modify header include order, as that will cause a ton of compiler errors. I'll work on fixing/sorting includes using a different tool in a separate PR.


.git-blame-ignore-revs
```diff
@@ -1,2 +1,5 @@
 # To enable the use of this file in your local repo, type the following at the CLI:
 # git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Run clang-format on all .cc and .h files in code/code/**
+1c508231f77464bf5807fba89ca72ab8c445ffc1
```
cizra, line 5: Depends on how we merge, this hash might change and need updating. But let's see how it goes.

run_clang_format.sh
```diff
@@ -0,0 +1,6 @@
+#! /bin/bash
+#
+# Finds all .cc and .h files in code/code/** and runs clang-format on them, editing them in-place
+# using the rules in the .clang-format file in the root folder
+
+find code/code/ -iname '*.h' -o -iname '*.cc' | xargs clang-format -i
```
cizra, line 6: Not really crucial, but fun for the future:
filenames can contain any old junk, including terminal bells, newlines, nonprintables, invalid UTF-8-like sequences, etc (only the null terminator and `/` are not allowed). A newline in filename would break this command.
You can work around this issue by using null characters as the separators between files:
`find ... -print0 | xargs -0 ...`

code/code/sys/database.cc
```diff
@@ -9,347 +9,316 @@
...
+std::vector<std::string> db_passwords(DB_MAX);
+
+IDatabase::~IDatabase() {}
+
+const char* TDatabaseConnection::getConnectParam(dbTypeT type) {
   assert(type >= 0);
   assert(type < DB_MAX);
-  if (db_names[type] != "")
-    return db_names[type].c_str();
+  if (db_names[type] != "") return db_names[type].c_str();
```
cizra, line 61: Was it a considered decision to pull these on one line? I'm staring at it, at the moment, and it feels less readable than the 2-line version.

cizra (dismissed):

code/code/sys/database.cc
drewhershey: That's something that's handled by a specific clang-format rule and is easily changed. I've gone back and forth on it myself, but remembered you saying you liked preserving vertical space, so I went with it. However, I have zero qualms about changing it as I don't have any strong feelings about it either way.

Were there any other stylistic things you didn't care for? Clang format has a ton of rules so most things can be changed and I'm pretty familiar with them at this point.

run_clang_format.sh
drewhershey: I might as well add this to the script, though hopefully people will use the `pre-commit` tool instead, which accepts specific filetypes as a configuration variable and shouldn't be susceptible to the issue you pointed out with my script. That tool was what I used to format all the files and just included the script as a "might as well".

.git-blame-ignore-revs
drewhershey: Hm, yeah, hadn't thought about that. If it does, I'd probably have to modify the SHA in this file then merge that separately using the "merge" option instead "rebase", I assume?

cizra: I vote not to worry about it today. Get the PR in, fix the hash in the next PR, if needed.

code/code/sys/database.cc
cizra: There are ifs around that have so many parentheses and conditions and stuff that they don't really accommodate the next statement on the same line. I'd vote to keep them on the next line, unless they're trivial. Tho I'm afraid it's hard to teach clang-format what "trivial" means, so let's default to 2 lines.

drewhershey: This rule is specifically configured to only put if statements that have no else clause on a single line. Any that do have an else clause, no matter how short, get put on separate lines. Knowing this, do you still have a preference? It doesn't make a difference to me, so whichever you want is fine.

drewhershey:
Just updated the bash script, removed the SHA from the .git-blame-ignore-revs file so I can just add it after merging this PR, and updated the formatting to always put the if body on a separate line then force pushed.

cizra (dismissed):

cizra (approved):

Issue #431, opened 2023-01-01
stack-buffer-overflow crash in TBeingSave (create_mobs.cc)


drewhershey:
```c++
2022-12-31T15:21:40.400771000Z 2022|1231|15:21:40 :: Vasco (12687):medit resave du
2022-12-31T15:21:40.401489000Z =================================================================
2022-12-31T15:21:40.402020000Z ==7==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x7ffcdd1e8950 at pc 0x558e2d8665e4 bp 0x7ffcdd1e7fa0 sp 0x7ffcdd1e7f90
2022-12-31T15:21:40.402218000Z WRITE of size 1 at 0x7ffcdd1e8950 thread T0
2022-12-31T15:21:41.177706000Z     #0 0x558e2d8665e3 in TBeingSave code/misc/create_mobs.cc:342
2022-12-31T15:21:41.178630000Z     #1 0x558e2d867078 in msave code/misc/create_mobs.cc:428
2022-12-31T15:21:41.179061000Z     #2 0x558e2d86bfb5 in TPerson::doMedit(char const*) code/misc/create_mobs.cc:2082
2022-12-31T15:21:41.271875000Z     #3 0x558e2e1360d9 in TBeing::doCommand(cmdTypeT, sstring const&, TThing*, bool) code/misc/parse.cc:1403
2022-12-31T15:21:41.272572000Z     #4 0x558e2e139877 in TBeing::parseCommand(sstring const&, bool, bool) code/misc/parse.cc:2086
2022-12-31T15:21:41.413791000Z     #5 0x558e2d43da9d in processAllInput() code/sys/connect.cc:2750
2022-12-31T15:21:41.465988000Z     #6 0x558e2d626064 in TMainSocket::handleTimeAndSockets() code/sys/socket.cc:491
2022-12-31T15:21:41.466674000Z     #7 0x558e2d626583 in procHandleTimeAndSockets::run(TPulse const&) const code/sys/socket.cc:1664
2022-12-31T15:21:41.488655000Z     #8 0x558e2d5ff67d in TScheduler::run(int) code/sys/process.cc:251
2022-12-31T15:21:41.488979000Z     #9 0x558e2d631524 in TMainSocket::gameLoop() code/sys/socket.cc:1843
2022-12-31T15:21:41.571258000Z     #10 0x558e2d32f160 in run_the_game() code/sys/comm.cc:86
[...]

PR #435, opened 2023-03-15
Medit string buffers

sntwo:
Addresses #431 

* Prevents Descriptor string from growing over max_str limit (if set)
* Adjusts some string length limits in Medit, moves CRLF process upstream to string creation


cizra (approved):

drewhershey:
Sorry this took so long - I thought for some reason it had gotten merged back when it was approved and didn't notice until I just randomly thought to see what the open PRs were.

PR #446, opened 2023-11-13
Format all .cc/.h files in code/code/** via clang-format

drewhershey:
Re-do of the previous full codebase format PR, now that test is merged back into master.


wershlak (approved):
Thanks!

PR #448, opened 2023-11-13
Add code to support two new quests written by Vasco

drewhershey:
- Quest starting at the first warrior GM with a low level weapon reward
- Thief-specific class quest available at level 5, starting at first thief GM
- Includes help files for the "quests" command in-game

The actual mobresponse entries for these quests have to be manually inserted into the database (with all \n characters converted to the actual string "\n"), as there's no in-game mechanism for adding or modifying mobresponses currently. However, I'm also saving them to the web/quests folder just for posterity.


wershlak:
These responses should be in https://github.com/sneezymud/sneezymud/blob/master/_Setup-data/sql_data/sneezy/mobresponses.sql

Have these responses been added to the production DB? We should be able to dump the DB and get those files to commit here as well. 

The developer experience for quests sucks. We should brainstorm some way to break out quests such that developers can add new ones via a commit like normal code. Perhaps either convert them back to one text file per quest that is loaded at boot or loaded into the DB at compile time.

cizra:
> The developer experience for quests sucks. We should brainstorm some way to break out quests such that developers can add new ones via a commit like normal code. Perhaps either convert them back to one text file per quest that is loaded at boot or loaded into the DB at compile time.

The whole reason for having them in DB is that it allows online editing. Online editing is not very compatible with storing stuff in Git. Which option do we want, long-term? If quests should live in Git, does this mean that mobs should also live in Git, lest they get out of sync with quests (imagine a vnum changes, and now a lady-of-the-evening garrottes you, instead of asphyxiating)?

We can combine the advantages of data-at-rest in files with advantages of a DB at runtime by creating a transient SQLite database in memory, filling it on boot, then using that for complex queries, if we want.

Oh, and as for online editing, we have some rudimentary something for mobresponses at https://github.com/sneezymud/sneezymud-docker/blob/master/web/controller.py#L83 and https://github.com/sneezymud/sneezymud-docker/blob/master/web/templates/mobresponse.html

drewhershey:
> lest they get out of sync with quests (imagine a vnum changes, and now a lady-of-the-evening garrottes you, instead of asphyxiating)?

First of all, this sounds awesome.

> Which option do we want, long-term? If quests should live in Git, does this mean that mobs should also live in Git, lest they get out of sync with quests

We've discussed this a bit previously and I've been doing some thinking about it myself lately.

The main thing to me is that the game is open source, so it should be easy for anyone who wants to to spin up a new instance of it. It should also be easy for anyone who wants to to contribute to the game, whether by submitting code changes *or* builder-related changes like new items/mobs/etc.

With that in mind, the setup I've come to favor the most, personally, is:
1) Keep in the database all data that's related to a *specific instance* of the game - e.g. stuff we wouldn't want to store in the repo because it's something a new instance of the game being spun up would most likely overwrite.
     - Just quickly glancing at the current table structure, this would include things like player/account files, rent files, ban lists, holiday quest objective tracking, global and individual toggle state, pets, etc.
     - There are some tables that - though they fit the previous criteria - might make sense to keep copies of in the database just in case they were wanted for a new instance. Things like the old player owned shops and houses, or anything else that's likely to be somehow nostalgic to returning players.
     - Currently there's a lot of game data a new instance *would* want (mob, object, and room data for instance) that's stored only in the database. This forces us to keep SQL dumps in the repo as part of the codebase, and to update them somehow.
     - Some of the stuff currently being stored as flat files would need to be converted to be stored in the database instead, like pfiles
 2) Store as flat files/data-at-rest in some human-friendly format (JSON, YAML, XML, etc) all game data that's primarily used as a reference and loaded into memory when the game first boots. Nest related data when possible, instead of keeping it normalized out.
    - This would be any data that would definitely be wanted by a new instance of the game; mobs, items, rooms, shops, etc.
    - Data that fits in this category is rarely or almost never modified via non-immortal characters simply playing the game. It would almost always only be modified by a builder using in-game tools like oedit.
    - If these files were stored in the repo, as data-at-rest - in a format easy to read, understand, and edit by humans - changes to them could be submitted and approved via PRs, and we would have a running history of changes to the data through git.
    - This would make it possible for anyone to contribute to the game as a builder in the same way one contributes code - edit or create a file following appropriate syntax and formatting guidelines, and submit a PR for approval and merging.
    - By storing the game data in this format, we could keep related data nested inside its relevant parent data, solving some issues such as the one Cizra brought up with a mobresponse becoming detached from the mob it's supposed to belong to due to a VNum change. Nesting a mob's responses as a simple property inside the rest of the mob's data solves this problem.
    - We'd have to discuss if it makes sense to have individual files for each mob/item/room/etc or if we'd want to group data in some way to reduce the number of files.

As far as this specific PR goes, I'll update the mobresponses.sql file in _Setup_data/sql_tables/sneezy to include these changes.

Thank you for coming to my TED talk.

drewhershey:
Force pushed because there was an extra apostrophe in one line. 🙄

wershlak:
> The whole reason for having them in DB is that it allows online editing. Online editing is not very compatible with storing stuff in Git. Which option do we want, long-term? If quests should live in Git, does this mean that mobs should also live in Git, lest they get out of sync with quests (imagine a vnum changes, and now a lady-of-the-evening garrottes you, instead of asphyxiating)?
> 

I assumed the reason we have things in a DB was for performance or something? I know when folks made changes to items it required a reboot so I don't think speed of getting changes in is a concern. 

I'm not really tied to any specific philosophy here. Only trying to brainstorm how to reduce friction in getting people to contribute to the mud. I think editing a text file of some sort is a lot easier than dealing with the set of tools we have for object editing.

> 
> Oh, and as for online editing, we have some rudimentary something for mobresponses at https://github.com/sneezymud/sneezymud-docker/blob/master/web/controller.py#L83 and https://github.com/sneezymud/sneezymud-docker/blob/master/web/templates/mobresponse.html

Didn't realize that was there. That's pretty awesome!

wershlak:
> 
> As far as this specific PR goes, I'll update the mobresponses.sql file in _Setup_data/sql_tables/sneezy to include these changes.
> 
> Thank you for coming to my TED talk.

Thanks for getting this updated and I appreciate the ideas here

wershlak (approved):
Thanks for updating this

PR #449, opened 2023-11-13
Fix a ton of small bugs in various zonefiles and remove extraneous whitespace

drewhershey:
In the process of learning more about how zonefiles work a while back, I used regex to search the entire zonefiles folder at once for various little bugs/errors and fix them. Things like mismatches between a ? command and the following command type.
Overall won't change anything drastically but will fix some loads not occurring, or occurring too often, etc. And just makes things
"correct", which is always nice.

Also used EditorConfig to remove all extraneous whitespace from the zonefiles, of which there was a ton.


wershlak (approved):
Great thanks!

PR #450, opened 2023-11-22
move weapon check to only where needed

wershlak:
Moving weapon level check to inside the where we strictly expect to need a weapon CMD_OBJ_HIT. This allows the proc to be used on other objects without error.


drewhershey (approved):
Looks fine to me - I assume this is to allow the CMD_GENERIC_PULSE stuff to happen on other objects without having to write an entirely separate proc for just those?

PR #453, opened 2023-12-13
update zonefiles per Vasco

wershlak:
Update these zone files based on vasco's changes.


drewhershey:
I made a few very minor fixes - mostly just making sure Y commands come after any E commands for the same mob. Otherwise it all looks good to me.

drewhershey (approved):

PR #456, opened 2023-12-14
Add a GitHub action to build and push images for the test branch, to streamline test server operation

drewhershey:
Functions very similarly to the action for the main branch. This way we can push to test, have it automatically build and push the image to the existing sneezymud/sneezymud Docker Hub repo under the 'test' tag, and do `docker pull sneezymud/sneezymud:test` -> `docker-compose up -d` on the test server instead of having to manually build the image on there.

Will require very minor tweaks to the docker-compose.yml file on the test server to use the new image name, which won't even need to be committed.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -1,0 +1,33 @@
+name: Build and Publish - Test Server
+on:
+  push:
+    branches:
+      - "test"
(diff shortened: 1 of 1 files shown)


drewhershey:
Closing this out as we're no longer going to be running a test server

PR #458, opened 2023-12-17
Fix low mvroom command

drewhershey:
The "zone" field was left out of the query which was causing the inserts to the room table to fail.

```
2023|1217|00:44:35 :: Database: query failed: Field 'zone' doesn't have a default value
2023|1217|00:44:35 :: Database: insert into room (vnum,x,y,z,name,description,room_flag,sector,teletime,teletarg,telelook,river_speed,river_dir,capacity,height,spec) values (35722,-68,14,-2, 'Large Room in the Burrow','   This large room is somewhat cleaner then the other tunnels and warrens in\nthe burrow.  There is another large room to the west.  More rooms lie to the\nnorth, northeast, and east.  A curtain hangs over a large hole in the south\nwall.  The room itself is decorated with shiney objects and colorful, rotting\ntapestries.  A slightly rank smell drifts in from the west.\n',131080,33,0,0,0,0,-1,0,100,0)
```

Tested working in my dev build using the latest backup from the live server. Was able to run `low mvroom vasco 2 35700-35799` and confirm there was no errors in the log this time, and the room entries now show up in the database as expected.

### DIFF
--- a/code/code/cmd/cmd_low.cc
+++ b/code/code/cmd/cmd_low.cc
@@ -732,7 +732,7 @@ void TMonster::checkMobStats(tinyfileTypeT forReal) {
   }
 #if 0
   if (strlen(getName()) > MAX_NAME_LENGTH-1) {
-    vlogf(LOG_LOW, format("%s (%d) had excessive mob name length.") % 
+    vlogf(LOG_LOW, format("%s (%d) had excessive mob name length.") %
          getName() % mobVnum());
   }
 #endif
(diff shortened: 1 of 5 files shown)


drewhershey:
Added a map recalc update to this PR as it's closely related to mvroom. It makes recalc assignable as an actual imm power (either `@set wizpower <person> 114` or as part of the `god` package of imm powers) instead of only usable by level 60s.

wershlak (approved):
Thanks for tracking this down.

PR #460, opened 2023-12-20
Fix mistake in final line of updated Dockerfile

drewhershey:
The final line of the updated Dockerfile has a mistake, which this PR fixes.

RUN executes the command during the build of the image, which isn't what we want. CMD executes the command on container startup, which is correct. For some reason RUN still worked when testing in my local Docker setup, but causes a build failure
when done by the Github Action that builds/pushes a new image on commit.


parkscs (approved):
Take 2!

drewhershey:
Force push was because I rebased the branch onto master to avoid a merge commit being included. Nothing actually changed, code-wise.

PR #462, opened 2023-12-22
Make TDatabase own the pimpl by reference, not pointer

cizra:


code/code/sys/database.h
```diff
@@ -126,7 +126,7 @@ class TDatabase : public IDatabase {
     virtual ~TDatabase();
 
   private:
-    TDatabasePimpl* pimpl;
+    TDatabasePimpl& pimpl;
```
drewhershey, line 129: Purely out of curiosity, why make it a reference with manual heap allocation/deallocation rather than a unique_ptr?

drewhershey (approved):

code/code/sys/database.h
cizra: To avoid having `#include <memory>` in the header.
We don't need all the fancy move semantics of unique_ptrs. A reference ought to work just as nicely as a pointer, yet look nicer.

drewhershey: Makes sense, thanks.

drewhershey (approved):

PR #463, opened 2023-12-22
Optimize boot: TBeing::initSkillsBasedOnDiscLearning

cizra:


code/code/misc/spell_info.cc
```diff
@@ -3786,4 +3759,9 @@ void buildSpellArray() {
           spell->name % spell->start % spell->learn);
     }
   }
+
+  for (auto spellInfo : discArray) {
+    if (spellInfo != nullptr)
+      spellsOfDiscipline[spellInfo->disc].push_back(spellInfo->spellNum);
```
drewhershey, line 3765: I get the following ASan crash when I try to boot the game up on this branch:
```
2024|0111|09:40:09 :: Misc: Initializing Spells.
=================================================================
==260132==ERROR: AddressSanitizer: global-buffer-overflow on address 0x555560aad590 at pc 0x55555982359f bp 0x7fffffff8b40 sp 0x7fffffff8b30
READ of size 8 at 0x555560aad590 thread T0
    #0 0x55555982359e in std::vector<spellNumT, std::allocator<spellNumT> >::push_back(spellNumT const&) /usr/include/c++/9/bits/stl_vector.h:1186
    #1 0x555559c60e08 in buildSpellArray() code/misc/spell_info.cc:3765
    #2 0x5555592217c3 in bootDb() code/sys/db.cc:429
    #3 0x55555910d401 in run_the_game() code/sys/comm.cc:77
    #4 0x5555591015ac in main code/main.cc:64
    #5 0x7ffff6e54d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f)
    #6 0x7ffff6e54e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f)
    #7 0x555559100764 in _start (/home/drew/source/repos/sneezymud/code/sneezy+0x3bac764)

0x555560aad590 is located 16 bytes to the right of global variable 'discArray' defined in 'code/misc/spell_info.cc:18:12' (0x555560aac700) of size 3712
0x555560aad590 is located 16 bytes to the left of global variable 'spellsOfDiscipline' defined in 'code/misc/spell_info.cc:19:23' (0x555560aad5a0) of size 1656
SUMMARY: AddressSanitizer: global-buffer-overflow /usr/include/c++/9/bits/stl_vector.h:1186 in std::vector<spellNumT, std::allocator<spellNumT> >::push_back(spellNumT const&)
```

drewhershey (changes requested):

code/code/misc/spell_info.cc
drewhershey: I compiled with G++-9 and optimization level set to 0. Going to compile with olevel=2 and see if it makes any difference.

drewhershey: So strangely I don't get the crash when compiling with optimizations at level 2, only when compiling with optimizations turned off. So maybe this isn't actually a problem? What do you think?

cizra: It totally is... but I don't know what might be up, gotta figure it out

wershlak:
@cizra Is this still needed? Do you want to pick it up again or should i close it?

PR #466, opened 2024-01-05
Update/fix/add some zonefiles

drewhershey:
- Fixes some small errors in Prison (though the I commands still won't work correctly due to a code bug, but the command will now be correct once the code is fixed)
- Adds Sidartha's Avalon zonefile (features new tailor/smith mobs)
- Adds Sidartha's North Valley zonefile
- Includes Vasco's updates to Drow City and Haon'Dor Hunting Grounds


wershlak (approved):
Seems ok

PR #468, opened 2024-01-10
Fix null dereference in runResetCmdE, add null check to TBeing::canUseEquipment

drewhershey:
Obj was getting passed in to mob->canUseEquipment as nullptr due to a bug in runResetCmdE, and that function doesn't test for null before dereferencing.

Tested and confirmed this fixes the crash seen on the live server.

### DIFF
--- a/code/code/misc/equip.cc
+++ b/code/code/misc/equip.cc
@@ -119,6 +119,12 @@ bool TBeing::isWieldingWeapon() {
 
 bool TBeing::canUseEquipment(const TObj* o, silentTypeT silent,
   wearKeyT key) const {
+  if (!o) {
+    vlogf(LOG_BUG,
+      format("canUseEquipment: null object passed (being: %s)") % getName());
+    return false;
+  }
+
   bool held = false;
 
   if (key == WEAR_KEY_HOLD || key == WEAR_KEY_HOLD_R || key == WEAR_KEY_HOLD_L)
--- a/code/code/sys/db.cc
+++ b/code/code/sys/db.cc
@@ -2889,12 +2889,12 @@ void runResetCmdE(zoneData& zd, resetCom& rs, resetFlag flags, bool&,
     obj = !IS_SET(flags, resetFlagPropLoad)
             ? read_object_buy_build(mob, rs.arg1, REAL)
             : read_object(rs.arg1, REAL);
+  }
 
-    if (!obj) {
-      repoCheck(mob, rs.arg1);
-      last_cmd = objload = false;
-      return;
-    }
+  if (!obj) {
+    repoCheck(mob, rs.arg1);
+    last_cmd = objload = false;
+    return;
   }
 
   // these are just safety logs, equipping will be done regardless
--- a/lib/txt/news
+++ b/lib/txt/news
@@ -1,3 +1,5 @@
+01-10-24 : Fixed a couple crash bugs
+
 08-10-22 : Spells/rituals/prayers commands updated. All three commands now do the same
            thing, and should make seeing all spells/prayers/rituals available to a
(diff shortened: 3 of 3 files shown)


code/code/misc/equip.cc
```diff
@@ -119,6 +119,12 @@ bool TBeing::isWieldingWeapon() {
 
 bool TBeing::canUseEquipment(const TObj* o, silentTypeT silent,
   wearKeyT key) const {
+  if (!o) {
```
wershlak, line 122: Makes sense to me

code/code/sys/db.cc
```diff
@@ -2889,12 +2889,12 @@ void runResetCmdE(zoneData& zd, resetCom& rs, resetFlag flags, bool&,
...
             ? read_object_buy_build(mob, rs.arg1, REAL)
             : read_object(rs.arg1, REAL);
+  }
 
-    if (!obj) {
-      repoCheck(mob, rs.arg1);
-      last_cmd = objload = false;
-      return;
-    }
+  if (!obj) {
```
wershlak, line 2894: Can you explain moving this? Not sure I understand what repoCheck is for or why we would do it if !obj here

drewhershey: So at this point in the code, an object that's loaded via the `E` command in a zonefile has passed its load chance and is attempting to load either to be equipped on the mob (if load on death is off) or for the player to loot after the mob's death. However it only loads the object if the item hasn't already hit its max exist value in game.

repoCheck is left over from the days when repo was turned on. The reason it's called there is because at that point we've determined that the object *should* have loaded but was prevented due to max exist already being hit, which is exactly what repo was intended to address.

The repoCheck function is and has been short-circuited now for a while to just do nothing, but I left it there for posterity in case we ever for some reason wanted to revisit it, since that's where we'd want to check for repo.

The `if (!obj) { ...` block being within that previous `if` statement body was causing a crash because the code beyond that point expects `obj` to not be null. I accidentally created this bug when I fixed the load rates recently.

The intended and correct behavior is:
1) `TObj* obj` is null as we prepare to copy an item from the cache
2) If the item to be loaded hasn't reached its max_exist, create a copy of the item from the cache and store a pointer to the new item in `obj`
3) Check if `obj` is still null. If it is, we know the item was at max_exist and couldn't be loaded (or there was a problem copying from the cache). Therefore, immediately return as there's nothing left to do (besides check for repo, if we're using that)
4) If `obj` isn't null it means the item wasn't at max_exist and we can continue loading it

How it was working previously, causing the crash:
1) Same
2) Same
3) If the item to be loaded hasn't reached its max_exist **and for some reason copying the new item from the cache failed (meaning `obj` is still null) immediately return**
4) **If the item to be loaded HAS reached its max_exist, skip over the contents of that `if` statement and continue trying to load the item - which is guaranteed to still be null because the code that actually copies the item from the cache only happens if it wasn't at max exist**

So the behavior of steps 3 and 4 previously were wrong due to not testing `obj` for null outside of that `if` statement. Moving the test below as I did fixes that behavior. I added the null check to TBeing::canUseEquipment function as well just because I figured it might be called from elsewhere in the code with a null object, so it might as well check for null itself too. But that was just a "because I was there" addition.

code/code/misc/equip.cc
drewhershey: Technically this function should be receiving a reference to a TObj instead of a pointer, because the function won't work if `o` is null and references can't be null. But fixing that would be a bigger project and out of scope of this PR.

code/code/sys/db.cc
wershlak: We chatted offline and I now understand the change.

wershlak (approved):

PR #470, opened 2024-01-12
Integration and configuration for Discord Webhooks

grumpygabe:
Adds some simple discord webhook integration to SneezyMUD.  Use discord.cfg to assign URLs for certain events to be sent to a discord webhook.

### DIFF
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -91,13 +91,16 @@ opt_cxx =            get_option('CXX',         'g++')
 
 
 ## ensure submodules are ready to go
-submodules = subprocess.check_output(['git', 'submodule', 'status']).decode('utf-8').split()
-if any(sub.startswith('-') for sub in submodules):
-    info('Initializing submodule dependencies, please wait...')
-    try:
-        subprocess.check_call(['git', 'submodule', 'update', '--init'])
-    except submodule.CalledProcessError:
-        error_exit('Submodule init failed, is your Internet connection working?')
+try: 
+    submodules = subprocess.check_output(['git', 'submodule', 'status']).decode('utf-8').split()
+    if any(sub.startswith('-') for sub in submodules):
+        info('Initializing submodule dependencies, please wait...')
+        try:
+            subprocess.check_call(['git', 'submodule', 'update', '--init'])
+        except submodule.CalledProcessError:
+            error_exit('Submodule init failed, is your Internet connection working?')
+except Exception:
+    warning('Submodules could not be listed')
 
 
 ## detect compiler and version
@@ -236,6 +239,8 @@ if flag_twitter:
     env.Append(CXXFLAGS='-DTWITTER')
     env.Append(extra_libs=['curl'])
 
+## we want to use this for discord webhooks too, so include it
+env.Append(extra_libs=['curl'])
 
 ## universal flags
 env.Append(LINKFLAGS='-g')
@@ -276,12 +281,17 @@ if flag_pretty:
     Progress([ "\033[32mChecking targets\033[0m %s\r" % s for s in '-\\|/' ],
              interval=5)
 
-git_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
+try:
+    git_hash = subprocess.check_output(['git', 'rev-parse', 'HEAD']).decode('utf-8').strip()
+except Exception:
+    warning("git_hash could not be read.")
+    git_hash = "?"
 
 try:
     git_dirty = "+" if subprocess.check_output(['git', 'status', '--short']).decode('uft-8') else ""
 except Exception:
-    # Nobody ever figured out why it fails on some systems
+    # Nobody ever figured out why it fails on some systems  
+    warning("git status could not be read.")
     git_dirty = "?"
 
 with open('code/version.h', 'w') as f:
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -237,9 +237,9 @@ if flag_shared:
 ## twitter: use curl to post shouts
 if flag_twitter:
     env.Append(CXXFLAGS='-DTWITTER')
-    env.Append(extra_libs=['curl'])
+    #env.Append(extra_libs=['curl'])
 
-## we want to use this for discord webhooks too, so include it
+## discord: use curl for discord webhooks as well, so always include
 env.Append(extra_libs=['curl'])
 
 ## universal flags
--- a/code/SConstruct
+++ b/code/SConstruct
@@ -237,7 +237,6 @@ if flag_shared:
 ## twitter: use curl to post shouts
 if flag_twitter:
     env.Append(CXXFLAGS='-DTWITTER')
-    #env.Append(extra_libs=['curl'])
 
 ## discord: use curl for discord webhooks as well, so always include
 env.Append(extra_libs=['curl'])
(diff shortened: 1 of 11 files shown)


grumpygabe:
realized I needed to do this asynchronously - will update today

cizra (changes requested):
I'm marking it as "Request changes" only to make sure it'll work withOUT the Discord config file. The rest is negotiable.

code/code/misc/discord.cc
```diff
@@ -0,0 +1,99 @@
...
+    config_options.add(configOnly);
+
+    po::variables_map vm;
+    po::notify(vm);
+    std::ifstream ifs(configFile.c_str());
+
+    if (!ifs.is_open()) {
+        vlogf(LOG_MISC, format("Discord webhooks: Failed to open config file '%s'") % configFile);
+        vlogf(LOG_MISC, format("Discord webhooks: ensure config is located in: '%s'") % std::filesystem::current_path());
+        return false;
```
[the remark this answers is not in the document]
grumpygabe: It should not - those are more informational.  I will double check and test, though.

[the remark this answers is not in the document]
grumpygabe: confirmed, no discord.cfg loads with defaults (no urls), which exits the function gracefully

code/SConstruct
```diff
@@ -234,8 +237,10 @@ if flag_shared:
 ## twitter: use curl to post shouts
 if flag_twitter:
     env.Append(CXXFLAGS='-DTWITTER')
-    env.Append(extra_libs=['curl'])
+    #env.Append(extra_libs=['curl'])
```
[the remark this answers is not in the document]
grumpygabe: fixed

code/code/misc/discord.cc
```diff
@@ -0,0 +1,99 @@
...
+    vlogf(LOG_MISC, format("Discord webhooks: '%s', '%s'") % channel % msg);
+
+    CURL* curl = curl_easy_init();
+    if (!curl) {
+        vlogf(LOG_MISC, "Discord webhooks: curl_easy_init() failed");
+        return false;
+    }
+
+    // sanitize and format the message
+    msg = format("{\"content\": \"%s\"}") % stripColorCodes(msg.cap()).escapeJson();
```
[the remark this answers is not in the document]
grumpygabe: I ended up removing it from the general case, in part because I hadn't actually removed it from the specific case.

code/code/misc/damage.cc
```diff
@@ -1116,12 +1124,34 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
             if (f->follower->desc) {
               f->follower->desc->session.groupKills++;
               f->follower->desc->career.group_kills++;
+              grouplist = format("%s, %s") % grouplist % f->follower->getName();
```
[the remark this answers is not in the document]
grumpygabe: this has been refactored to be O(n)

code/code/misc/discord.h
```diff
@@ -0,0 +1,17 @@
...
+
+    public:
+        static sstring CHANNEL_DEATHS;
+        static sstring CHANNEL_SYS;
+        static sstring CHANNEL_ACHIEVEMENT;
+
+        static int ACHIEVEMENT_THRESHOLD;
+
+        static bool doConfig();
+        static bool sendMessage(sstring channel, sstring msg);
```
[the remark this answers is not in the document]
grumpygabe: I am not currently planning on changing this for the initial pull, but we can revisit if you would like.

code/code/misc/discord.cc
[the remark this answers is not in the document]
grumpygabe: This would probably be better as part of a larger refactoring!

code/code/misc/discord.cc
```diff
@@ -96,4 +90,27 @@ bool Discord::sendMessage(sstring channel, sstring msg)
...
+{
+    if (channel == "") {
+        // no channel configuration, so bail
+        return;
+    }
+
+    vlogf(LOG_MISC, format("Discord webhooks: '%s'")  % msg);
+
+    // we want to do this asynchronously so POST lag doesn't hang the mud
+    std::thread thread(
```
cizra, line 105: I don't know, but pehaps you could say 
```cpp
std::thread thread(sendMessageAsync, channel, msg);
```
?

code/code/misc/discord.h
```diff
@@ -13,5 +14,6 @@ class Discord {
         static int ACHIEVEMENT_THRESHOLD;
 
         static bool doConfig();
-        static bool sendMessage(sstring channel, sstring msg);
+        static bool sendMessageAsync(sstring channel, sstring msg);
```
cizra, line 17: Is there a reason why this thing has to be publicly visible? Implementation details preferably wouldn't be exposed.

code/code/misc/damage.cc
```diff
@@ -1114,7 +1116,11 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
           if (k->desc) {
             k->desc->session.groupKills++;
             k->desc->career.group_kills++;
-            grouplist = k->getName();
+          }
+          // discord group naming
+          if (k->getName() != getName()) {
+            lastname = k->getName();
+            group_members++;
```
cizra, line 1123: It doesn't appear like this thing would be in a loop. Thus you'll get group_members = 0 or 1, below, making line 1137 always true.

code/code/misc/damage.cc
```diff
@@ -1124,7 +1130,17 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
             if (f->follower->desc) {
               f->follower->desc->session.groupKills++;
               f->follower->desc->career.group_kills++;
-              grouplist = format("%s, %s") % grouplist % f->follower->getName();
+            }
+              
+            // discord group naming
+            if (f->follower->getName() != getName()) {
```
cizra, line 1136: What exactly does this line do? I can't quite wrap my head around these pointers. Is it double-checking that the follower is following the leader? Checking that the leader is not following itself?

code/code/misc/damage.cc
cizra: Tho that's probably OK, as we update the counter later on.

code/code/misc/damage.cc
```diff
@@ -1133,21 +1149,22 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
...
+    if((isPc() || (isPet(PETTYPE_PET | PETTYPE_CHARM | PETTYPE_THRALL) && (master && master->isPc()))) && !v->isPc() && v->GetMaxLevel() >= Discord::ACHIEVEMENT_THRESHOLD) {
       // player killed an notable mob, send message to our discord webhook
       // ideally we'd like to list all the group members...
-      if(grouplist == "") {
-        // if no grouplist, just use the killer's name
-        grouplist = format("%s has") % killer;
+      if (group_members == 0) {
+        // killer is solo
+        achievement_msg = format(":crossed_swords: **%s** has taken down **%s**!") % getName() % v->getName();
       } else {
```
cizra, line 1158: Several ways to make it nicer.
```cpp
if (group_members == 0) {
...
} else if (group_members == 1) {
...
} else {
...
}
```
or
```cpp
achievement_msg =
  group_members == 0
  ? format(solo)
  : group_members == 1
  ? format(duo)
  : format(group)
```

code/code/misc/discord.h
grumpygabe: oversight, good catch!

code/code/misc/discord.cc
grumpygabe: this is better, I was overthinking it

cizra (approved):

code/code/misc/damage.cc
grumpygabe: flattened this up along the lines of the first example

code/code/misc/damage.cc
grumpygabe: this and the following conditionals are to ensure the entity with the killing blow is not listed twice

code/code/misc/damage.cc
grumpygabe: see above - the current entity will be either the master or in the follower list, so we check to make sure they don't get added again.

cizra (approved):

drewhershey (approved):
Just re-approving since Cizra's got canceled due to the required force push.

PR #471, opened 2024-01-12
Use pkg-config for finding libmariadbclient

cizra:


cizra:
This makes switching between different OSes / base images easier.

drewhershey (commented):
I'm not really sure if this is a problem on my end or with the PR, but here's what happens when I try to run the build using these changes:
```
$ scons -C code -j`nproc` CXX=g++-9 olevel=0 sneezy
scons: Entering directory `sneezymud/code'
scons: Reading SConscript files ...
/bin/sh: 1: pkg-config: not found
OSError: 'pkg-config --cflags --libs mariadb' exited 127:
  File "sneezymud/code/SConstruct", line 200:
    env.ParseConfig('pkg-config --cflags --libs mariadb')
  File "/usr/lib/python3/dist-packages/SCons/Environment.py", line 1612:
    return function(self, self.backtick(command))
  File "/usr/lib/python3/dist-packages/SCons/Environment.py", line 599:
    raise OSError("'%s' exited %d" % (command, status))
```

It looks like these changes expect the `pkg-config` executable to be available, which isn't available by default or installed as part of `build-essentials` on Ubuntu, at least. So I suspect this PR will break the Docker builds without further changes, though I haven't actually tested that.

Is there something I should do differently to test this?

cizra:
> Edit: I just realized that the Build and Test workflow was able to run successfully with these changes, so I'm trying to figure out why it didn't have the same problem I did. The only real difference I can see is that it installs the `software-properties-common` package - do you think that's what does it?

I don't know... does `software-properties-common` have a transitive dependency on `pkg-config`? https://packages.ubuntu.com/jammy/software-properties-common

Anyway, just install pkg-config and see how it fares. We'll fix the Dockerfile as needed. At least it's the official way to find the right libs, it appears. Before that, I had to maintain manual patches for compiling on Arch Linux, NixOS (all different, of course).

drewhershey (approved):
I installed the `pkgconf` package in Ubuntu, which is a fork of pkg-config that's actively maintained and provides the same binary and this branch compiled and ran fine, so looks like that's all that needs to happen.

Also, (not a problem with this PR, but just noting for posterity) clangd - the code completion/linting server I run in Ubuntu and hook into with VS Code - could no longer find the `mysql.h` file with the prefix removed from the include. Clangd requires a [compile_commands.json](https://clang.llvm.org/docs/JSONCompilationDatabase.html) file (which can be [generated by scons](https://scons.org/doc/production/HTML/scons-user/ch27.html)) to work fully/optimally, so I had to add the flag `-Ilibs/usr/include/mariadb` to this file to allow clangd/VS Code to find it. This would've happened automatically if I regenerated the compile_commands.json file via my modified SConstruct file I use while developing but it was just easier for me to do it manually.

So anyway, looks good on my end now.

PR #472, opened 2024-01-12
Update Build and Test GitHub Action and add scons package

drewhershey:
This workflow should optimally use the exact same compilation setup as in the Dockerfile that's used to create the production Docker images. This wasn't the case, so this PR updates it as such.

It also adds the scons package, which was already in the Dockerfile, and is needed here.


wershlak:
I kind of feel like this file is relevant. No?

wershlak (approved):
Looks good

PR #474, opened 2024-01-16
Add ca-certificates package to sneezy Docker image

drewhershey:
Discord integration requires this package installed to be able to send curl requests to Discord webhooks over HTTPS.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ FROM ubuntu:focal AS run
 
 ENV DEBIAN_FRONTEND=noninteractive
 
-RUN apt-get update && TZ=utc apt-get install --yes --no-install-recommends libboost-program-options1.71.0 libboost-regex1.71.0 libboost-filesystem1.71.0 libboost-system1.71.0 libmariadb3 libcurl4 libasan5 gdb && apt-get clean && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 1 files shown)


cizra (approved):
I thought you're switching away from this Dockerfile?

drewhershey:
> I thought you're switching away from this Dockerfile?

I did push another PR that would do so, but I think @wershlak had some reservations about it, so I started working on merging the two repos together instead as a possible alternative.

I haven't pushed anything related to that work yet, as I shelved it for the moment to help with @grumpygabe's Discord integration.

I'll incorporate this change into whichever method we end up using to unify the two repos (if any).

drewhershey:
But I also need to figure out how to get this required test to pass to allow merging after https://github.com/sneezymud/sneezymud/pull/472. I had assumed if a file got merged that didn't match a path in the Action configuration, the action wouldn't be required to merge the PR, which apparently isn't the case.

drewhershey:
The issue with this check never starting will be fixed once https://github.com/sneezymud/sneezymud/pull/475 is merged and this branch is rebased onto it.

drewhershey:
@cizra https://github.com/sneezymud/sneezymud/actions/runs/7577799904/job/20639279866?pr=474 - can see it successfully utilized the cache and reduced the total runtime of the job by about 10 minutes! Once we get version.h switched to a .txt file so main.cc/connect.cc don't recompile every time and force re-linking it should be even faster still.

PR #475, opened 2024-01-18
GitHub Actions - Add master branch compilation cache workflow and update test workflow

drewhershey:
This PR solves several issues related to our GitHub Actions.

The current `Build and Test` workflow defined in `.github/workflows/test.yml` simply ensures that the codebase compiles successfully and runs on every submitted PR. 

This is often a waste of time, as it's common for a PR not to modify any source files at all. Use, for instance, the example of a PR that updates some zonefiles. Compiling the code tells us nothing useful in this instance.

However, it's completely possible that a change to a zonefile could still cause the game to crash on boot. With the current configuration of the Build and Test workflow, this would only be discovered after the PR containing the zonefile changes had been approved and merged to master (assuming that the problem wasn't somehow caught before submission or by review/approval).

This problem can be solved as shown in the updated version of `test.yml` in this PR. By including MariaDB server in the packages, we can use the SQL dumps in `_Setup_files` to create and seed a test database within the VM used by the workflow, compile the code, and with a bit of shell scripting confirm that the game successfully boots up completely.

Now, when a PR featuring only non-source changes is submitted, the workflow will confirm that the game can boot utilizing these new changes before allowing merging the PR to master. This will even work for database updates, assuming the SQL dumps are updated as part of the PR.

However, this leaves the issue of having to compile the code for every PR, and even every time a PR is modified - even when no source files were modified. We can solve this problem by adding a new workflow - `create-master-cache.yml` - that runs whenever new code is merged to the master branch.

This new workflow utilizes the `actions/cache` GitHub Action to compile the master branch then save the resulting `.o` files, `libsneezy.a` file, and the `.sconsign.dblite` compilation database created by scons to a repo-specific cache provided for free by GitHub. With these files available during a build, scons will perform incremental compilation, only re-compiling files that have changed since the previous run and even skipping the linking step if possible.

`test.yml` can then be configured to utilize the most recent cache of build objects from the master branch by restoring the cache before beginning compilation, greatly speeding up the build (or skipping it almost entirely, if no source files changed).

I've tested these extensively in my own fork and this submission is the result of many iterations, so I'm very confident they'll work as expected. Here's some links to completed workflows utilizing this PR's code as proof:
- `create-master-cache.yml` compiling and caching the build in response to changes being pushed to master: https://github.com/drewhershey/sneezymud/actions/runs/7573889527/job/20627018125
- `test.yml` utilizing the cached build from master to speed up compilation, creating and seeding a test db, and verifying the game successfully boots:
https://github.com/drewhershey/sneezymud/actions/runs/7573926822/job/20627149859

Nothing will need to change as far as the configuration of the actual GitHub repo settings goes. Since I used the same workflow/job name in the new version of `test.yml`, it will still be set up to run on every PR and be required to pass before the PR can be merged. Because `create-master-cache.yml` is triggered by merges to master, it will run automatically in the background whenever a new PR is merged but won't block anything if it for some reason fails.

Notes:
- The compilation command used by both workflows is identical to that used by the Dockerfile that creates the production Docker images, except that fortify=1 is removed.
    - fortify=1 doesn't actually do anything. It's possible whoever added that meant to use harden=1, which adds many security features to the executable. However, the SConstruct file automatically ignores harden=1 if debug=1 is used, and debug=1 is required for sanitize=1. Changing this behavior would require  updates to the SConstruct file. Assuming we'd rather have ASan active than security features that are very unlikely to be useful, I just removed fortify=1 and didn't replace it with harden=1.

- The `test.yml` workflow now uses `ubuntu-22.04` (the current LTS) as the image for the VM, instead of `ubuntu-20.04`. This is because 20.04 has an older version of MariaDB installed by default that causes conflicts with the newer version we use, and causes the `create and seed db` step (and therefore the entire workflow) to fail.
    - Because Ubuntu 22.04 comes with a newer version of GCC installed by default but our production Docker images still use g++-9 on Ubuntu 20.04, I explicitly install and use g++-9 in these workflows to ensure compatibility. We'll want to remove this when we update the Docker builds to use 22.04.

### DIFF
--- a/.github/workflows/create-master-cache.yml
+++ b/.github/workflows/create-master-cache.yml
@@ -8,7 +8,7 @@ on:
 
 jobs:
   build_and_cache:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: Build and Cache
     steps:
       - name: Checkout this commit
@@ -45,7 +45,7 @@ jobs:
 
       - name: Compile code
         run: |
-          scons -C code -j`nproc` CXX=g++-9 pretty=1 debug=1 sanitize=1 olevel=2 sneezy
+          scons -C code -j`nproc` CXX=g++-9 pretty=0 debug=1 sanitize=1 olevel=2 sneezy
 
       - name: Cache compilation results
         uses: actions/cache/save@v4
--- a/.github/workflows/create-master-cache.yml
+++ b/.github/workflows/create-master-cache.yml
@@ -1,0 +1,66 @@
+# Caches created via actions/cache are not shared between PRs. However, a cache created on a parent branch is accessible by a child branch, including PRs that are attempting to merge to the parent branch. Therefore, if we create a cache of the build files every time changes are pushed to master, new PRs should be able to utilize those during their compilation to enable incremental compilation via scons's .sconsign.dblite file.
+name: Create Master Branch Compilation Cache
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  build_and_cache:
+    runs-on: ubuntu-latest
+    name: Build and Cache
+    steps:
+      - name: Checkout this commit
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Install dependencies
+        run: |
+          set -e
+          sudo apt-get update
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt-get install --yes --no-install-recommends g++-9 build-essential libboost-dev libboost-program-options-dev libboost-regex-dev libboost-filesystem-dev libboost-system-dev libmariadb-dev scons libcurl4-openssl-dev ca-certificates
+
+      # Used cached build objects from previous version of master, if available, to avoid a full recompilation if no source files changes.
+      - name: Restore cached build objects
+        uses: actions/cache/restore@v4
+        with:
+          path: |
+            code/.sconsign.dblite
+            code/objs/libsneezy.a
+            code/objs/main.o
+            code/objs/cmd
+            code/objs/disc
+            code/objs/game
+            code/objs/misc
+            code/objs/obj
+            code/objs/spec
+            code/objs/sys
+            code/objs/task
+          key: ${{ runner.os }}-scons-master-${{ hashFiles('code/code/**.cc', 'code/code/**.h') }}
+          restore-keys: |
+            ${{ runner.os }}-scons-master-
+
+      - name: Compile code
+        run: |
+          scons -C code -j`nproc` CXX=g++-9 pretty=1 debug=1 sanitize=1 olevel=2 sneezy
+
+      - name: Cache compilation results
+        uses: actions/cache/save@v4
+        with:
+          path: |
+            code/.sconsign.dblite
+            code/objs/libsneezy.a
+            code/objs/main.o
+            code/objs/cmd
+            code/objs/disc
+            code/objs/game
+            code/objs/misc
+            code/objs/obj
+            code/objs/spec
+            code/objs/sys
+            code/objs/task
+          # Use a unique key for each cache based on a hash of the source files, as GitHub Actions caches are immutable once created and a new cache won't overwrite the old one if the key is the same. Using restore-keys in the test.yml workflow will still allow the PR to find the most recent cache starting with ${{ runner.os }}-scons-master-.
+          key: ${{ runner.os }}-scons-master-${{ hashFiles('code/code/**.cc', 'code/code/**.h') }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -4,7 +4,7 @@ on:
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     name: Build and Test
     steps:
       - name: Checkout this commit
@@ -56,7 +56,7 @@ jobs:
 
       - name: Compile code. Will utilize cached build output from last update to master to perform incremental compilation, if available.
         run: |
-          scons -C code -j`nproc` CXX=g++-9 pretty=1 debug=1 sanitize=1 olevel=2 sneezy
+          scons -C code -j`nproc` CXX=g++-9 pretty=0 debug=1 sanitize=1 olevel=2 sneezy
 
       - name: Verify the game boots up successfully
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,16 +1,10 @@
 name: Build and Test
 on:
   pull_request:
-    paths:
-      # Only run on changes to files that are actually included in the build or this file itself
-      - "code/code/**.cpp"
-      - "code/code/**.cc"
-      - "code/code/**.h"
-      - ".github/workflows/test.yml"
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     name: Build and Test
     steps:
       - name: Checkout this commit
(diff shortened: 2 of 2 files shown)


.github/workflows/create-master-cache.yml
```diff
@@ -0,0 +1,66 @@
...
+            code/objs/spec
+            code/objs/sys
+            code/objs/task
+          key: ${{ runner.os }}-scons-master-${{ hashFiles('code/code/**.cc', 'code/code/**.h') }}
+          restore-keys: |
+            ${{ runner.os }}-scons-master-
+
+      - name: Compile code
+        run: |
+          scons -C code -j`nproc` CXX=g++-9 pretty=1 debug=1 sanitize=1 olevel=2 sneezy
```
cizra, line 48: Checking targets |
Compiling code/task/task_sharpen.cc
Compiling code/task/task_sit.cc
Checking targets /
Compiling code/task/task_skin.cc
Compiling code/task/task_sleep.cc
Compiling code/task/task_spell_friends.cc
Checking targets -

Can we have `pretty=0`? Or squelch these "Checking targets" messages someelsehow.

drewhershey:
The version of test.yml added by this PR was run here and succeeded, so there's another example of it working.

drewhershey:
Also, merging this PR will unblock this one: https://github.com/sneezymud/sneezymud/pull/474

.github/workflows/test.yml
```diff
@@ -1,25 +1,74 @@
...
-      # Only run on changes to files that are actually included in the build or this file itself
-      - "code/code/**.cpp"
-      - "code/code/**.cc"
-      - "code/code/**.h"
-      - ".github/workflows/test.yml"
 
 jobs:
   build_and_test:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
cizra, line 7: Going to fail after an incompatible update. Perhaps pin it to 22.04? Or are we fine with an occasional breakage?

.github/workflows/test.yml
```diff
@@ -1,25 +1,74 @@
...
+            for phase in tables views data ; do
+              [ -d "_Setup-data/sql_$phase/$db" ] || continue
+              for sql in _Setup-data/sql_$phase/$db/*.sql ; do
+                echo "loading '$sql'"
+                mysql $db < $sql
+              done
+            done
+          done
+
+      - name: Restore cached build objects from master branch
```
cizra, line 38: We now have a sample build that did *not* have the cache, so the next build, which will use the cache, will show us the difference in speed.

.github/workflows/test.yml
```diff
@@ -1,25 +1,74 @@
...
+            code/objs/task
+          key: ${{ runner.os }}-scons-master-${{ hashFiles('code/code/**.cc', 'code/code/**.h') }}
+          restore-keys: |
+            ${{ runner.os }}-scons-master-
+
+      - name: Compile code. Will utilize cached build output from last update to master to perform incremental compilation, if available.
+        run: |
+          scons -C code -j`nproc` CXX=g++-9 pretty=1 debug=1 sanitize=1 olevel=2 sneezy
+
+      - name: Verify the game boots up successfully
```
cizra, line 61: Should be possible now to plug in some character creation and test commands. Not necessarily in this PR, but still.

.github/workflows/test.yml
```diff
@@ -1,25 +1,74 @@
...
+        run: |
+          set -e
+          ./code/sneezy 2>&1 | tee output.log &
+          SNEEZY_PID=$!
+          while kill -0 $SNEEZY_PID ; do
```
cizra, line 66: How does this while loop work anyway? Which signal are we sending, TERM? And why -0 - my manpage says that it'll go to all processes in the current group, which would kill everything, including the GitHub action runner. Obviously that can't be right, so what's the idea here?

cizra (approved):

.github/workflows/create-master-cache.yml
drewhershey: Just pushed a commit that takes care of this.

.github/workflows/test.yml
drewhershey: For some reason I was under the impression that ubuntu-latest and ubuntu-20.04 were the only ones available, but after looking ubuntu-22.04 does exist, so I've switched them over to it. Good call.

.github/workflows/test.yml
drewhershey: We won't have a cached build to utilize until the master branch has had changes merged to it at least once, as that's what triggers the workflow that creates the cache.

The reason I didn't have the test.yml workflow create a cache of its build is because `actions/cache` doesn't allow caches to be shared between PRs for some reason. It's just a cache miss every time, no matter what. There's a longstanding issue open about that "feature" in its repo.

But caches are shareable with PRs if the cache was made from the branch the PR is trying to merge with, so therefore if we keep a cached build from the newest version of the master branch, any new PR should be able to basically compile just the "diff" between itself and master.

So merging this PR will trigger the script to create a build cache from the master branch, after which point any new PR or existing PR that's rebased/updated will be able to utilize that cache to speed up its compilation.

.github/workflows/test.yml
drewhershey: This is exactly why I was looking at your functional tests recently, and asked about them on Discord.

In the future I'd like to add a step after booting the game up that executes those tests, and fails the workflow if any of them fail.

drewhershey:
@cizra One thing I've noticed as I added this new commit is that `main.cc` and `sys/connect.cc` are being recompiled every time, even when a usable cache exists and they weren't modified, because they both include `version.h` which *is* modified as defined in the SConstruct on every new commit.

I'm not entirely sure what the purpose of that version.h file is, but I noticed that you added it a while back. Can you think of any way we could modify things to avoid having it force recompilation of those two files?

.github/workflows/test.yml
cizra: Interesting trick. Just tested, it totally works.
I would have done something like this instead:
```sh
# let's have a timeout
(sleep 600; killall sneezy) &
timeoutpid=$!
cleanup() {
  kill $timeoutpid
}
trap cleanup EXIT

# Sneezy output gets sent to a temporary anonymous pipe
# ./sneezy | grep -q "Entering game loop." doesn't work :(
# grep quits after getting the first response, shutting down the pipe, which makes the shell send SIGPIPE to Sneezy, which it'll ignore for some reason. So this trick doesn't work, after all.
# This one works fine tho.
./sneezy 2>&1 | (grep -q "Entering game loop"; killall sneezy)
```

cizra (approved):

cizra:
> @cizra One thing I've noticed as I added this new commit is that `main.cc` and `sys/connect.cc` are being recompiled every time, even when a usable cache exists and they weren't modified, because they both include `version.h` which _is_ modified as defined in the SConstruct on every new commit.
> 
> I'm not entirely sure what the purpose of that version.h file is, but I noticed that you added it a while back. Can you think of any way we could modify things to avoid having it force recompilation of those two files?

It's there to provide version information to the running MUD. We could have SCons drop the Git hash into version.txt and rewrite the code to read it.

drewhershey:
Assuming you meant to resolve the conversation and not close the PR entirely. 😄

.github/workflows/test.yml
cizra: Oops, we should also make sure it distinguishes crash from timeout from success.

Also, extract into a separate shell script for testing it locally.

drewhershey:
> > @cizra One thing I've noticed as I added this new commit is that `main.cc` and `sys/connect.cc` are being recompiled every time, even when a usable cache exists and they weren't modified, because they both include `version.h` which _is_ modified as defined in the SConstruct on every new commit.
> > I'm not entirely sure what the purpose of that version.h file is, but I noticed that you added it a while back. Can you think of any way we could modify things to avoid having it force recompilation of those two files?
> 
> It's there to provide version information to the running MUD. It gets displayed in logs, and also shown in the banner during connection. We could have SCons drop the Git hash into version.txt and rewrite the code to read it.

That sounds like a good solution. I'll tackle this sometime soon in a new PR.

PR #476, opened 2024-01-23
Replace version.h with lib/version.txt to fix test workflow caching

drewhershey:
The SConstruct file was writing the current commit hash to version.h on every compile. This file was then being
included in two places (main.cc and connect.cc), which was causing those two files to be recompiled on every
build, whether or not they actually changed. This was causing the updates test.yml GitHub workflow to create
a new cache every run, making it less useful.

This PR changes the SConstruct file to instead write the commit hash (and now the commit date as well)
to a text file in the lib folder, version.txt. The hash and date in this file are then read in inside the `main` function
and stored in a global variable, which is then used in the code where the hash from version.h was used previously.
This prevents the recompilation issue and allows the test workflow to fully utilize the caching behavior.

Additionally, as the development of Sneezy is now open source and more of a steady, ongoing process of
adding and updating features, there's likely no more reason to use specific version numbers
when referring to future "eras" of the game. This PR removes all references to SneezyMUD
5.4, instead just using simply SneezyMUD going forward. Versioning will be provided via the
lib/version.txt file, which can be used in the code to display to users the hash and date of the
commit from which the game is currently running.


code/code/sys/client.cc
```diff
@@ -570,10 +570,9 @@ int Descriptor::read_client(char* str2) {
         return 0;
       }
 
-      sprintf(buf, "$n just rented with the %s client.", MUD_NAME_VERS);
-      act(buf, FALSE, character, NULL, NULL, TO_ROOM);
+      act(format("$n just rented with the %s client.") % MUD_NAME, false,
```
cizra, line 573: I wonder if we still have anyone using that client.
Also, even if we rename the MUD, we're not really renaming the old Sneezy client, so I don't see a reason to change this line. As it's already done, though, there's also not much reason to revert this change :)

code/code/sys/configuration.cc
```diff
@@ -183,3 +186,30 @@ bool Config::doConfiguration(int argc, char* argv[]) {
...
+      format("Failed to open %s.") % (std::filesystem::current_path() / path));
+    return {};
+  }
+
+  return {(std::istreambuf_iterator<char>(file)), {}};
+}
+
+std::string readVersionFromFile() {
+  const std::filesystem::path versionPath =
+    std::filesystem::current_path() / Config::DataDir().c_str() / "version.txt";
```
cizra, line 204: If desired, you could avoid setting this path by moving the reading of version to happen after `if (chdir(Config::DataDir().c_str())` in `main()`.

code/code/sys/configuration.cc
```diff
@@ -183,3 +186,30 @@ bool Config::doConfiguration(int argc, char* argv[]) {
   }
   return true;
 }
+
+std::string readFileContents(const std::string& path) {
```
cizra, line 190: Is there a reason why this is not in an anonymous namespace?

cizra: Also, I'm confused - is this meant to return a string or an istringstream?

code/code/sys/configuration.cc
```diff
@@ -183,3 +186,30 @@ bool Config::doConfiguration(int argc, char* argv[]) {
...
+  std::ifstream file(path);
```
cizra, line 191: Does it _actually_ work? Are you leaking the file descriptor, or else are you accessing a stack-allocated object after it goes out of scope?

code/code/sys/configuration.h
```diff
@@ -1,7 +1,15 @@
 #pragma once
 
+#include <string>
+#include <string_view>
```
cizra, line 4: Propagating this include into 70 code files is hurting compilation times, even if just a bit.

code/code/sys/configuration.cc
cizra: What is the reason you split the reading into a different function than the parsing?

code/code/sys/configuration.h
```diff
@@ -1,7 +1,15 @@
...
 #include "sstring.h"
 
+inline constexpr const char* MUD_NAME = "SneezyMUD";
+extern sstring MUD_NAME_VERS;
+
+std::string readFileContents(std::string_view);
+std::string readVersionFromFile();
```
cizra, line 11: Short and simple enough be in main.cc.

code/code/sys/configuration.h
```diff
@@ -1,7 +1,15 @@
 #pragma once
 
+#include <string>
+#include <string_view>
 #include "sstring.h"
 
+inline constexpr const char* MUD_NAME = "SneezyMUD";
+extern sstring MUD_NAME_VERS;
```
cizra, line 8: Nit: Owned by main, but exported by configuration.

code/code/sys/connect.cc
```diff
@@ -2881,77 +2879,65 @@ int Descriptor::sendLogin(const sstring& arg) {
...
+      tm current_date;
+      time_t conv_time;
+      conv_time = time(&conv_time);
+      localtime_r(&conv_time, &current_date);
+      int years_of_service = current_date.tm_year;
+      if ((current_date.tm_mon < 4) ||
+          (current_date.tm_mon == 4 && current_date.tm_mday <= 5)) {
+        years_of_service -= 93;
+      } else {
+        years_of_service -= 92;
```
cizra, line 2922: I'd expect there to be some API / data structures where you can take the difference of 2 dates, and round that to years.

code/code/sys/connect.cc
```diff
@@ -2881,77 +2879,65 @@ int Descriptor::sendLogin(const sstring& arg) {
...
+      }
+      return years_of_service;
+    }();
```
cizra, line 2925: Why a lambda? If you want a scope, you could just use {}.

code/code/sys/create_character.cc
```diff
@@ -5,6 +5,7 @@
 //////////////////////////////////////////////////////////////////////////
 
 #include <sstream>
+#include "configuration.h"
```
cizra, line 8: How did it compile before?

lib/txt/ansi/login1.ans
```diff
@@ -1,4 +1,4 @@
-[?7h[255D[0m        [0m/    [0;32m'  [0;31m<[1;34m|   [0;31m<[1;34m|   [0m\  /     /[32m  "     '             ,   [0m\               [K
+[255D[0m        [0m/    [0;32m'  [0;31m<[1;34m|   [0;31m<[1;34m|   [0m\  /     /[32m  "     '             ,   [0m\               [K
```
cizra, line 1: What's this change do?

cizra (approved):

code/code/sys/client.cc
drewhershey: Something had to be changed either way, since MUD_NAME_VERS was previously a `const char*` and is now an `sstring` after my changes. So I figured I might as well fix the sprintf usage even though MUD_NAME is still a `const char*`.

It would be nice to get rid of all that old Brutius client code. Though I know we recently had at least one player using it still, miraculously. He even sent me a working copy of the client to use in tracking down a crash bug.

code/code/sys/configuration.cc
drewhershey: I was trying to keep the line that initializes MUD_NAME_VERS by calling this function as high up in `main()` as possible, so that it's the first output one sees when starting the boot process (since that's how you had it set up previously).

code/code/sys/configuration.cc
drewhershey: > Is there a reason why this is not in an anonymous namespace?
> What is the reason you split the reading into a different function than the parsing?

"Don't let good code go to waste", basically. 😄 I figured it was a useful function that could be used elsewhere in the code in the future, so might as well make it accessible. Could probably find a better place for it than `configuration.h/cc` I suppose, though not sure where that would be off the top of my head.

We do have a `file_to_sstring` function defined elsewhere that does something similar, but using much older/messier syntax and no RAII types. I tested short-circuiting that function by just calling my `readFileContents` function instead and it worked mostly fine, though needs more testing and a few tweaks, so I figured I'd do that in a separate PR sometime.

> Also, I'm confused - is this meant to return a string or an istringstream?

Not really sure why I did it this way now that I look at it again. Probably a leftover from how I had the `readVersionFromFile` function coded previously (it went through several iterations). But you're right, I do just convert it to an `istringstream` immediately after returning the value as a `std::string`, which seems silly. I'll change that.

code/code/sys/configuration.cc
drewhershey: It definitely works, I tested it about a hundred times in various ways while I was writing and proofing the code.

I don't see how it would leak a file descriptor, though, since `std::ifstream` uses RAII and manages the lifetime of the file descriptor it uses internally and automatically.

I'm not sure how to answer "are you accessing a stack-allocated object after it goes out of scope" in this context. Am I missing something obvious again? I don't see anything being used outside of its own local scope in those two functions.

code/code/sys/configuration.cc
cizra: > "Don't let good code go to waste"

There's also the YAGNI principle... Oh, yeah, I wanted but forgot to ask about `file_to_sstring`. If it's actually going to be used, feel free to retain it.

code/code/sys/configuration.cc
cizra: ifstream uses RAII to _open_ that file, fine, but who _closes_ it? Who owns the underlying OS-level handle? Is the file just left open?

I'm still a bit confused about the ownership of the stream buffer thing...
3) Initializes the iterator and stores the value of is.rdbuf() in a data member. If is.rdbuf() is null, then end-of-stream iterator is constructed. https://en.cppreference.com/w/cpp/iterator/istreambuf_iterator/istreambuf_iterator

Suppose you had a large file that takes several trips to disk to read. Would your code still work, after `std::ifstream file` goes out of scope?

code/code/sys/configuration.h
drewhershey: Is it actually owned by main? In my mind it was owned by configuration and simply initialized in main out of necessity. It "feels" like it should be in `configuration.h` to me, but after looking at all the files that are including `configuration.h` and which of those are actually accessing those variables, there's probably a better place to declare them. Though I'm not sure where that would be, yet. Perhaps `db.h`?

code/code/sys/configuration.h
drewhershey: Yeah, I think in retrospect there's probably no reason to worry about writing functions that could be used in the future like this, and instead I should just write stuff to be as efficient as possible in the context at hand. I'll combine these functions and move them into an anonymous namespace in `main.cc`.

code/code/sys/configuration.h
drewhershey: See above - I'll combine the functions and move `readVersionFromFile` into an anonymous namespace in `main.cc`.

code/code/sys/configuration.h
cizra: No idea, and not much of a preference. I'll leave it to you to pick a place for it.

code/code/sys/connect.cc
drewhershey: Just used a self-executing anonymous lambda to initialize a static variable, so that the code only gets executed once per boot since that value won't change once it's been calculated.

code/code/sys/create_character.cc
drewhershey: The MUD_NAME and MUD_NAME_VERS variables were declared in a different header previously, which this file already included. I had to add this once I moved their declarations into `configuration.h`.

lib/txt/ansi/login1.ans
drewhershey: I'm not sure what that ANSI character was intended to be, but it was messed up and was just displaying `7h` at the top-left of that login ASCII graphic. So I just took it out, and it looks fine now as far as I can tell.

cizra (approved):

code/code/sys/configuration.cc
drewhershey: Doesn't `ifstream` own the underlying handle? I thought that was the whole point of RAII types - the `ifstream` destructor gets called once the `ifstream` itself goes out of scope, and the destructor takes care of the underlying handle it created.

> Suppose you had a large file that takes several trips to disk to read. Would your code still work, after std::ifstream file goes out of scope?

I guess I'm confused too - how would this happen? The function reads the entire contents of the target file in at once via `return {(std::istreambuf_iterator<char>(file)), {}};`.

`std::istreambuf_iterator<char>(file)` is an iterator that can read characters from the file object. When default-constructed (which is what I'm doing via shorthand by simply passing empty initializer braces `{}` as the second parameter), it becomes an end-of-stream iterator.

So `{(std::istreambuf_iterator<char>(file)), {}}` is a range created from two iterators: the start of the file and the end of the file. This range represents the entire contents of the file.

So `std::istringstream fileContents{...}` creates a std::istringstream object and initializes it with the contents of the file, allowing us to treat the contents of the file as a stream.

Sorry if that's not even what you were asking about. Let me know if I still didn't address your actual question.

code/code/sys/configuration.h
drewhershey: So after looking through this again, it still feels to me that `configuration.h` is the logical place to declare these variables. Most of the files that reference either of them already had `configuration.h` included, so it just requires adding the include to a couple extra files. Any objections?

code/code/sys/configuration.cc
drewhershey: First I've heard of YAGNI, somehow, but I like it.

drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/476/commits/17de1b40cc65e5f489f434878ab0267af991d881

code/code/sys/configuration.h
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/476/commits/17de1b40cc65e5f489f434878ab0267af991d881

code/code/sys/configuration.h
drewhershey: Fixed in https://github.com/sneezymud/sneezymud/pull/476/commits/17de1b40cc65e5f489f434878ab0267af991d881

drewhershey:
Just pushed a new commit so I wouldn't mess up all the conversations via a force push. Let me know if you'd prefer cleaner commits instead, and I can combine this newest one into the first and force push before merging.

code/code/sys/configuration.cc
drewhershey: Just realized I was overcomplicating things, and didn't even need to use the `std::istreambuf_iterator` stuff now that I got rid of the `readFileContents` code. Maybe I didn't even need it then.🤦‍♂️

Anyway, pushed https://github.com/sneezymud/sneezymud/pull/476/commits/e6f061225d9ecb2e29ac3c59caa3fdf334b72033 to address that. Seems a lot cleaner now.

cizra: > the ifstream destructor gets called once the ifstream itself goes out of scope, and the destructor takes care of the underlying handle it created.

That's exactly why I asked if this thing works. How can you read from a file after its file handle has gotten out of scope?

I don't know how these iterators work, but files have a nasty habit of sometimes being larger than the whole RAM, thence my question.

cizra (approved):

code/code/sys/configuration.cc
drewhershey: This code no longer exists, but in the original commit the line `return {(std::istreambuf_iterator<char>(file)), {}};` was creating an initializer list containing iterators to the start and end of the file, which was then being passed to the `std::string` constructor via the `return { }` shorthand. `std::string` has a constructor that takes two iterators and constructs a string from them.

It's equivalent to
```c++
std::string fileContents{(std::istreambuf_iterator<char>(file)), {}};
return fileContents;
```

So - assuming my understanding of all of this is accurate - the ifstream and ifstream_iterators are never actually used outside of the local scope of that function, as all that's actually being returned is a std::string that's being constructed in-place from them.

As far as trying to read in a file larger than the available RAM - that definitely wasn't an edge case I had considered, haha. I suppose it could be a valid concern, though it seems unlikely and I don't know how I'd go about accounting for it. But regardless, I've removed the `readFileContents` function entirely at this point and am just reading in the first two lines directly, instead.

PR #478, opened 2024-01-28
discord threading + bonus new death shouts!

grumpygabe:
Made some minor adjustments to the discord code to use only a single long-lived worker thread.

Also added about nine new taunt shouts to the death code.

### DIFF
--- a/code/code/misc/damage.cc
+++ b/code/code/misc/damage.cc
@@ -1007,7 +1007,7 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
             taunt_buf = format("Back to the inn with you, %s!") % v->getName();
           } else if (chance == 31) {
             taunt_buf =
-              format("Is %s the best SneezMUD has to offer?  Pathetic!") %
+              format("Is %s the best SneezyMUD has to offer?  Pathetic!") %
               v->getName();
           } else if (chance == 32) {
             taunt_buf = format(
--- a/code/code/misc/damage.cc
+++ b/code/code/misc/damage.cc
@@ -1035,7 +1035,7 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
           doShout(taunt_buf);
           discord_taunt_msg =
             format(":skull: %s shouts, \"%s\"") % getName().cap() % taunt_buf;
-          Discord::sendMessage(Discord::CHANNEL_DEATHS, discord_taunt_msg);
+          Discord::sendMessageAsync(Discord::CHANNEL_DEATHS, discord_taunt_msg);
         } else {
 #if 1
           if (v == this && isPc())
@@ -1212,7 +1212,7 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
                           getName() % v->getName() % grouplist.str() % lastname;
       }
 
-      Discord::sendMessage(Discord::CHANNEL_ACHIEVEMENT, achievement_msg);
+      Discord::sendMessageAsync(Discord::CHANNEL_ACHIEVEMENT, achievement_msg);
     }
 
     strcpy(buf2, v->name.c_str());
--- a/code/code/misc/damage.cc
+++ b/code/code/misc/damage.cc
@@ -882,7 +882,7 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
           // added for information to other players that haven't died but
           // didn't want to write an information channel etc
           // should be fun to come up with new shouts
-          int chance = ::number(1, 26);
+          int chance = ::number(1, 35);
           if (chance == 1) {
             taunt_buf = format("WOO! And %s goes down! HA!") % v->getName();
           } else if (chance == 2) {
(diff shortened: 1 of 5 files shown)


code/code/misc/damage.cc
```diff
@@ -990,11 +990,51 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
...
+                "You should have stayed in the safety of the tavern, %s!") %
+              v->getName();
+          } else if (chance == 29) {
+            taunt_buf =
+              format("%s's skills are as rusty as their armor!") % v->getName();
+          } else if (chance == 30) {
+            taunt_buf = format("Back to the inn with you, %s!") % v->getName();
+          } else if (chance == 31) {
+            taunt_buf =
+              format("Is %s the best SneezMUD has to offer?  Pathetic!") %
```
cizra, line 1010: typo: SneezMUD

code/code/misc/discord.h
```diff
@@ -1,19 +1,29 @@
...
+#include <thread>
+#include <mutex>
+#include <utility>
 
 class Discord {
-    private: 
-        Discord();
-        static bool sendMessageAsync(sstring channel, sstring msg);
+  private:
+    Discord();
```
cizra, line 11: Unimplemented private constructors are so 2010. Since C++11, we have `Discord() = delete`;

code/code/misc/discord.h
```diff
@@ -1,19 +1,29 @@
...
+    static bool sendMessageAsync(sstring channel, sstring msg);
+    static void messenger();
+    static std::thread messenger_thread;
+    static std::mutex queue_mutex;
+    static bool stop_thread;
+    static std::queue<std::pair<sstring, sstring>> message_queue;
```
cizra, line 17: You might (or might not, it's mostly a matter of taste) want to look into the PIMPL way of hiding these implementation details. Check database.h and database.cc for an example.

code/code/misc/discord.cc
```diff
@@ -16,97 +19,137 @@ namespace po = boost::program_options;
...
-        return false;
-    }
-
-    po::store(parse_config_file(ifs, config_options), vm);
-    po::notify(vm);
-
-    return true;
+  using std::string;
+  // worker thread setup
+  stop_thread = false;
```
cizra, line 41: Can we provide that default value where we declare or define this variable?
Or do we expect doConfig to be called multiple times per boot?

code/code/misc/discord.cc
```diff
@@ -16,97 +19,137 @@ namespace po = boost::program_options;
...
+  message_queue.push({channel, msg});
+}
 
-    curl_easy_setopt(curl, CURLOPT_URL, webhookURL);
-    curl_easy_setopt(curl, CURLOPT_HTTPHEADER, headers);
-    curl_easy_setopt(curl, CURLOPT_POSTFIELDS, content);
+void Discord::messenger() {
+  while (true) {
+    std::lock_guard<std::mutex> lock(queue_mutex);
+    if (message_queue.empty() && !stop_thread) {
```
cizra, line 129: Busy loop, consuming 100% of available CPU time. Should use some sort of an event system instead.

code/code/misc/discord.cc
```diff
@@ -16,97 +19,137 @@ namespace po = boost::program_options;
...
+
+  po::options_description configOnly("Configuration File Only");
+  configOnly.add_options()("deaths_webhook",
+    po::value<string>(&CHANNEL_DEATHS)->default_value(empty_string),
+    "see discord.h")("sys_webhook",
+    po::value<string>(&CHANNEL_SYS)->default_value(empty_string),
+    "see discord.h")("achieve_webhook",
+    po::value<string>(&CHANNEL_ACHIEVEMENT)->default_value(empty_string),
+    "see discord.h")("achieve_threshold",
+    po::value<int>(&ACHIEVEMENT_THRESHOLD)->default_value(80), "see discord.h");
```
cizra, line 56: I strongly prefer the way the code was formatted before. This one is breaking lines at totally random places, apparently.

code/code/misc/discord.cc
```diff
@@ -16,97 +19,137 @@ namespace po = boost::program_options;
...
-    }
-
-    po::store(parse_config_file(ifs, config_options), vm);
-    po::notify(vm);
-
-    return true;
+  using std::string;
+  // worker thread setup
+  stop_thread = false;
+  messenger_thread = std::thread(Discord::messenger);
```
cizra, line 42: If Discord is not enabled, we shouldn't start the thread. Perhaps move the initialization after the config check?

code/code/misc/discord.cc
cizra: Perhaps you could `wait` for an event. I haven't done it in C++, so I don't know the very best way of doing this, but https://en.cppreference.com/w/cpp/atomic/atomic_wait looks promising.

cizra (changes requested):

code/code/misc/discord.cc
grumpygabe: That makes sense.

code/code/misc/discord.cc
grumpygabe: Agreed.  Something in my workspace went a bit overboard with the formatting.

code/code/misc/discord.cc
grumpygabe: I'm looking at adding a simple sleep to the thread:
```
    if (message_queue.empty() && !stop_thread) {
      std::this_thread::sleep_for(std::chrono::milliseconds(1000)); // sleep for a second
      continue;
    }
    ```

Comment (cizra) [reply code/code/misc/discord.cc:41]: A sleep incurs a context switch, unlike an event. When waiting for an event, the kernel removes the thread completely from scheduling.

File: code/code/misc/discord.cc (hunk 6 above)
Comment (drewhershey) [reply code/code/misc/discord.cc:56]: This is just clang-format not being set up to handle chained function calls like this, I think. I dealt with it in my branch by turning clang-format off and back on via comments, then manually formatting the code in between like this:
```c++
// clang-format off
configOnly.add_options()
  ("deaths_webhook", po::value<string>(&CHANNEL_DEATHS)->default_value(empty_string), "see discord.h") 
  ("sys_webhook", po::value<string>(&CHANNEL_SYS)->default_value(empty_string), "see discord.h") 
  ("achieve_webhook", po::value<string>(&CHANNEL_ACHIEVEMENT)->default_value(empty_string), "see discord.h") 
  ("achieve_threshold", po::value<int>(&ACHIEVEMENT_THRESHOLD)->default_value(80), "see discord.h");
// clang-format on
```

grumpygabe: neat!

code/code/misc/discord.h
grumpygabe: Done

code/code/misc/discord.h
grumpygabe: I think I am going to pass on doing it this way.  It seems interesting (and a useful technique), but I don't know if I want to get into those weeds just now.

code/code/misc/discord.cc
grumpygabe: Ended up using condition_variable for this, which seems to work nicely.

cizra (changes requested):

cizra (approved):
Looks fine, apart from the bit with shutdown.

cizra (approved):

code/code/misc/discord.cc
```diff
@@ -16,97 +21,137 @@ namespace po = boost::program_options;
...
+      sendMessage(message.first, message.second);
+      lock.lock();
     }
+  } while (!stop_thread);
+}
 
-    vlogf(LOG_MISC, format("Discord webhooks: '%s'")  % msg);
+bool Discord::doCleanup() {
+  // let the thread finish
+  messenger_thread.join();
```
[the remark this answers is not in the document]
grumpygabe: I think you're correct.  Should be corrected now.

code/code/misc/discord.h
```diff
@@ -1,19 +1,32 @@
...
-        Discord();
-        static bool sendMessageAsync(sstring channel, sstring msg);
+  private:
+    static bool sendMessage(sstring channel, sstring msg);
+    static void messenger();
+    static std::thread messenger_thread;
+    static std::condition_variable cv;
+    static std::mutex queue_mutex;
+    static bool stop_thread;
+    static std::queue<std::pair<sstring, sstring>> message_queue;
```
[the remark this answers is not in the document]
grumpygabe: I'm interested in figuring this out but I am going to leave this as is for now.

PR #489, opened 2024-05-30
Fully implement `I` and `J` zonefile commands for prop loads, and fix a bunch of miscellaneous bugs in various zonefiles

drewhershey:
I finally finished up a side project I started a while back to write a JavaScript script that validates and parses zonefiles into JSON format. In the process I found a bunch of small mistakes in various zonefiles that I went ahead and fixed.

Additionally, I went ahead and fully implemented the `I` and `J` zonefile commands that allow the loading of 'prop' items on mobs, allowing builders to equip mobs with throwaway weapons and armor when they spawn even through load-on-death.

Prop items are no-rent/no-locate with a value of 0 and decay 8 real-life hours from the time they're looted, and are clearly marked as props in their short description and name. Players that manage to loot them can freely use them until they decay but can't monogram/sell/scrap/rent them. Equipping a mob with a full set of prop armor and weapons should make the fight significantly more difficult, or can just allow the mobs to fully utilize their available skillsets.

Items with with `(max_exist - current_exist) < 5` can't be loaded as props to prevent prop loads from blocking the loads of the real version of items. Prop versions of low-max-exist items like artifact weapons can be done by creating a copy of the item under a new vnum, giving it a high max_exist, then only using that vnum for the prop load commands in the zonefile.


code/code/sys/db.cc
```diff
@@ -1570,8 +1599,9 @@ bool zoneData::bootZone(int zone_nr) {
 
     if (rs.command == 'M' || rs.command == 'O' || rs.command == 'B' ||
         rs.command == 'C' || rs.command == 'K' || rs.command == 'E' ||
-        rs.command == 'P' || (rs.command == 'T' && !rs.if_flag) ||
-        rs.command == 'R' || rs.command == 'D' || rs.command == 'L')
+        rs.command == 'I' || rs.command == 'P' ||
+        (rs.command == 'T' && !rs.if_flag) || rs.command == 'R' ||
```
cizra, line 1603: minor nit: These parentheses are redundant (tho that won't hurt).
I think readability would be improved by moving the R case onto the line with P, and leaving the && case in its own line. In this case, () wouldn't be necessary anymore (though this is a minor improvement).

cizra (approved):

code/code/sys/db.cc
drewhershey: I agree, but in interest of not forcing you to re-approve the PR I'm just going to go ahead and resolve this since it's just a aesthetic issue primarily, as I'd like to get this merged this morning. If I have to tweak anything in this code later on I'll make this change then.

PR #493, opened 2024-06-12
add new death shouts and refactor this a bit

wershlak:
Added a few shouts and I think it's a bit more readable this way

### DIFF
--- a/code/code/misc/damage.cc
+++ b/code/code/misc/damage.cc
@@ -27,6 +27,229 @@
 #define DAMAGE_DEFINE 0
 #define DAMAGE_DEBUG 0
 
+namespace {
+
+  void deathShout(TBeing* v, TBeing* being) {
+    sstring taunt_buf;
+    sstring discord_taunt_msg;
+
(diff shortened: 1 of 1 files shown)


code/code/misc/damage.cc
```diff
@@ -27,6 +27,229 @@
 #define DAMAGE_DEFINE 0
 #define DAMAGE_DEBUG 0
 
+namespace {
+
+  void deathShout(TBeing* v, TBeing* being) {
```
cizra, line 32: nit: Can it be `nullptr`? Probably not? Then we could use references, instead of pointers - use the typesystem to describe non-nullness.

code/code/misc/damage.cc
```diff
@@ -27,6 +27,229 @@
...
+    sstring taunt_buf;
+    sstring discord_taunt_msg;
```
cizra, line 34: nit: could be declared down below

cizra (approved):

PR #495, opened 2024-06-20
Fix a bug in RoomSave related to immortal db roomexit/roomextra tables

drewhershey:
This fixes two malformed queries that were causing entries in the roomexit/roomextra tables to not be properly deleted when using the rsave command. The queries had one too many parameters which was causing the where clauses to not find any matches, leading to many duplicate rows in those tables in the immortal db.

### DIFF
--- a/code/code/misc/create_rooms.cc
+++ b/code/code/misc/create_rooms.cc
@@ -11,8 +11,7 @@
  *****************************************************************************/
 
 #include <stdio.h>
-
-#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/predicate.hpp>
 
 #include "room.h"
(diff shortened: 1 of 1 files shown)


code/code/misc/create_rooms.cc
```diff
@@ -11,8 +11,7 @@
  *****************************************************************************/
 
 #include <stdio.h>
-
-#include <boost/algorithm/string.hpp>
+#include <boost/algorithm/string/predicate.hpp>
```
drewhershey, line 14: This change just resolves a deprecation warning I was seeing by using a more specific header file to provide the `iequals` symbol.

cizra (approved):

PR #496, opened 2024-07-02
Update SQL dumps from live database using _Setup-data/dump-databases.sh script

drewhershey:
Updates the SQL dump files in `_Setup-data/sql_data/[sneezy | immortal]` with the current contents of the live database from the sneezy-mud.com instance, to capture all the changes made by Vasco and others since 2021.

Only includes dumps for the tables that contained actual data in the original dumps (the sql_data directory), so no personal info, etc.


wershlak (approved):
Thanks for doing this!

PR #497, opened 2024-07-02
Add a new `mvresponse` option to the `low` command

drewhershey:
This command will allow imms with appropriate powers to move mob responses from the imm database to the production database. Currently the only way to actually add/edit mob responses is through the web builder tools. Also cleaned up the code for `doLow` a bit.

I intentionally didn't merge this functionality into `mvmob`, as Vasco said it's likely the LOW would want to be able to activate mob responses separately at a chosen time, instead of automatically when moving the mob itself.

### DIFF
--- a/code/code/cmd/cmd_low.cc
+++ b/code/code/cmd/cmd_low.cc
@@ -8,6 +8,7 @@
 #include <cmath>
 #include <algorithm>
 #include <stdio.h>
+#include <sstream>
 
 #include "extern.h"
 #include "handler.h"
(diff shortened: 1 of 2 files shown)


drewkht:
Just an aside: this was done by converting the old LOW command from `code/low/sqladdresp.cc` to an in-game command instead of a standalone executable that runs from the server.

wershlak (approved):
I think this command could use a rewrite to improve syntax and parsing but this looks better than before.

PR #498, opened 2024-07-06
Update 27100

kjsevertson:
cleaning up area randomizers to help quest mobs load correctly, adding loot to container objects


drewhershey (approved):
We chatted on Discord about these changes; converting the multiple, random-chance `A` commands to single, guaranteed `A` commands makes sense as the entire range of rooms should be viable spawn locations for the mobs, and this prevents the issue where sometimes all the ? commands would fail resulting in a mob simply not loading.

The new `L` commands give some incentive to venture in to this zone for the chance at level-appropriate consumables loot in the coffins.

PR #499, opened 2024-07-07
king's palace update - july 2024

kjsevertson:
Adding they key mob for non-violent solution to queen's headache quest. added loot chest to hidden room. fixed room number on cauldron load


wershlak (approved):
ok

PR #500, opened 2024-07-17
Removing Elderly Alchemist from trainer list

kjsevertson:
removing very confusing zonefile entry regarding quest mob that is no longer trainer


wershlak (approved):
Gunna merge this one and the other one. Probably makes sense to put both changes in a single merge request. Just FYI for next time

PR #502, opened 2024-07-30
Correcting mobs not loading 27100

kjsevertson:
After noticing that half of the mobs in the zone were not loading, I ran through some boot zone commands to see if I could force them to load. I discovered that some of the mobs would always load, some of them would sometimes load, and some of them would never load.

There were mistakes in some of the IF_FLAG characters for about half of the mob loads.

The area randomizer command in zonefiles is infrequently used, and I'm not that familiar with it.  So I didn't catch the error initially.


wershlak (approved):
seems sensible

PR #503, opened 2024-07-30
Toggle Revamp 7/31/2024

kjsevertson:
Reassigning old Ranger quests to Warrior class, adding new toggles for Thief quests Lvl 15, 20, and 45. Adding in new toggles for Hero of GH quests, and adding some toggles for exchanging Deikhan class 'holy swords' for blunt/pierce type weapons


drewkht:
Force pushed a second time after rebasing to include all the recent commits to master.

drewhershey:
Force pushed a second time after rebasing to include all the recent commits to master.

code/code/misc/gaining.cc
```diff
@@ -868,6 +895,114 @@ void TPerson::setSelectToggles(TBeing* gm, classIndT Class,
...
+
+        *this += *note;
+
+        act(
+          "You suddenly realize there's an unfamiliar note in your "
+          "inventory. Where did that come from?",
+          false, this, nullptr, nullptr, TO_CHAR, ANSI_CYAN);
+      }
+
+      if (getLevel(Class) >= 20 && !hasQuestBit(TOG_THIEF_L15_ELIGIBLE) &&
```
wershlak, line 920: I'll probably just merge this but it seems like it's not an ideal way to handle this. We want to know two things: Are they eligible and have we already given them the message or note. However because we are untoggling stuff as we progress in the quest these statements look unwieldy. 

We could have a specific toggle TOG_GIVEN_THIEF_L20_NOTE or something that _doesn't_ get toggled off as the quest progresses and make this statement a lot easier.

Is that better? Or just better here but worse in some other way?

wershlak (approved):

PR #508, opened 2024-09-26
Wind proc

wershlak:
Add three tiers of a gust themed weapon proc

### DIFF
--- a/code/code/misc/combat.cc
+++ b/code/code/misc/combat.cc
@@ -1892,6 +1892,7 @@ void TBeing::stopFighting() {
   REMOVE_BIT(specials.affectedBy, AFF_ENGAGER);
   REMOVE_BIT(specials.affectedBy, AFF_RIPOSTE);
   REMOVE_BIT(specials.affectedBy, AFF_FOCUS_ATTACK);
+  REMOVE_BIT(specials.affectedBy, AFF_FLURRY);
 
   if (gCombatList == this)
     gCombatList = next_fighting;
@@ -2535,6 +2536,15 @@ int TBeing::hit(TBeing* target, int pulse) {
       if (bSuccess(SKILL_CHIVALRY))
         ;  // do nothing, just for sake of learning
     }
+  }
+
+  // Add attacks for flurry
+  if (isAffected(AFF_FLURRY)) {
+    fx += 3;
+    REMOVE_BIT(specials.affectedBy, AFF_FLURRY);
+    act("<W>$n executes a <b>flurry<z><W> of attacks<z>", FALSE, this, 0, 0, TO_ROOM);
+    act("<W>You execute a <b>flurry<z><W> of attacks.<z>", FALSE, this, 0, 0, TO_CHAR);
+
   }
 
   // this affect is added after a successful parry
--- a/code/code/misc/defs.h
+++ b/code/code/misc/defs.h
@@ -121,6 +121,7 @@ const uint64_t AFF_CLARITY =
 const uint64_t AFF_FLIGHTWORTHY = uint64_t(0x0000000100000000LLU);
 const uint64_t AFF_FOCUS_ATTACK = uint64_t(0x0000000200000000LLU);
 const uint64_t AFF_ORIENT = uint64_t(0x0000000400000000LLU);
+const uint64_t AFF_FLURRY = uint64_t(0x0000000800000000LLU);
 const int AFF_MAX = 35;
 
 // these are used to pass deletion bitvectors through functions
--- a/code/code/spec/spec_objs.cc
+++ b/code/code/spec/spec_objs.cc
@@ -7333,6 +7333,9 @@ extern int livingVines(TBeing* vict, cmdTypeT cmd, const char*, TObj* o, TObj*);
 extern int dkSword(TBeing* vict, cmdTypeT cmd, const char*, TObj* o, TObj*);
 extern int moltenWeapon(TBeing* vict, cmdTypeT cmd, const char* arg, TObj* o, TObj*);
 extern int glacialWeapon(TBeing* vict, cmdTypeT cmd, const char* arg, TObj* o, TObj*);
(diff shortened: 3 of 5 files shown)


code/code/misc/defs.h
```diff
@@ -121,6 +121,7 @@ const uint64_t AFF_CLARITY =
 const uint64_t AFF_FLIGHTWORTHY = uint64_t(0x0000000100000000LLU);
 const uint64_t AFF_FOCUS_ATTACK = uint64_t(0x0000000200000000LLU);
 const uint64_t AFF_ORIENT = uint64_t(0x0000000400000000LLU);
+const uint64_t AFF_FLURRY = uint64_t(0x0000000800000000LLU);
 const int AFF_MAX = 35;
```
drewhershey, line 125: Does `AFF_MAX` need to be incremented?

code/code/spec/spec_objs_wind_procs.cc
```diff
@@ -0,0 +1,195 @@
...
+  TBeing* ch;
+  TBeing* combatant;
+
+  if (!o)
+    return FALSE;
+  if (!(ch = dynamic_cast<TBeing*>(o->equippedBy)))
+    return FALSE;  // weapon not equipped (carried or on ground)
+
+  if (cmd == CMD_GENERIC_PULSE) {
+    if ( (combatant = ch->fight()) && (!::number(0,11)) ) {
```
drewhershey, line 130: What's the point of assigning ch->fight() to combatant here?

Also, this is an 8.3% chance every 3.6 seconds while fighting to have celerite autocast on oneself while holding a weapon with this proc, correct? Which would be doubled when dual wielding weapons with the proc?

code/code/spec/spec_objs_wind_procs.cc
```diff
@@ -0,0 +1,195 @@
...
+      doCeler(ch, o);
+      return TRUE;
+    }
+    return FALSE;
+  }
+
+  if (cmd == CMD_OBJ_HIT && vict) {
+    if (!::number(0, 3)) {
+      if (!::number(0, 4))
+        return doGust(ch, vict, o);
```
drewhershey, line 140: `25% * 20% = 5%` chance per hit to knock the victim down with no chance for the victim to save, right?

What will happen if two characters that get lots of attacks both dual wield weapons with this proc and attack the same mob? Could this make even high level casters completely trivial?

code/code/spec/spec_objs_wind_procs.cc
```diff
@@ -0,0 +1,195 @@
...
+    if ( (combatant = ch->fight()) && (!::number(0,7)) ) {
+      doCeler(ch, o);
+      return TRUE;
+    }
+    return FALSE;
+  }
+
+  if (cmd == CMD_OBJ_HIT && vict) {
+    if (!::number(0, 3)) {
+      if (!::number(0, 3))
```
drewhershey, line 168: Even higher gust chance - `25% * 25% = 6.25%` per attack, which I'm assuming is intentional as the proc is higher tier?

code/code/spec/spec_objs_wind_procs.cc
```diff
@@ -0,0 +1,195 @@
...
+        return doGust(ch, vict, o);
+      return doWindDamage(ch, vict, o);
+    }
+  }
+
+  if (cmd == CMD_OBJ_HITTING && vict) {
+    if (!::number(0, 49)) {
```
drewhershey, line 175: 2% chance per hit to generate three extra attacks with the mainhand (since the extra hits are just added directly to fx in `TBeing::hit()`), each of which could potentially generate three more attacks (though it would be really rare)?

drewhershey (changes requested):
A lot of these comments are just me parsing the code to make sure I understand what it's doing and that it's the behavior/numbers you intended. Feel free to just resolve those without responding to them if you want.

The only actual potential bugs I think I see are:
1) Needing to increment AFF_MAX
2) Not using `TBeing::fallOffMount` to dismount the victim in `doGust`, as otherwise certain edge cases aren't handled.

Everything else is just balance or formatting related, so if you're good with it I'm good with it.

code/code/misc/combat.cc
```diff
@@ -2537,6 +2538,15 @@ int TBeing::hit(TBeing* target, int pulse) {
     }
   }
 
+  // Add attacks for flurry
+  if (isAffected(AFF_FLURRY)) {
+    fx += 3;
```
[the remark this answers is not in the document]
wershlak: Astute observation. I actually was thinking to use flurry for a thief ability later. So it was coded in this way on purpose and might make more sense in that context. I'd be open to changing it if people feel it's just dumb or wrong.

code/code/spec/spec_objs.cc
```diff
@@ -7271,145 +7271,113 @@ extern int weaponBlinder(TBeing*, cmdTypeT, const char*, TObj*, TObj*);
...
-  TObj*);
-extern int sweepsScratch(TBeing* ch, cmdTypeT cmd, const char* arg, TObj* o,
-  TObj*);
-extern int sweepsSplitJoin(TBeing* ch, cmdTypeT cmd, const char* arg, TObj* o,
-  TObj*);
-extern int graffitiMaker(TBeing* ch, cmdTypeT cmd, const char* arg, TObj* o,
-  TObj*);
-extern int graffitiObject(TBeing* ch, cmdTypeT cmd, const char* arg, TObj* o,
-  TObj*);
+extern int deikhanSword(TBeing* vict, cmdTypeT cmd, const char* arg, TObj* o, TObj*);
```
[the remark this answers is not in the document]
wershlak: Is it set to 80 by default? Mind if I change it?

code/code/misc/defs.h
wershlak: Maybe. I'll see.

PR #511, opened 2024-11-21
Updating Jewel of Judgment proc to trade the HP loss for mana regen

kjsevertson:
Jewel of Judgment (currently on an artifact item that does not load) drains Hp/pulse. This update returns a value of Mana to simulate a trade.


drewhershey:
Closing this as it's being redone in greater detail in https://github.com/sneezymud/sneezymud/pull/529

PR #518, opened 2025-02-11
update summon to ignore immunity for group members

wershlak:
Just that

### DIFF
--- a/code/code/disc/disc_hand_of_god.cc
+++ b/code/code/disc/disc_hand_of_god.cc
@@ -485,14 +485,14 @@ int summon(TBeing* caster, TBeing* victim, int level, short bKnown) {
   int location = 0;
   TRoom* room = NULL;
 
-  int immun = victim->getImmunity(IMMUNE_SUMMON);
   TMonster* tmon = dynamic_cast<TMonster*>(victim);
 
   // check for immunity to summon
   // prevent some mobs used in quests from being moved around
-  if (((immun > 0) && ::number(0, 100) <= immun) ||
-      (tmon && ((tmon->mobVnum() == Mob::TIGER_SHARK) ||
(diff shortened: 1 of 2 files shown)


code/code/disc/disc_hand_of_god.cc
```diff
@@ -490,7 +490,8 @@ int summon(TBeing* caster, TBeing* victim, int level, short bKnown) {
 
   // check for immunity to summon
   // prevent some mobs used in quests from being moved around
-  if (((immun > 0) && ::number(0, 100) <= immun) ||
+  if ((!caster->inGroup(*victim)) ||
+      ((immun > 0) && ::number(0, 100) <= immun) ||
       (tmon && ((tmon->mobVnum() == Mob::TIGER_SHARK) ||
                  (tmon->mobVnum() == Mob::ELEPHANT)))) {
     act("You feel unable to summon $N.", FALSE, caster, NULL, victim, TO_CHAR);
```
drewhershey, line 497: I think this should be `&&` shouldn't it? As in, if they're not in your group AND any of these other things are true, don't summon them? Also you can simplify the logic a bit here since the `isImmune` function does the roll for you.

```c++
  TMonster* tmon = dynamic_cast<TMonster*>(victim);

  if (!caster->inGroup(*victim) &&
      (victim->isImmune(IMMUNE_SUMMON, WEAR_BODY) ||
        (tmon && (tmon->mobVnum() == Mob::TIGER_SHARK ||
                   tmon->mobVnum() == Mob::ELEPHANT)))) {
```

wershlak: Oops. Yeah totally. It was such a small change I forgot to pay attention. I fixed it and did some testing

drewhershey (approved):

PR #521, opened 2025-02-26
update spell parser

wershlak:
Updated the spell parser for mage/shaman spellcasting. The parser
now requires a unique match for spells. You might notice you have 
to give it a bit more.
For instance: 'cast d i' used to match detect invisibility but now
requires you to type 'cast de i' to disambiguate it from 
dispel invisibility

---

```
H:261 M:900 V:144 E:1,008,628,612 > cast protection from
No such spell exists.

H:261 M:900 V:144 E:1,008,628,612 > cast protection from fire
Although you no longer need to, you begin an incantation to facilitate your spell.
While not absolutely necessary, you trace a rune with your left hand to facilitate your spell in forming.
You touch a fiery red pebble to your forehead.
You glow with a faint red aura for a brief moment.
```

```
H:14 P:100.0 V:333 E:1.000 > cast telepathy blah blah blah
You concentrate intensely on your task.

H:14 P:100.0 V:333 E:1.000 >
With a flourish, you draw the final line of the magic pattern.
You utter the final words of the incantation.
You telepathically send the message, "blah blah blah"
```

### DIFF
--- a/code/code/misc/spell_parser.cc
+++ b/code/code/misc/spell_parser.cc
@@ -42,6 +42,18 @@
 
 #include <boost/algorithm/string.hpp>
 
+std::tuple<spellNumT, sstring> TBeing::parseSpellNum(const sstring& base, const sstring& args) const {
+
+  if (base.trim() == "")
+    return std::make_tuple(TYPE_UNDEFINED, args);
+
+  spellNumT found = searchForSpellNum(base, EXACT_NO);
+  if (found != TYPE_UNDEFINED)
+    return std::make_tuple(found, args.trimLeft());
+
+  return parseSpellNum(base.dropLastWord(), " " + base.lastWord() + args);
+}
+
 int TBeing::useMana(spellNumT spl) {
   int arrayMana;
   int rounds;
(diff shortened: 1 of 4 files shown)


code/code/misc/spell_parser.cc
```diff
@@ -538,24 +550,36 @@ char* skip_spaces(char* sstring) {
...
+        ret = i;        
+      }
     }
   }
 
-  return TYPE_UNDEFINED;
+  if (matches != 1)
+    return TYPE_UNDEFINED;
+
+  return ret;
```
drewhershey, line 582: By adding the third, optional parameter with a default of `true`, it looks like you've effectively modified the default behavior of `searchForSpellNum`, as now when the function is called without overriding that default with `false` it will use the new `matches` functionality.

It's only called a couple other places in the code, like https://github.com/sneezymud/sneezymud/blob/0bf0010a2e65a04a65f705672ea78ee8061abd72/code/code/disc/disc_shaman_alchemy.cc#L26 for instance, but now that code will potentially behave differently than before.

Have you tested the code in the other places `searchForSpellNum` is used to see if they still work? If they do, the extra parameter might not even be necessary, and using `matches` could just be the default behavior everywhere. Otherwise, you'd need to pass `false` for the third param in those other places the function is called.

Alternatively, if you just change the default value of `bool unique` to `false`, you could then just do `spellNumT found = searchForSpellNum(base, EXACT_NO, true);` where you call it in `parseSpellNum`, since that's the only location that needs the new behavior and the other call sites would continue using the previous behavior. Though it'd probably still be good to double check that `searchForSpellNum` still works everywhere, if you haven't already.

drewhershey (commented):
Looks good to me, aside from the one thing I pointed out that may or may not actually be an issue.

Otherwise, I'm assuming you've tested the new parsing with a pretty wide array of spells and confirmed it works for them all? Should be a nice improvement.

code/code/misc/spell_parser.cc
wershlak: I thought implementing the change in this way gave me an easy path to change the behavior back easily if we discovered the change didn't work for those use cases. So to answer your question... I considered that it's used in other places, I brewed a couple potions but didn't test scroll creation. Overall, I think it's fine but happy to change the default if you think that's better

drewhershey: If you tested it out and are good with it all then it's cool with me 👍

drewhershey (approved):

PR #529, opened 2025-03-30
Thiefredux1

kjsevertson:
the beginning of the Thief class redo. mostly, the big chase for thief updates is more world and object interaction. Traps will become Gadgets and will include a wide range of opportunities to interact with gear. Thieves will become specialists in the application of bleeding, bruising, poisoning, etc. New skills will support the direct actual application of these, like the revamp to Stab, but other new skills will allow thieves to impart some chances for these effects to their group mates by way of temporary gear manipulation and alteration. 

Spikes - adding interaction with several skills, including Slam, Stab, Chop, and several warrior skills. Most of these skills already add a small amount of damage, but the spikes will now have a chance to cause bleeding. Additionally, there is a chance that the SPIKED tag will be removed from the object as they tear off. This is in preparation for an upcoming Thief class advanced discipline skill that will be able to add the SPIKED tag to objects.

Stab - this skill now targets a specific limb, and has a much higher chance of causing a bleeding condition on the limb that has been hit. If certain conditions are met, Stab will also sever the limb. This only happens when the limb is at half health or lower, is Bleeding or Bruised, and if the mob fails a Constitution check.

Sap - this skill is meant as a play style option for the thief class as a blunt-weapon Backstab alternative. It's a stealth-based ambush style attack, much like the current iteration of Backstab, but with slightly less damage, a later learning curve, and a limb damage feature. The limb damage aspect, which hits only upper limbs,  stacks to do a 2-round limb paralysis effect if certain conditions are met. It's unlikely that these will be met early, which gives a reason to give multiple attempts to for more than one thief to attempt the skill. Namely, these require lower limb health, a failure on a Constitution save, and high enough damage from the attack. As this is meant to thematically be an ambush overpower, attackers that are taller than their victims will receive a damage bonus. Further bonuses can be gained from using two-handed weapons, or using spiked weapons.

Trap effects - for all trap types except Teleport, Disease, and Poison, new effects have been added. Mostly these have to do with the application of limb damage, embedded objects, or conditionals like Blind or Curse. The number of targeted limbs and the severity of the effect is determined by trap level.

Rest/Sleep - as an attempt to deepen a QoL roleplay, there are now factors which will affect the rate of regeneration as provided by the Rest/Sleep task. These include being encamped, being Hidden, having a campfire, and a dog/cat to cuddle with. It isn't huge, but it does allow players some options for agency while passive healing.

Caltrops - Thrown glass bottles that break upon hitting their target will not have a possibility of creating a pile of glass object which can sometimes create an object that embeds in the WEAR_FOOT of a mob or player that leaves the room and isn't flying or levitating. It's a manipulation of some small object physics which will play into upcoming advanced discipline thief skills.

Wetness - Thrown drink containers that break upon hitting their target will now apply Wetness to the victim in proportion to the drinkUnits in the container. This is some ground work for upcoming thief skill interaction.

Frostengulfed - several spells and spell-like effects apply this, which mostly checks to break objects held in bags and inventories. This update will now potentially apply the frostbite disease to victims that are wet.

Ability checks - There are already many existing situations, like spells and skills, which check against specific stats. More are being added with new skills and some older skills will include them to even out. The nature of this change is that each stat corresponds to a status debuff like Poison, Curse, Stupidity, etc. which will apply 'disadvantage' to the check by requiring two successes. This allows a player to stack on debuffs to build toward a more dangerous skill/spell to land with greater effect.

Procs - adding some bow and quiver props the add poison to arrows or ignite them. while they stand on their own, the intent is for thieves to be able to temporarily add them to objects. The IcyDeath proc is where I cut my teeth on a lot of this coding project and so it stands as a little bit of a monolith, but it also blends scaled success into success with a cost/danger. I had initially planned to make more developed procs in this manner for many of the old artifacts, but it seems unlikely that I will continue on this path until after thieves have been deeply massaged. 

Many existing procs have been massaged to either carry a cost or require a situational. For instance, the blood drain proc now steals life and causes a bleed, but also steals life from the owner if there isn't a pool of blood in the room. Regeneration now fires on a pulse and provides output. Wand recharging requires that a mage be holding the wand in order for it to replace the charge.


code/code/cmd/cmd_stab.cc
```diff
@@ -5,604 +5,308 @@
...
+        tAff.level = 0;
+        tAff.duration = PERMANENT_DURATION;
+        tAff.modifier = DISEASE_EYEBALL;
+        tAff.location = APPLY_NONE;
+        tAff.bitvector = AFF_BLIND;
+        victim->affectTo(&tAff);
+        victim->rawBlind(this->GetMaxLevel(), tAff.duration, SAVE_NO);
+      } else {
+        victim->makePartMissing(limb, false, this);
+        sstring buf = format("You slice $N's %s right off!") % victim->describeBodySlot(limb);
```
wershlak, line 251:
```
You puncture an arena guard's right leg with your shortsword, leaving a bloody gash!
The bloody, mangled right leg of an arena guard goes flying through the air and bounces once before it rolls to a stop.
You slice an arena guard's right leg right off!
Debug: rc value is 1
```
This seems to happen about a third of the time in my limited test. How oftern should it happen? Is it dependent on something?

wershlak:
```
H:483 P:100.0 V:170 E:1,001,381,531 > stab guard
You puncture an arena guard's head with your shortsword, leaving a bloody gash!
The bloody, mangled head of an arena guard goes flying through the air and bounces once before it rolls to a stop.
You slice an arena guard's head right off!
Debug: rc value is 1

H:483 P:100.0 V:167 E:1,001,384,219  <guard=hurt> <Rogue/tank=excellent> <2.1 secs lockout> >
An arena guard does a nifty fighting move, but then fumbles about looking foolish.

H:483 P:100.0 V:199 E:1,001,384,219  <1.7 secs lockout> > lim guard
Check whose limbs?

H:483 P:100.0 V:199 E:1,001,384,219 > lim guard
Check whose limbs?

H:483 P:100.0 V:199 E:1,001,384,219 >
The bruising on your right foot throbs painfully.

H:483 P:100.0 V:199 E:1,001,384,219 > loo
Inside the Arena Walls
  You are standing in a passageway that makes its way through the inside of the
arena.  The solid stone walls, floor, and ceiling give the arena a feeling of
stability.  A wooden door fortified with metal bands lies to the west.
This area is brightly lit.
You can see exits to the west, northeast, and southwest.
An arena guard's corpse is headless, and congealed blood lies around the neck.
The bloody, mangled, severed head of an arena guard is lying here.
Warluk is standing here.
A sparkling green sprite is flying about.
.....a sparkling green sprite is flying around.
.....a sparkling green sprite glows with a bright light!
A small puddle of blood is on the floor here.
```
need to handle beheadings better

code/code/misc/skill_dam.cc
```diff
@@ -409,6 +409,11 @@ int TBeing::getSkillDam(const TBeing* victim, spellNumT skill, int level,
       dam = genericDam(victim, this, skill, DISC_THIEF, level, adv_learn, 2.01,
         REDUCE_NO, !isPc(), TRIM_NO);
       break;
+    case SKILL_SAP:
+      // Similar to backstab (2.00) but using 1.90 multiplier for slightly less damage
+      dam = genericDam(victim, this, skill, DISC_THIEF, level, adv_learn, 1.90,
```
wershlak, line 414: dam is calculated based on when you learn it so I think the fact that sap learns faster means it's doing a lot less damage

code/code/cmd/cmd_stab.cc
kjsevertson: Ah, I see. Debug line removed, and problem found.

`    if (victim->getCurLimbHealth(limb) >= victim->getMaxLimbHealth(limb) / 2) {`

it should be...
`    if (victim->getCurLimbHealth(limb) <= victim->getMaxLimbHealth(limb) / 2) {`

In order for a limb to become severed, it should need to be either Bruised, Bleeding, or Infected and have less than half of its maximum limb health. This should mean that a stab early in the fight won't outright sever on the first shot, but the chance gets higher as the fight goes on, if RNG falls in favor of a limb repeat. Things like weapon sharpness and whether the object isSpiked will affect limb damage also.

kjsevertson: Also, I don't want beheadings to be a thing via Stab. So some clarifications to how WEAR_HEAD is handled were added. I missed a loop hole that was allowing them to pass through.

Also I'm reordering a few lines, just to make the output a little bit more narratively linear.

code/code/misc/skill_dam.cc
kjsevertson: Ah, okay. I'll bring it into alignment with Backstab in terms of learning rate, but retain the start_20

code/code/misc/spell_info.cc
```diff
@@ -2571,9 +2571,9 @@ void buildSpellArray() {
       LEARN_DO_1, START_DO_NO, LEARN_DO_NO, LEARN_DIFF_UNUSUAL, 0.0, 0, 0);
 
   discArray[SKILL_SAP] = new spellInfo(SKILL_THIEF, DISC_THIEF, DISC_THIEF_FIGHT, 
-    STAT_STR, "sap", TASK_NORMAL, LAG_3, POSITION_STANDING, MANA_0, LIFEFORCE_0, 
-    PRAY_0, 0, SYMBOL_STRESS_0, "", "", "", "", START_20, LEARN_3, START_DO_30, 
-    LEARN_DO_5, START_DO_NO, LEARN_DO_NO, LEARN_DIFF_SKILLS, 0.0, 0, 0);
+    STAT_STR, "sap", TASK_NORMAL, LAG_5, POSITION_STANDING, MANA_0, LIFEFORCE_0, 
+    PRAY_0, 0, SYMBOL_STRESS_0, "", "", "", "", START_20, LEARN_1, START_DO_20, 
+    LEARN_DO_2, START_DO_NO, LEARN_DO_NO, LEARN_DIFF_SKILLS, 0.0, 0, 0);
```
wershlak, line 2576:
```
sap                         Current:  (poor)      Potential:  (very good)

H:478 P:100.0 V:196 E:1,001,530,693 > prac
Basic disciplines:
               Thief Abilities : Learnedness: 100%
```
With START_20 and LEARN_1 you can't fully learn this skill. You'd have to somehow get to 120% learnedness

LEARN_2 means you learn 2% per 1% in the disc so you could do START_50 and LEARN_2 to make it max at 100% for example.

code/code/cmd/cmd_stab.cc
wershlak: This seems better in limited testing.

drewhershey:
I'm gonna close this PR out now that I've finished separating all the changes into more targeted, smaller PRs. Should allow us to get this thing tackled more quickly.

PR #531, opened 2025-04-21
Tweaking crit fail to reduce damage, especially for healthy characters

parkscs:


parkscs:
This prayer always has a small chance to crit fail and is nuking players ever since I took the handcuffs off prayer damage.  We've had a few frustrating deaths recently due to this and so I'm proposing a minor tweak, essentially cutting the damage by between 1/3 and 2/3, depending on how healthy the player is.  The general idea is that this will help low max HP characters survive more often, as long as they're casting at full health.

See:

Rood chants mysterious and sacred words.
Rood's symbol glows with power!
Something goes dramatically wrong.
Rood suddenly erupts in a tower of flames!
Rood is dead! R.I.P.
Rood loses his ability to fly.
The yellow aura fades from Rood's eyes.
The silver aura in Rood's eyes fades.
The protection around Rood fades.
Rood doesn't seem as holy.
Someone stops following you.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -1041,8 +1041,13 @@ int spontaneousCombust(TBeing* caster, TBeing* victim, int level, short bKnown,
...
           NULL, TO_ROOM);
-        act("You are suddenly consumed in a tower of flames! <R>FIRE!!!<1>",
-          FALSE, caster, NULL, NULL, TO_CHAR);
+        act("You are suddenly consumed in a smaller tower of flames, as the Gods are merciful... for now! <R>FIRE!!!<1>", 
+              FALSE, caster, NULL, NULL, TO_CHAR);
+
+        // Reducing the damage to avoid one-shotting players at full health (as often)
+        dam /= 3;
+        dam *= 2 -
+		      1.0*((double) caster->getHit() / (double) caster->hitLimit());
```
wershlak, line 1050: Why reduce it more if the health is full? I don't see the benefit of added complexity here.
If we're trying to reduce the damage and death from backfires why not just divide the dam by 3 as you've done above and be done with it? It shock and penalize failure but with less lethality.

parkscs:
I originally just had damage /= 2.  However the thought was to split the
difference based on health because people advocated for counterplay and
players can definitely control their health %.  I also wanted to try and
account for the pure caster clerics who are likely to have less max hp, but
who may be able to more easily stay at max hp as they aren’t tanking.  This
is slightly more complexity but adds a bit more risk for cleric tanks
casting offensive prayers, while making it a bit safer for pure casters who
may not have a lot of HP to begin with.

In practice I don’t expect it to change the outcome in many cases, but that
was the thought process.

wershlak (commented):
Looks good thanks!

wershlak (approved):

PR #533, opened 2025-05-07
Add an armor set to Brightmoon Library zone and increase chances for some smaller loads

drewhershey:
Pulled from https://github.com/sneezymud/sneezymud/pull/529


wershlak (approved):
Seems ok

PR #539, opened 2025-05-07
Add some hobbit race-specific effects to the pipeweed drug

drewhershey:
Pulled from https://github.com/sneezymud/sneezymud/pull/529


wershlak (approved):
Be sure to add more snacks in the shire

PR #544, opened 2025-06-12
Quick and easy shaman drain fix

parkscs:
Quick and easy fix to the issue Vasco identified.  Admittedly I haven't tested it locally so do sanity check my logic, but we've pinpointed the issue to it treating the calculation as ints and rounding the percentage to 0 unless the player has the skill at 100.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved accuracy of health and vitality gains from certain spells, resulting in more precise effects during gameplay.

### DIFF
--- a/code/code/disc/disc_shaman_skunk.cc
+++ b/code/code/disc/disc_shaman_skunk.cc
@@ -235,9 +235,9 @@ int lichTouch(TBeing* caster, TBeing* victim, int level, short bKnown,
 
   int dam = caster->getSkillDam(victim, SPELL_LICH_TOUCH, level, adv_learn);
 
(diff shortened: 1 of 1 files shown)


parkscs:
I took a look at task_whittle flagged by the rabbit - tValue is a double, so the / 100 shouldn't present a problem.

drewhershey:
> I took a look at task_whittle flagged by the rabbit - tValue is a double, so the / 100 shouldn't present a problem.

I *think* it's identifying places where both numerator and denominator are ints and the division occurs first due to order of operation. If that's the case it could still potentially result in a bad value that then gets implicitly converted to a double when passed into tValue or multiplied with a double or whatever.

However, even if I'm not wrong, it's outside the scope of this PR so I'm of the opinion you can just decide what you want to do with it. Maybe we could just paste the coderabbit results into a new issue to document it and get back to it some other time?

PR #555, opened 2025-07-10
Update suitsets

agednotdead:


kjsevertson (commented):
I don't think the space in "#satyr hide" is correct. I don't see any sets in loudest that have spaces. I think you might need to use a hyphen or underscore

kjsevertson (approved):

PR #588, opened 2025-10-27
debride task and new getToolSlot function

kjsevertson:
There was some request recently in the discord about being able to remove ITEM_RUSTY from objects. It seemed like a pretty easy fix and since I've been looking into mechanics to remove/add/manipulate other item flags, I took on the task. No pun intended.

It's nothing wild. It checks against the Blacksmithing skill, since it doesn't seem like there's any real reason to add a new skill just for this. No difficulty modifier, but a bonus to task time reduction if there's advanced discipline learning. Measures task time against item volume. Doesn't discriminate against object type. Requires a file to be held, though. And to that point, I found it frustrating that there is no standardized method of checking for an equipped tool of an appropriate type in an appropriate slot. So, I made one.

Introducing getToolSlot(wearSlotT, toolType). it's pretty straightforward, but clean and doesn't make my head want to die when I see a task/skill looking for the string of the tool "name" in the object's name.

This function can (maybe will) be applied to quite a number of messy calls in repair. 

## Summary by CodeRabbit

* **New Features**
  * Added a learnable "debride" command/skill to remove rust from steel/iron equipment.
  * Debriding runs as a multi-step task with busy messaging, time based on item size, and consumes stamina/movement.
  * Requires a file in the primary hand; tools wear with use and may break, affecting outcomes.
* **Documentation**
  * New help entry describing syntax, requirements, and behavior.


wershlak (approved):
Love it thanks!

PR #594, opened 2025-11-14
added the ability to add room procs via redit

kjsevertson:
Adding a line to the Redit menu to allow for the assigning of room procs.

Also added the 'rproc' option to show command, just like 'show oproc' and 'show mproc'.

Tested and effective.

## Summary by CodeRabbit

* **New Features**
  * Added "show rproc" subcommand to list Room Specials with assignable status filtering, mirroring existing special listing commands.
  * Introduced room specification editing capability in the room editor with a new dedicated menu option.
  * Room editor now displays the current spec assignment status for easy reference.


agednotdead:
"Tested and effective."
-Vasco

PR #608, opened 2025-12-16
stab offhand and chaining

kjsevertson:
Refactored Stab to cut out the clutter. The previous version had lots of flavor scenarios but they were gated behind what-ifs with long odds.

This new version functions inherently the same at the base level, with a new check against sharpness to apply a bleeding condition to the limb.  With pracs in Fighting, it starts to function differently by allowing an attempt to be made with a weapon held in the off-hand. Compounding on that, when a thief using two stabbers makes a Stab attack, it will check against the thief's dual wielding skill. On a success, a Stab attack is made for each individually.

This also allows a provision for a bonus Stab attack to be made at the end of a Backstab. This bonus Stab can also chain into an off-hand Stab.

## Summary by CodeRabbit

* **New Features**
  * Backstab can chain into an automatic follow-up stab when attacker meets advanced stabbing skill thresholds.

* **Changes**
  * Stabbing skill reassigned to the Thief Fight discipline.

* **Improvements**
  * Overhauled stab resolution with clearer hit/miss messaging and consistent poison and bleeding handling.
  * Improved dual-wield stabbing flow for more consistent, predictable attack outcomes.

### DIFF
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -115,6 +115,16 @@ static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   int dam = thief->getSkillDam(victim, SKILL_STABBING, level,
     thief->getAdvLearning(SKILL_STABBING));
   dam = thief->getActualDamage(victim, weapon, dam, SKILL_STABBING);
+  
+  if (dam <= 0) {
+    act("Your $o fails to penetrate $N's thick hide.", FALSE, thief, weapon,
+      victim, TO_CHAR);
+    act("$n's $o fails to penetrate your thick hide.", FALSE, thief, weapon,
+      victim, TO_VICT);
+    act("$n's $o fails to penetrate $N's thick hide.", FALSE, thief, weapon,
+      victim, TO_NOTVICT);
+    return TRUE;
+  }
 
   // Get target limb
   wearSlotT limb = victim->getPartHit(thief, FALSE);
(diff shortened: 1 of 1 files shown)


code/code/cmd/cmd_stab.cc
```diff
@@ -1,345 +1,239 @@
...
-        tStringOthr = format(
-                        "$n stabs $N in $S %s, ending their life in a final "
-                        "<r>spray of blood<z>!") %
-                      tStLimb;
-        break;
-    }
+// Apply poison from weapon if present
+static void stabPoisonCheck(TBeing* victim, TGenWeapon* weapon) {
+  if (weapon->isPoisoned())
+    weapon->applyPoison(victim);
```
cizra: It would be cool to move the `isPoisoned` into `applyPoison` itself (and update all call sites) - saves a dozen lines of code. Could make isPoisoned private, then.

code/code/cmd/cmd_stab.cc
```diff
@@ -1,345 +1,239 @@
...
+    act("$n's $o fails to penetrate your thick hide.", FALSE, thief, weapon,
+      victim, TO_VICT);
+    act("$n's $o fails to penetrate $N's thick hide.", FALSE, thief, weapon,
+      victim, TO_NOTVICT);
+    return TRUE;
+  }
+
+  // Get target limb
+  wearSlotT limb = victim->getPartHit(thief, FALSE);
+  if (!victim->hasPart(limb))
```
cizra: Why do other call sites of `getPartHit` not check for this? Is there a bug in other call sites? Is this check redundant?

code/code/cmd/cmd_stab.cc
```diff
@@ -1,345 +1,239 @@
...
+  // Apply damage
+  if (thief->reconcileDamage(victim, dam, SKILL_STABBING) == -1)
+    return DELETE_VICT;
+
+  // Check for bleeding and poison
+  stabBleedCheck(thief, victim, weapon, limb);
+  stabPoisonCheck(victim, weapon);
+
+  // Check weapon spec proc
+  if (weapon->checkSpec(victim, CMD_STAB, reinterpret_cast<char*>(limb),
```
cizra: nit, beyond the scope of your PR:

> `reinterpret_cast<char*>(limb)`

The traditional way of doing this was to pass `void*`, the modern is `std::variant`. `char*` is bad because it affords printing/logging/similar.

PR #612, opened 2025-12-17
Fix several issues in Docker flow

drewhershey:
Fixes three separate, small issues with the Docker flow stemming from my recent CMake conversion and Docker changes:
- Disables PCH by default, as it's incompatible with ccache, which we want to use in the CI workflows (and during dev for faster incremental builds)
    - The only downside of this is slower initial builds from a clean state, but the benefits of ccache working should be worth it, especially in the CI
- Reverts a change I made in the sneezymud Dockerfile that had it using the built-in `ubuntu` user to run the game inside the container instead of the original `sneezy` user. I did this from a "best practices" standpoint but forgot to update the volume paths in the Docker Compose file to match, which meant the volume was getting mounted to the old location still and the mutable files weren't visible to the game inside the container. Switching back to the `sneezy` user is the cleanest fix (already did so temporarily in the server to fix the issue right away), as it only requires modifying the Dockerfile in this repo and nothing in the Docker repo.
- Removes a healthcheck directive I added to the Dockerfile that connected to the game every 30 seconds then disconnected. Another "best practices" thing that ended up being more trouble than it's worth - it's generating `// Player I/O: EOF encountered on socket read.` in the game for immortals every time it runs. Nothing in the compose build is actually using this healthcheck - the monitor service uses `docker ps`. So I'm just removing it.

## Summary by CodeRabbit

* **New Features**
  * Added CMake caching pipeline to accelerate builds in CI/CD workflows.
  * Made precompiled headers optional during compilation.

* **Chores**
  * Simplified Docker image to packaging-only configuration with reduced complexity.
  * Improved logging context for socket read operations.
  * Updated GitHub Actions to newer versions for compatibility.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,8 +29,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
   libboost-regex1.83.0 \
   libboost-system1.83.0 \
   libcurl4 \
-  libmariadb3 \
-  netcat-openbsd && \
+  libmariadb3 && \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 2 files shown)


kjsevertson (approved):
Looks good, LGTM

PR #613, opened 2025-12-19
Fix oversight in Dockerfile causing incorrect volume permissions

drewhershey:
When I made the previous change to the Dockerfile to revert the user inside the container back to `sneezy` I overlooked the fact that the Ubuntu container already includes a built-in 'ubuntu' user with UID 1000. Therefore the `sneezy` user was getting created with UID other than 1000. But the monitor service was configured to set the volume permissions to UID 1000, which led to the volume being owned by 'ubuntu' user instead of 'sneezy' user, and the game not being able to read or write to the volume.

Simply renaming the built-in `ubuntu` user to `sneezy` ensures that the `sneezy` user inside the container has UID 1000, matching the volume ownership set by the monitor service.

## Summary by CodeRabbit

* **Chores**
  * Optimized Docker container build configuration to improve user and group management with enhanced file ownership and permission handling. Changes ensure proper UID/GID alignment for volumes and mounted storage, providing better consistency and reliability for containerized deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,9 +33,11 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
   apt-get clean && \
   rm -rf /var/lib/apt/lists/*
 
-# useradd -r flag creates a system account, which is preferable when running as a service
-# useradd -m flag forces creation of home directory, which -r flag prevents by default
-RUN groupadd -r sneezy && useradd -r -g sneezy -m sneezy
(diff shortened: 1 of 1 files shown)


drewhershey:
Sorry, going to merge without approval (besides CRabbit) again just because this is a simple yet urgent fix affecting the live server, as until I push this the automatic reboot from the monitor service won't work, meaning shutdowns or crashes will have to be recovered from by someone manually logging into the server.

PR #617, opened 2026-01-02
Adding line modifier to isStat checks

kjsevertson:
Previously, only isAgile allowed for the carrying of a modifier into the call. This update expands that option to all stats and also includes some affect-based modifiers within each. It also sends a message to the actor to let them know if their stat check was a success or a failure, for some transparency.

isLucky() is changed to isKarmic(num) because isLucky(num) already existed. Luckily, isLucky() was called for zero times. So, it changes nothing.

## Summary by CodeRabbit

* **Refactor**
  * Overhauled character attribute checks to parameterized variants, affecting stealth, perception, toughness, agility, charisma and related checks across combat, AI, tasks, items and traps.

* **Chores**
  * Relaxed local clangd diagnostics and added the clangd config to ignore rules to keep dev tooling preferences out of source.

### DIFF
--- a/code/code/cmd/cmd_bash.cc
+++ b/code/code/cmd/cmd_bash.cc
@@ -362,14 +362,14 @@ int TBeing::bashSuccess(TBeing* victim, spellNumT skill, bool isHoldingShield,
   // Determine which limb gets hit - same logic for all damage types
   wearSlotT limb = WEAR_BODY;
   if (!this->isTanking()) {
-    limb = WEAR_BACK; 
+    limb = WEAR_BACK;
   }
-  if (victim->isAgile(0)){
+  if (victim->isAgile(0, true)){
     limb = WEAR_ARM_R;
     if (percentChance(50)){
       limb = WEAR_ARM_L;
     }
-    
+
     if ((victim->getHeight()*2/3) > this->getHeight()) {
       limb = WEAR_LEG_R;
       if (percentChance(50)){
--- a/code/code/cmd/cmd_grapple.cc
+++ b/code/code/cmd/cmd_grapple.cc
@@ -88,7 +88,7 @@ static int grapple(TBeing* c, TBeing* victim, spellNumT skill) {
         return rc;
 
       rc = c->trySpringleap(victim);
-      
+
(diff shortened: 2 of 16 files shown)


.clangd
```diff
@@ -7,7 +7,7 @@ Diagnostics:
   ClangTidy:
     FastCheckFilter: None
   UnusedIncludes: Strict
-  MissingIncludes: Strict
+  MissingIncludes: None
```
drewhershey: Looks like you accidentally committed the changes to .clangd and .gitignore along with this stuff. If you add `.clangd` to your `~/.gitignore_global` file instead you should be able to modify it like this without the change being picked up by git. You'll probably have to undo these changes first, then change the global gitignore, then re-commit and force push to fix this unfortunately.

drewhershey (changes requested):
I'm not sure this is a good way to handle what you're trying to do here, honestly. It's conflating three different concerns (stat checking, modifier aggregation, and player feedback) which should really be separate. Having to add a "modifier" parameter to a function (the `silent` param in this case) is usually a telltale sign that code is violating the [https://en.wikipedia.org/wiki/Single-responsibility_principle](single-responsibility principle) of programming. We don't have to stick to that **100% of the time**, and I know this codebase is already rife with violations of it, but for new code it feels like we should probably at least attempt to keep it following good practices.

The main issue I have with doing it this way is that now various stat functions require knowledge of:
- The spell system (`affectedBySpell()`)
- The disease system (`hasDisease()`)
- The injury system (`eitherLegHurt(), isLimbFlags()`)
- The player messaging system

This sort of tight coupling between mechanics adds a lot of complexity and maintenance burden in the long-term.

Additionally, doing things like modifying isStrong checks by arbitrary amounts based on active diseases and hurt limbs is **in essence** just giving those mechanics pseudo-strength penalties. Why don't we just have the diseases and/or damaged limbs give **actual strength penalties**, and let `isStrong` keep its pure functionality? This would have the same overall effect, while keeping concerns much more clean and separate and easier to reason about in the future.

And I think CRabbit was probably correct about the high potential for spammy messages from these changes, or for players to accidentally get feedback when it doesn't make sense. Unless you did very thorough testing and tracing of when these functions get called it seems like it'd be really easy to have one get called in a loop accidentally and cause lots of accidental messages to be sent.

PR #618, opened 2026-01-02
altered the way taskGainExp gathers exp to prevent abuse and plevel

kjsevertson:
There were some outlier skills that were allowing for A LOT of exp gain via taskGainExp. The entire thing has been compressed to be nice and feasible for leveling up to level 20, after that, it plateaus.

Previously, butcher was basing the exp based on mob corpse level, which would let a low level pc gain a massive amount of exp from a single butchering of a high level mob.

## Release Notes

* **Refactor**
  * Rebalanced task experience rewards across gathering, crafting, and processing activities. Experience gains now dynamically scale based on activity difficulty and character progression rather than using fixed amounts. Updated activities include butchering, fishing, cooking, planting, logging, pickpocketing, skinning, and whittling.

### DIFF
--- a/code/code/disc/disc_advanced_adventuring.cc
+++ b/code/code/disc/disc_advanced_adventuring.cc
@@ -119,10 +119,11 @@ int forage(TBeing* caster, short bKnown) {
       *caster->roomp += *obj;
       foodpile /= 3;
     }
-    
-    caster->gainTaskExp(0, 50.0);
+
+    // Foraging is a simple task with moderate difficulty
+    caster->gainTaskExp(0.5);
     caster->doSave(SILENT_YES);
-    
+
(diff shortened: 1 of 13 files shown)


drewhershey (changes requested):
To be fully honest this whole system makes me a little nervous that it hasn't been tested thoroughly enough, to determine how abusable it could be. Is there anything to prevent someone from setting up a simple script to search the same secret door over and over to slowly gain exp while AFK? What about dumping a bunch of high level corpses for a low-level character to effectively power-level themselves through butcher/skinning? I haven't tested it myself - this is just based on reading the code - but my initial impression is that there are some potential oversights.

I think the concept itself is neat and could end up being a cool addition, but it seems like something we have to be **very** careful about to prevent unintended consequences.

PR #645, opened 2026-02-10
Fix Discord notification step in Build and Push GitHub action

drewhershey:
The HTTP request is being rejected because Discord automatically blocks the default user agent for Python's native HTTP request library for security reasons. Updating to a custom user agent should fix it.

Also adds a timeout to the request, just to prevent hangs if the response takes too long.

## Summary by CodeRabbit

* **Chores**
  * Enhanced Discord webhook request handling with timeout and request header configuration.

### DIFF
--- a/.github/scripts/notify_discord.py
+++ b/.github/scripts/notify_discord.py
@@ -40,6 +40,6 @@ req = urllib.request.Request(
         "content": msg,
         "allowed_mentions": {"roles": [admin_role_id]} if admin_role_id else {},
     }).encode(),
-    headers={"Content-Type": "application/json"},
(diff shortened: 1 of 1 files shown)


.github/scripts/notify_discord.py
```diff
@@ -40,6 +40,6 @@
         "content": msg,
         "allowed_mentions": {"roles": [admin_role_id]} if admin_role_id else {},
     }).encode(),
-    headers={"Content-Type": "application/json"},
+    headers={"Content-Type": "application/json", "User-Agent": "SneezyMUD-CI"},
```
wershlak: Did you check if this works? Seems like the recommendation is to pretend to be a browser.

PR #663, opened 2026-02-26
Refactor news and wiznews to directory-based files

wershlak:
## Summary

- Splits the monolithic `txt/news` and `txt/wiznews` files into individual dated files in `txt/news.d/` and `txt/wiznews.d/` (format: `YYYY-MM-DD-slug`)
- Adds `cmd_news.cc`/`cmd_news.h` and `cmd_wiznews.cc`/`cmd_wiznews.h` with modern C++20 implementations of both commands
- Both commands now support an optional search argument (e.g. `news spells`)
- `wiznews` now uses the same directory-based system as `news`
- Login news summary and MOTD "last updated" display updated for both

## Motivation

The old single-file approach made every news or wiznews post a merge conflict waiting to happen. Individual files per entry mean contributors never touch the same file.

## Impact on users

None — output format is identical to before.

## Test plan

- [x] `news` displays entries newest-first with correct date prefix and indented continuation lines
- [x] `news <term>` filters entries by search term
- [x] `wiznews` works the same way (immortals only)
- [x] Login summary shows new entries since last logon for both news and wiznews
- [x] MOTD shows correct "last updated" date for both


drewhershey (approved):
Everything looks good to me now. Just a couple of very minor things I noticed while reading through, in case you feel like cleaning them up:

- `connect.cc:2750` and `connect.cc:2382`: The `showstr_head` extraction left behind vestigial bare `{ }` blocks where the old `else` braces used to be. Harmless, just reads as unfinished refactoring.
- `connect.cc:817`: `format("%s") % line.toCRLF()` is a no-op wrapper since `sendTo(const sstring&)` exists and the string already ends with `\n\r`. Could just be `ch->sendTo(line)`.

PR #689, opened 2026-03-12
Rain brimstone overhaul and pray command bugfix

kjsevertson:
## Summary

- **Rain brimstone rework**: Significantly buffed damage (base multiplier 3.0 → 6.0), restricted to outdoor-only casting, added brimstone ash mechanic that embeds burning shards in victims on crit successes and non-saved normal hits
- **Brimstone ash effect**: New `brimstoneAsh()` function creates a sulphuric ash shard (vnum 7813) that either sticks into a random body slot or drops to the ground burning. Wisdom-scaled chance (40-62%) for the embedded shard to ignite
- **Adjusted learning curve**: Changed from START_5/LEARN_20 to START_18/LEARN_5 — requires more discipline training before becoming available, with a slower learning rate
- **Separated crit tiers**: CRIT_S_KILL/CRIT_S_TRIPLE now deal 3x damage (previously fell through to 2x), with distinct messaging per tier
- **Saving throw moved to non-crit only**: Lucky victims can reduce damage by 1/3, but only on normal (non-crit) hits. Agility and caster wisdom factor into the save difficulty
- **Damage formula**: Moved to its own case in `getSkillDam` with `HAS_SAVING_THROW * OUTDOOR_ONLY` modifiers, separating it from the harm/pillar/earthquake group
- **Pray command bugfix**: Fixed `doPray` special-case name lookup (heal/harm disambiguation) bypassing the `doesKnowSkill` check, allowing players to attempt prayers they hadn't learned

## Test plan

- [x] Cast rain brimstone indoors — should fail with "open heavens" message
- [x] Cast rain brimstone outdoors — verify damage, messaging, and ash embed
- [x] Verify crit success produces ash shard stuck in victim
- [x] Verify non-crit with lucky victim reduces damage and skips ash
- [x] Verify ash drops to ground burning when victim's slot is occupied
- [x] Attempt to pray a heal/harm spell you don't know — should say "You don't know that prayer!"
- [x] Attempt to pray a spell you do know — should work normally


code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+  }
+  wearSlotT slot = victim->getRandomPart(0, false, false);
+  if (slot == WEAR_NOWHERE) {
+    // No valid slots available, clean up and return
+    delete brimstone;
+    return 0;
+  }
+
+  if (!victim->equipment[slot] && !victim->getStuckIn(slot)) {
+    const int rc = victim->stickIn(brimstone, slot);
```
drewhershey: `stickIn` with the default `SILENT_NO` emits its own "$p sticks in your body part!" messages, then the code below sends custom "A shard of $p embeds itself..." messages. Players see both. Pass `SILENT_YES` here: `victim->stickIn(brimstone, slot, SILENT_YES)` - same pattern as `combat.cc:5865`.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+        act("The heavens <y>crack<1> and <r>HOT BRIMSTONE<1> rains down on $N!",
+          false, caster, nullptr, victim, TO_VICT);
+        ashRc = brimstoneAsh(caster, victim);
+        break;
       case CRIT_S_DOUBLE:
         CS(spell);
         dam *= 2;
         act(
-          "You hear a loud cracking sound and white hot brimstone rains down "
+          "Thunder <W>CRACKS<z> as <r>flaming hot<1> <y>brimstone<1> rain down "
```
drewhershey: "brimstone rain down" should be "brimstone rains down" - brimstone is an uncountable noun, so the verb needs third-person singular. The TO_VICT message at line 498 has it right ("rains"). Not a big deal, but easy to fix while you're in here.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
           "on $N!",
-          FALSE, caster, NULL, victim, TO_NOTVICT);
+          false, caster, nullptr, victim, TO_NOTVICT);
         act(
-          "You hear a loud cracking sound and white hot brimstone rains down "
+          "Thunder <W>CRACKS<z> as <r>flaming hot<1> <y>brimstone<1> rain down "
```
drewhershey: Same "rain" → "rains" as line 490.

code/code/misc/skill_dam.cc
```diff
@@ -309,7 +309,13 @@ int TBeing::getSkillDam(const TBeing* victim, spellNumT skill, int level,
       break;
     case SPELL_HARM:
     case SPELL_PILLAR_SALT:
+      dam = genericDam(victim, this, skill, DISC_CLERIC, level, adv_learn,
```
drewhershey: This splits HARM and PILLAR_SALT from the old grouped case, but the split also changes their effective damage multiplier from 2.25 (old: 3.0 * percModifier 0.75) to 4.0 (new: 2.0 * HAS_SAVING_THROW, no percModifier) - a 78% damage increase. Every other cleric damage spell in this function applies `percModifier()` after `genericDam`. EARTHQUAKE at line 319 still has the old treatment. Was the HARM/PILLAR_SALT damage increase intentional, or was the goal just to give them `HAS_SAVING_THROW` while keeping them at parity? If intentional, no action needed - just wanted to flag it since the PR description focuses on rain brimstone. If not, consider either keeping the same effective multiplier or calling it out as an intentional buff.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+    act("A shard of $p falls to the ground and ignites!", false, caster,
+      brimstone, nullptr, TO_CHAR);
+  }
+  return 0;
+}
+}  // namespace
+
 int rainBrimstone(TBeing* caster, TBeing* victim, int level, short bKnown,
   spellNumT spell, int adv_learn) {
+  if (!victim->outside()) {
```
drewhershey: Every other outdoor-only spell check uses `caster->outside()` - `plagueOfLocusts` at line 14 of this file, `meteorSwarm` at `disc_mage_air.cc:1173`, `tornado` at `disc_mage_air.cc:1193`. Since `TAR_CHAR_ROOM` means they're in the same room it works fine, but swapping to `caster->outside()` would match the convention. Feel free to leave it if you don't think it's worth the churn.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+        act(
+          "You hear <y>thunder<1> and <r>red hot brimstone<1> rains down on "
+          "$N!",
+          false, caster, nullptr, victim, TO_CHAR);
+        act(
+          "You hear <y>thunder<1> and <r>red hot brimstone<1> rains down on "
+          "you!",
+          false, caster, nullptr, victim, TO_VICT);
+
+        if (victim->isAgile(-(caster->getStatMod(STAT_WIS)))) {
```
drewhershey: `getStatMod(STAT_WIS)` returns 0.8-1.25, so negated and truncated to `int` this is -1 or 0 - at most 1 percentage point against the 5-95 range of the agility check. The caster's WIS has essentially no effect on the save. If the intent is for WIS to meaningfully contest AGI, the value needs scaling - something like `static_cast<int>(-(caster->getStatMod(STAT_WIS)) * 15)` would give a range of -12 to -19. If the near-zero effect is intentional, it might be worth a comment so nobody tries to "fix" it later. This isn't blocking - just flagging it in case the behavior doesn't match the intent.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
         act(
-          "You hear a loud cracking sound and then your skin starts to burn "
-          "from white hot brimstone $n has called down upon your head!",
-          FALSE, caster, NULL, victim, TO_VICT);
+          "Thunder <W>CRACKS<z> and <r>flaming hot<1> <y>brimstone<1> rains "
+          "down on $N!",
+          false, caster, nullptr, victim, TO_VICT);
+        ashRc = brimstoneAsh(caster, victim);
         break;
       case CRIT_S_NONE:
```
drewhershey: Just an FYI - the save mechanic here is substantially different from the old version in two ways: it changed from `isLucky(spellLuckModifier)` (KAR+level-based, scaled with level difference) to `isAgile` (AGI-based, level-independent), and it now only fires on non-crit hits where the old save applied to all successful casts including crits. Together these change who saves and how often. Worth noting that NPCs never get crit successes (hardcoded at `spell_parser.cc:1563`), so NPC casters always go through this CRIT_S_NONE path with the save.

drewhershey (changes requested):
Found a couple of bugs in the new `brimstoneAsh` function that should be fixed before merging - `setObjStat` wiping existing object flags, and `stickIn` producing duplicate messages. I also left some inline comments on the damage formula changes and a few smaller things.

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Good catch on the scope of the change. Reverted to `isLucky` with `spellLuckModifier` as the base, so the save is KAR+level-scaled again. The one remaining intentional change: crits bypass the save entirely (and guarantee ash). The agility/WIS interaction is now layered on top as an additional penalty rather than replacing the whole mechanic.

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Fixed — switched to `caster->outside()` to match convention.

code/code/misc/skill_dam.cc
kjsevertson: Not intentional — HARM and PILLAR_SALT are back in their original case group with EARTHQUAKE (`3.0 * percModifier`). Only RAIN_BRIMSTONE has the new formula.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+      brimstone->setObjStat(ITEM_BURNING);
+      act(format("The $p burns your %s!") % victim->describeBodySlot(slot),
+        false, caster, brimstone, victim, TO_VICT);
+      act(format("The $p burns $N's %s!") % victim->describeBodySlot(slot),
+        false, caster, brimstone, victim, TO_NOTVICT);
+      act(format("The $p burns $N's %s!") % victim->describeBodySlot(slot),
+        false, caster, brimstone, victim, TO_CHAR);
+    }
+  } else {
+    brimstone->setObjStat(ITEM_BURNING);
```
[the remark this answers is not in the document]
kjsevertson: Fixed alongside line 438.

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Reworked — reverted to `isLucky(spellLuckModifier)` as the base save, with a WIS-scaled agility penalty: if the victim fails `isAgile(0)`, the luck mod increases by `100 * caster WIS modifier` (~6–9% harder to save). This restores the level-scaled KAR-based save while giving wise casters an edge against non-agile targets.

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Fixed — both TO_NOTVICT and TO_CHAR messages corrected to "rains".

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Fixed — now passing `SILENT_YES` to match the pattern in `combat.cc`.

code/code/disc/disc_cleric_wrath.cc
kjsevertson: Fixed alongside line 490.

code/code/disc/disc_cleric_wrath.cc
```diff
@@ -403,96 +403,172 @@ int pillarOfSalt(TBeing* caster, TBeing* victim) {
...
+
+    act(format("A shard of $p embeds itself in $N's %s!") %
+          victim->describeBodySlot(slot),
+      false, caster, brimstone, victim, TO_NOTVICT);
+    act(format("A shard of $p embeds itself in your %s!") %
+          victim->describeBodySlot(slot),
+      false, caster, brimstone, victim, TO_VICT);
+
+    if (percentChance(caster->getStatMod(STAT_WIS) * 50)) {
+      brimstone->setObjStat(ITEM_BURNING);
```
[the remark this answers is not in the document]
kjsevertson: Fixed — both instances changed to `addObjStat` to preserve existing flags.

code/code/misc/skill_dam.cc
```diff
@@ -308,9 +308,12 @@ int TBeing::getSkillDam(const TBeing* victim, spellNumT skill, int level,
       dam = genericDam(victim, this, skill, DISC_MAGE, level, adv_learn,
         2.0 * HAS_SAVING_THROW * OUTDOOR_ONLY, REDUCE_YES, !isPc(), TRIM_NO);
       break;
+    case SPELL_RAIN_BRIMSTONE:
+      dam = genericDam(victim, this, skill, DISC_CLERIC, level, adv_learn,
+        2.0 * HAS_SAVING_THROW * OUTDOOR_ONLY, REDUCE_YES, !isPc(), TRIM_NO);
+      break;
```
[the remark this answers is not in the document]
kjsevertson: this is fine. the factions are a dead system

drewhershey (approved):

PR #696, opened 2026-03-17
Fix room spec proc crashes and modernize roomSpecials array

drewhershey:
## Summary

- Fix use-after-free in `blazingroom` and `BankVault` caused by iterator
  invalidation when deleting beings mid-iteration. Collect dead beings in a
  vector and delete after iteration.
- Fix `reconcileDamage` death check: was using `IS_SET_DELETE` on a return
  value of -1 (which worked by accident due to all bits being set). Now
  checks `== -1` directly.
- Refactor `blazingroom` and `BankVault` into a shared data-driven
  `damageRoomOccupants()` helper with `DamageEffect` configs.
- Modernize `roomSpecials` from a C-style runtime array with positional
  indexing and bogus placeholder entries into an `inline constexpr std::array`
  with explicit IDs and compile-time lookup functions. Update all call sites.
- Simplify `change_room_spec`: replace VT100 cursor-positioned grid layout
  with sequential text output.
- Add `static_assert` for unique IDs in the room specials array.
- Update Clang compat warnings from C++20 to C++23.

## Test plan

- [x] C++ unit tests pass (`make test`)
- [x] Builds clean with zero warnings (Clang 20, dev-clang preset)
- [x] Functional test added for `show rproc` command
- [x] `show rproc` and `show rproc assignable` verified against live server

## Summary by CodeRabbit

* **New Features**
  * Added "assignable" filter to the room specials listing command.

* **Improvements**
  * Room specials list now includes numeric IDs alongside names.
  * Room-edit workflow gains stronger validation, permission checks, and clearer prompts.
  * Room statistics show special-procedure info in a single-line, cleaner format.

* **Tests**
  * Added end-to-end tests covering room specials listing and the assignable filter behavior.

### DIFF
--- a/code/code/spec/spec_rooms.cc
+++ b/code/code/spec/spec_rooms.cc
@@ -1427,12 +1427,20 @@ namespace {
 
   void damageRoomOccupants(TRoom& room,
     std::initializer_list<DamageEffect> effects) {
-    std::vector<TBeing*> toDelete;
+    std::vector<TBeing*> targets;
 
     for (TThing* thing : room.stuff) {
-      auto* tb = dynamic_cast<TBeing*>(thing);
+      if (auto* tb = dynamic_cast<TBeing*>(thing); tb && tb->isPc()) {
+        targets.push_back(tb);
+      }
+    }
 
-      if (!tb || !tb->isPc()) {
+    std::vector<TBeing*> toDelete;
+
+    for (auto* tb : targets) {
+      // Re-validate: earlier damage may have moved this PC out of the room
+      // (e.g. linkdead PCs get moved to Room::STORAGE during makeCorpse)
+      if (tb->roomp != &room) {
         continue;
       }
 


code/code/spec/spec_rooms.cc
```diff
@@ -1368,121 +1368,166 @@ int monkQuestProcFall(TBeing* ch, cmdTypeT cmd, const char*, TRoom* rp) {
...
-        "<r>It burns your skin and you choke uncontrollably!<1>\n\r");
+      auto sendToChar = [tb](const char* msg) {
+        colorAct(COLOR_ROOMS, msg, false, tb, nullptr, nullptr, TO_CHAR);
+      };
 
-      //  vlogf(LOG_PEEL, format("Bank: %s caught in vault") %  tb->getName());
+      for (const auto& effect : effects) {
+        if (effect.onTrigger) {
+          sendToChar(effect.onTrigger);
+        }
```
wershlak: The deferred-delete fixes the common case nicely, but `reconcileDamage` -> `rawKill` -> `makeCorpse` (makecorpse.cc:51-60) does `--(*this)` to move linkdead PCs to `Room::STORAGE` during the loop, invalidating the range-for iterator. This is extremely rare (linkdead PC in a trap room who dies) and a huge improvement over the old code which crashed on every death.

Collecting all `TBeing*` pointers into a vector first, then iterating the vector, would eliminate this remaining edge case.

PR #701, opened 2026-03-21
Refactor body-hits-ground chain; add knockOffMount and getRideMod

kjsevertson:
## Summary

Builds on the original `TBeing::stumble()` work to refactor the entire "character hits the ground" chain (stumble → crashLanding → springleap), introduce a new hostile-dislodge path (`knockOffMount`) for mounted victims, and centralize riding modifiers (`getRideMod`) so every ride check pulls from the same comprehensive state model.

### `crashLanding` — universal landing primitive
- Signature: `crashLanding(positionTypeT, bool force, bool dam, bool falling)` → `crashLanding(int severity = 0, bool falling = false)`.
- Internal `isAgile(-severity)` roll picks SITTING (pass) vs RESTING (fail). Position is **downgrade-only** — never promotes.
- Catfall, the force branch, and direct fall damage are gone. `checkFalling` still handles fall-sector lethality; the sit/sprawl flavor is gated on actual position change *and* suppressed in fall sectors (where checkFalling provides its own narrative).
- Tail-calls `springleap()` for monks so they may bounce back to standing as part of the chain.

### `springleap` — monk-only stand-up
- Drops target, drops the entire combat path (specialAttack, getActualDamage, kick flavor, reconcileDamage, lag).
- `bSuccess(SKILL_SPRINGLEAP)` gates whether the caster actually stands up; failure leaves them on the ground.
- Position guard: `RESTING <= getPosition() <= SITTING`. Cannot promote unconscious characters.

### `stumble`, `trySpringleap`, `doSpringleap`
- `stumble()` drops `opponent` parameter and the post-call `trySpringleap` (now internal to `crashLanding`).
- `trySpringleap` deleted entirely.
- `doSpringleap` becomes no-arg; `monkMove` calls `springleap()` directly. Player command keeps its "You're not in position for that!" message.

### `canClimb` — sprawled drops
- Gated at `< POSITION_SITTING`. Sitting characters can still hold a branch; sprawled characters can't. A bashed mounted-tree victim now actually falls via `checkFalling`.

### `knockOffMount(int severity = 0)` — new hostile dislodge
- Internal `rideCheck(-severity)` save: pass = "hangs on tight" flavor, early return; fail = "thrown clear" flavor + `dismount` + `crashLanding` chain.
- Stays-aloft branch: if mount was flying and rider can fly themselves (or room is flying-sector), dismount to POSITION_FLYING instead of crashing.

### `getRideMod()` — new helper
- Sums situational modifiers for any riding skill check: saddle bonus + advanced-riding amplifier, secondary-rider penalty, deikhan class bonus, combat state (tanking / engager / fighting / mount-fighting with deikhan offsets), mount-type-specific skill, stat reactions (agi / dex / bra, doubled when positive by advanced riding), drunk / hunger / thirst, leg / arm injuries, exhaustion.
- `bSuccess(SKILL_ADVANCED_RIDING)` cached once per call so it doesn't grind 4× per check.

### `rideCheck` — formula change
- New: `learn = getSkillValue(SKILL_RIDE) + getRideMod() + mod`.
- The old 3× amplification on caller mod is removed; `getRideMod` now provides the comprehensive baseline.
- Doubled rideCheck patterns ("force two failures") consolidated to single checks.

### Skill migrations
- **bash, bodyslam, spin, grapple** — success path now uses `wasMounted ? knockOffMount() : crashLanding()`. Each preserves skill-specific takedown flavor in both branches. Grapple is special-cased: if the rider hangs on, grapple aborts cleanly with "slips" flavor and a 2-round lag.
- **combat.cc canFight** (mounted attacker entering combat, secondary riders when their mount enters combat) — replaced `rideCheck → fallOffMount(POSITION_SITTING)` with `knockOffMount()`.
- **damage.cc reconcileDamage** (rider takes damage, secondary riders when their mount takes damage) — replaced with `knockOffMount(3)`. `addToWait` only fires if the rider was actually thrown.

## Gameplay changes

| Aspect | Before | After |
|---|---|---|
| **Skill miss recovery** | Always fall to sitting | Agile: stay standing. Non-agile: fall (sitting or resting via internal roll). |
| **Flying + fall sector miss** | Mixed | Unified "stay aloft" |
| **Hit the ground** | Took 2-8 fall damage | No automatic damage (severity-based future tuning) |
| **Mounted character hit by combat skill** | Always knocked off (or just printed flavor) | Riding skill check: skilled riders / deikhans / saddled hold their seat; impaired riders fall |
| **Rider fall in vert sector while sprawled** | Stayed on the branch | Falls via checkFalling |
| **Knocked off a flying mount with own wings** | Not handled | Catches yourself, stays aloft |
| **Monk recovery** | Auto-springleap only after specific skills | Auto-springleap after any "hit the ground" event (stumble, crashLanding, knockOffMount fall) |

### Tuning notes

- Severity values are placeholders (mostly `0` everywhere, `3` for damage events). Future work will tie severity to skill damage so each caller passes a meaningful magnitude.
- `getRideMod` produces a much wider dynamic range than the old simple modifiers — skilled riders are noticeably harder to dislodge, impaired riders much easier. Intentional model shift.

## DELETE flag safety

All migrated sites propagate `DELETE_THIS` correctly. `knockOffMount` returns whatever `crashLanding` returns, so DELETE_THIS flows through to callers. Iteration loops in `combat.cc canFight` and `damage.cc reconcileDamage` are safe — single-iteration on throw, or `t2 = t->nextRider` cached before the call.

Two pre-existing latent UAFs surfaced (handler.cc affectChange + disc_thief_stealth.cc subterfugeMiss both discard `crashLanding`'s return). Neither was introduced or worsened by this branch — they predate it. Documented for follow-up PRs.

## Test plan

- [x] Trip / bash / bodyslam / spin / grapple / headbutt — verify variable SITTING/RESTING outcome
- [x] Sleeping / stunned / incapacitated victim — verify no position promotion
- [x] Monk gets bashed — verify auto-springleap fires after crashLanding
- [x] Non-monk gets bashed — stays on ground (no springleap)
- [x] `springleap` while standing — "not in position" message; while sitting (monk) — springs up or fails
- [x] Bash/bodyslam/spin a mounted victim — skill flavor + knockOffMount chain (rideCheck save → hang-on or throw)
- [x] Grapple a mounted victim that hangs on — "grapple slips" flavor, no pin
- [x] Vertical-sector strong man bashed — falls via checkFalling (canClimb=false at RESTING)
- [x] Knocked off flying mount with own wings — catches in air, stays at POSITION_FLYING
- [x] Dispel fly affect — crashLanding fires, sprawls (handler.cc:198 path)
- [x] Build clean, formatted, tests pass

### DIFF
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -112,11 +112,11 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   } else {
-    act("$n tries to bodyslam $N but can't get a grip.", false, this, 0, victim,
+    act("$n tries to bodyslam $N but loses $s footing.", false, this, 0, victim,
       TO_NOTVICT);
-    act("You try to bodyslam $N but can't get a grip.", false, this, 0, victim,
+    act("You try to bodyslam $N but lose your footing.", false, this, 0, victim,
       TO_CHAR);
-    act("$n tries to bodyslam you but can't get a grip.", false, this, 0,
+    act("$n tries to bodyslam you but loses $s footing.", false, this, 0,
       victim, TO_VICT);
 
     int rc = stumble(victim);
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -133,22 +133,27 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
 int TBeing::bodyslamHit(TBeing* victim) {
   const bool wasMounted = (victim->riding != nullptr);
 
+  // Setup: the lift attempt. Then a payoff line that varies by mount status.
+  // crashLanding/knockOffMount narrates the impact result afterward.
+  act("You grab $N around the middle, attempting to lift $M overhead!", false,
+    this, nullptr, victim, TO_CHAR);
+  act("$n grabs you around the middle, attempting to lift you overhead!", false,
+    this, nullptr, victim, TO_VICT, ANSI_RED);
+  act("$n grabs $N around the middle, attempting to lift $M overhead!", false,
+    this, nullptr, victim, TO_NOTVICT);
+
   if (wasMounted) {
-    act("You lift $N off $p and slam $M to the $g!", false, this,
-      victim->riding, victim, TO_CHAR);
-    act("$n lifts $N off $p and slams $M to the $g!", false, this,
-      victim->riding, victim, TO_NOTVICT);
-    act("$n lifts you off $p and slams you to the $g!", false, this,
-      victim->riding, victim, TO_VICT);
-  } else {
-    act("$n lifts $N over $s head and slams $M to the $g.", FALSE, this, 0,
-      victim, TO_NOTVICT);
-    act("You lift $N over your head and slam $M to the $g.", FALSE, this, 0,
+    act("You pull $N from $p, throwing $M down!", false, this, victim->riding,
       victim, TO_CHAR);
-    act("You get a great view as $n lifts you over $s head.", FALSE, this, 0,
-      victim, TO_VICT);
-    act("Suddenly, the $g rushes upward and knocks the wind out of you!", FALSE,
-      this, 0, victim, TO_VICT, ANSI_RED);
+    act("$n pulls you from $p, throwing you down!", false, this, victim->riding,
+      victim, TO_VICT, ANSI_RED);
+    act("$n pulls $N from $p, throwing $M down!", false, this, victim->riding,
+      victim, TO_NOTVICT);
+  } else {
+    act("You throw $N down hard!", false, this, nullptr, victim, TO_CHAR);
+    act("$n throws you down hard!", false, this, nullptr, victim, TO_VICT,
+      ANSI_RED);
+    act("$n throws $N down hard!", false, this, nullptr, victim, TO_NOTVICT);
   }
 
   int rc = wasMounted ? victim->knockOffMount() : victim->crashLanding();
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -82,8 +82,6 @@ bool TBeing::canBodyslam(TBeing* victim, silentTypeT silent) {
 }
 
 int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
-  int rc;
-
   if (type == TYPE_DEX) {
     act("$N deftly avoids your bodyslam attempt.", FALSE, this, 0, victim,
       TO_CHAR);
@@ -99,45 +97,29 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     act("$N deftly counters $n's bodyslam attempt, and heaves $m to the $g.",
       FALSE, this, 0, victim, TO_NOTVICT);
 
-    rc = crashLanding(POSITION_SITTING);
+    int rc = stumble(victim);
     if (IS_SET_DELETE(rc, DELETE_THIS))
-      return rc;
-
-    rc = trySpringleap(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+      return DELETE_THIS;
+    if (IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   } else if (type == TYPE_STR) {
-    act("$n collapses as $e fails to pick $N up.", FALSE, this, 0, victim,
+    act("$n tries to bodyslam $N but fails to lift $M.", false, this, 0, victim,
       TO_NOTVICT);
-    act("Your strength gives out as you try to pick $N up for bodyslamming.",
-      FALSE, this, 0, victim, TO_CHAR);
-    act("$n's strength gives out as $e tries to pick you up for bodyslamming.",
-      FALSE, this, 0, victim, TO_VICT);
-
-    rc = crashLanding(POSITION_SITTING);
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return rc;
-
-    sendTo(format("%sYou fall to the %s.%s\n\r") % blue() %
-           roomp->describeGround() % norm());
-
-    rc = trySpringleap(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
-      return rc;
-  } else {
-    act("$n tries to bodyslam $N, but ends up falling down.", FALSE, this, 0,
-      victim, TO_NOTVICT);
-    act("You try to bodyslam $N, but end up falling on your face.", FALSE, this,
-      0, victim, TO_CHAR);
-    act("$n fails to bodyslam you, and tumbles to the $g.", FALSE, this, 0,
+    act("You try to bodyslam $N but fail to lift $M.", false, this, 0, victim,
+      TO_CHAR);
+    act("$n tries to bodyslam you but fails to lift you.", false, this, 0,
       victim, TO_VICT);
 
-    rc = crashLanding(POSITION_SITTING);
+    int rc = stumble(victim);
     if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_THIS;
+    if (IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
-
-    rc = trySpringleap(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+  } else {
+    int rc = stumble(victim);
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_THIS;
+    if (IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   }
 
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -90,12 +90,12 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     act("$N deftly avoids $n's bodyslam attempt.", FALSE, this, 0, victim,
       TO_NOTVICT);
   } else if (type == TYPE_MONK) {
-    act("$N deftly counters your attempt, throwing you to the $g.", FALSE, this,
-      0, victim, TO_CHAR, ANSI_RED);
-    act("You deftly counter $n's bodyslam attempt, and throw $m to the $g.",
-      FALSE, this, 0, victim, TO_VICT);
-    act("$N deftly counters $n's bodyslam attempt, and heaves $m to the $g.",
-      FALSE, this, 0, victim, TO_NOTVICT);
+    act("$N deftly counters your bodyslam, and throws you to the side.", false,
+      this, 0, victim, TO_CHAR, ANSI_RED);
+    act("You deftly counter $n's bodyslam, and throw $m to the side.", false,
+      this, 0, victim, TO_VICT);
+    act("$N deftly counters $n's bodyslam, and throws $m to the side.", false,
+      this, 0, victim, TO_NOTVICT);
 
     int rc = stumble(victim);
     if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -142,23 +142,32 @@ int TBeing::bodyslamHit(TBeing* victim) {
   act("$n grabs $N around the middle, attempting to lift $M overhead!", false,
     this, nullptr, victim, TO_NOTVICT);
 
+  int rc;
   if (wasMounted) {
-    act("You pull $N from $p, throwing $M down!", false, this, victim->riding,
-      victim, TO_CHAR);
-    act("$n pulls you from $p, throwing you down!", false, this, victim->riding,
-      victim, TO_VICT, ANSI_RED);
-    act("$n pulls $N from $p, throwing $M down!", false, this, victim->riding,
-      victim, TO_NOTVICT);
+    TThing* mount =
+      victim->riding;  // capture before knockOffMount may dismount
+    rc = victim->knockOffMount(getSkillValue(SKILL_BODYSLAM) / 2);
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_VICT;
+    // Only narrate the throw when the rider was actually dismounted; if they
+    // hung on, knockOffMount printed its own "hangs on tight" flavor.
+    if (!victim->riding) {
+      act("You pull $N from $p, throwing $M down!", false, this, mount, victim,
+        TO_CHAR);
+      act("$n pulls you from $p, throwing you down!", false, this, mount,
+        victim, TO_VICT, ANSI_RED);
+      act("$n pulls $N from $p, throwing $M down!", false, this, mount, victim,
+        TO_NOTVICT);
+    }
   } else {
     act("You throw $N down hard!", false, this, nullptr, victim, TO_CHAR);
     act("$n throws you down hard!", false, this, nullptr, victim, TO_VICT,
       ANSI_RED);
     act("$n throws $N down hard!", false, this, nullptr, victim, TO_NOTVICT);
+    rc = victim->crashLanding();
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_VICT;
   }
-
-  int rc = wasMounted ? victim->knockOffMount() : victim->crashLanding();
-  if (IS_SET_DELETE(rc, DELETE_THIS))
-    return DELETE_VICT;
 
   // see the balance notes for details on what's going on here.
   float wt = combatRound(discArray[SKILL_BODYSLAM]->lag);
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -98,9 +98,7 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
       FALSE, this, 0, victim, TO_NOTVICT);
 
     int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return DELETE_THIS;
-    if (IS_SET_DELETE(rc, DELETE_VICT))
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   } else if (type == TYPE_STR) {
     act("$n tries to bodyslam $N but fails to lift $M.", false, this, 0, victim,
@@ -111,15 +109,18 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
       victim, TO_VICT);
 
     int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return DELETE_THIS;
-    if (IS_SET_DELETE(rc, DELETE_VICT))
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   } else {
+    act("$n tries to bodyslam $N but can't get a grip.", false, this, 0, victim,
+      TO_NOTVICT);
+    act("You try to bodyslam $N but can't get a grip.", false, this, 0, victim,
+      TO_CHAR);
+    act("$n tries to bodyslam you but can't get a grip.", false, this, 0,
+      victim, TO_VICT);
+
     int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return DELETE_THIS;
-    if (IS_SET_DELETE(rc, DELETE_VICT))
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
       return rc;
   }
 
--- a/code/code/cmd/cmd_bodyslam.cc
+++ b/code/code/cmd/cmd_bodyslam.cc
@@ -97,8 +97,8 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     act("$N deftly counters $n's bodyslam, and throws $m to the side.", false,
       this, 0, victim, TO_NOTVICT);
 
-    int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+    int rc = stumble();
+    if (IS_SET_DELETE(rc, DELETE_THIS))
       return rc;
   } else if (type == TYPE_STR) {
     act("$n tries to bodyslam $N but fails to lift $M.", false, this, 0, victim,
@@ -108,8 +108,8 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     act("$n tries to bodyslam you but fails to lift you.", false, this, 0,
       victim, TO_VICT);
 
-    int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+    int rc = stumble();
+    if (IS_SET_DELETE(rc, DELETE_THIS))
       return rc;
   } else {
     act("$n tries to bodyslam $N but loses $s footing.", false, this, 0, victim,
@@ -119,8 +119,8 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
     act("$n tries to bodyslam you but loses $s footing.", false, this, 0,
       victim, TO_VICT);
 
-    int rc = stumble(victim);
-    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+    int rc = stumble();
+    if (IS_SET_DELETE(rc, DELETE_THIS))
       return rc;
   }
 
@@ -131,9 +131,16 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
 }
 
 int TBeing::bodyslamHit(TBeing* victim) {
-  int rc;
+  const bool wasMounted = (victim->riding != nullptr);
 
-  if (!victim->riding) {
+  if (wasMounted) {
+    act("You lift $N off $p and slam $M to the $g!", false, this,
+      victim->riding, victim, TO_CHAR);
+    act("$n lifts $N off $p and slams $M to the $g!", false, this,
+      victim->riding, victim, TO_NOTVICT);
+    act("$n lifts you off $p and slams you to the $g!", false, this,
+      victim->riding, victim, TO_VICT);
+  } else {
     act("$n lifts $N over $s head and slams $M to the $g.", FALSE, this, 0,
       victim, TO_NOTVICT);
     act("You lift $N over your head and slam $M to the $g.", FALSE, this, 0,
@@ -142,29 +149,11 @@ int TBeing::bodyslamHit(TBeing* victim) {
       victim, TO_VICT);
     act("Suddenly, the $g rushes upward and knocks the wind out of you!", FALSE,
       this, 0, victim, TO_VICT, ANSI_RED);
-  } else {
-    act("$n lifts $N off $S $o and slams $M to the $g.", FALSE, this,
-      victim->riding, victim, TO_NOTVICT);
-    act("You lift $N off $S $o and slam $M to the $g.", FALSE, this,
-      victim->riding, victim, TO_CHAR);
-    act("You get a great view as $n lifts you off your $o over $s head.", FALSE,
-      this, victim->riding, victim, TO_VICT);
-    act("Suddenly, the $g rushes upward and knocks the wind out of you!", FALSE,
-      this, victim->riding, victim, TO_VICT, ANSI_RED);
-    victim->dismount(POSITION_RESTING);
   }
 
-  rc = victim->crashLanding(POSITION_SITTING);
+  int rc = wasMounted ? victim->knockOffMount() : victim->crashLanding();
   if (IS_SET_DELETE(rc, DELETE_THIS))
     return DELETE_VICT;
-
-  rc = victim->trySpringleap(this);
-  if (IS_SET_DELETE(rc, DELETE_THIS) && IS_SET_DELETE(rc, DELETE_VICT))
-    return rc;
-  else if (IS_SET_DELETE(rc, DELETE_THIS))
-    return DELETE_VICT;
-  else if (IS_SET_DELETE(rc, DELETE_VICT))
-    return DELETE_THIS;
 
   // see the balance notes for details on what's going on here.
   float wt = combatRound(discArray[SKILL_BODYSLAM]->lag);
--- a/code/code/cmd/cmd_doorbash.cc
+++ b/code/code/cmd/cmd_doorbash.cc
@@ -25,7 +25,20 @@ int TBeing::slamIntoWall(roomDirData* exitp) {
   sendTo("OUCH!  That REALLY Hurt!\n\r");
   sprintf(buf, "$n crashes against the %s with no effect.\n\r", doorname);
   act(buf, FALSE, this, 0, 0, TO_ROOM);
-  if (reconcileDamage(this, (::number(1, 10) * 2), DAMAGE_COLLISION) == -1)
+  TBeing* fightTarget = fight();
+  int rc = stumble(fightTarget);
+  if (IS_SET_DELETE(rc, DELETE_THIS))
+    return DELETE_THIS;
+  if (IS_SET_DELETE(rc, DELETE_VICT)) {
+    delete fightTarget;
+    fightTarget = nullptr;
+  }
+
+  int dam = ::number(1, 10) * 2;
+  if (isBrawny())
+    dam /= 2;
+  addToWait(combatRound(dam));
+  if (reconcileDamage(this, dam, DAMAGE_COLLISION) == -1)
     return DELETE_THIS;
 
   if (!isImmortal()) {
@@ -36,15 +49,6 @@ int TBeing::slamIntoWall(roomDirData* exitp) {
     aff.bitvector = AFF_STUNNED;
     affectTo(&aff, -1);
   }
-
-#if 0
-  addToWait(combatRound(12));
-#else
-  int rc;
-  rc = crashLanding(POSITION_RESTING);
-  if (IS_SET_DELETE(rc, DELETE_THIS))
-    return DELETE_THIS;
-#endif
 
   return TRUE;
 }
--- a/code/code/cmd/cmd_doorbash.cc
+++ b/code/code/cmd/cmd_doorbash.cc
@@ -25,15 +25,7 @@ int TBeing::slamIntoWall(roomDirData* exitp) {
   sendTo("OUCH!  That REALLY Hurt!\n\r");
   sprintf(buf, "$n crashes against the %s with no effect.\n\r", doorname);
   act(buf, FALSE, this, 0, 0, TO_ROOM);
-  int rc = crashLanding(POSITION_SITTING);
-  if (IS_SET_DELETE(rc, DELETE_THIS))
-    return DELETE_THIS;
-
-  int dam = ::number(1, 10) * 2;
-  if (isBrawny())
-    dam /= 2;
-  addSkillLag(SKILL_DOORBASH, 0);
-  if (reconcileDamage(this, dam, DAMAGE_COLLISION) == -1)
+  if (reconcileDamage(this, (::number(1, 10) * 2), DAMAGE_COLLISION) == -1)
     return DELETE_THIS;
 
   if (!isImmortal()) {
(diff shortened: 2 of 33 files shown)


code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,49 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
+      int rc = trySpringleap(springTarget);
+      if (IS_SET_DELETE(rc, DELETE_THIS))
+        return DELETE_THIS;
+      if (IS_SET_DELETE(rc, DELETE_VICT))
+        victimDeleted = true;
+      if (rc)
+        stumbleDam /= 2;
+    }
+
+    if (reconcileDamage(this, stumbleDam, DAMAGE_FALL) == -1)
```
wershlak: This makes 1d4 fall damage routine on every miss for non-agile characters, where pre-PR `crashLanding(POSITION_SITTING)` only damaged the caster on flying / fall-sector paths (`movement.cc:3465-3469, 3483-3487`). The damage path is fine in isolation, but it exposes a latent UAF in every caller that runs `addSkillLag(skill, rc)` on `this` after the miss function returns:

- `cmd_bash.cc:460-461` (`doBash`)
- `cmd_bodyslam.cc:296-298` (`doBodyslam`)
- `cmd_spin.cc:323-325` (`doSpin`)
- `cmd_headbutt.cc:214-216` (`doHeadbutt`)
- `disc_monk_leverage.cc:350-354` (`doShoulderThrow` - also `cantHit += loseRound(1)` on the next line)
- `disc_monk_leverage.cc:170-173` (`aiHurl`)
- `disc_monk_leverage.cc:314-317` (`hurl`)
- `disc_monk_leverage.cc:796-799` (`defenestrate`)

Each of those does `if (rc) addSkillLag(skill, rc);` before checking DELETE_THIS, and `addSkillLag` is virtual so it dereferences `this` through the vtable. A low-HP attacker dying to the new fall damage on a missed bash crashes the server.

This was a pre-existing latent bug for the rare crashLanding fall-sector case, but the PR turns it into a routinely reachable crash for any low-HP melee/monk character.

Two ways out:

1. Make stumble's caster damage opt-in (default off; only callers known to handle DELETE_THIS opt in). Keeps the PR's blast radius small.
2. Fix the wrappers to check `IS_SET_DELETE(rc, DELETE_THIS)` before `addSkillLag` (and `cantHit += loseRound(1)` in `doShoulderThrow`). More work but cleans up the whole class of bugs.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,49 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
   return TRUE;
 }
 
+int TBeing::stumble(TBeing* victim) {
+  if (!isAgile(0)) {
+    setPosition(POSITION_SITTING);
```
wershlak: Feel free to handle in a follow-up if you'd rather not expand scope here: the non-agile branch unconditionally drops the character to `POSITION_SITTING` without checking `isFlying()` / `roomp->isFallSector()`, while the agile branch below carefully handles both. `canBash` permits flying-vs-flying combat (`cmd_bash.cc:93`), so a flying non-agile character missing a bash will now drop to sitting mid-air with no flying acknowledgement, take fall damage with no fall-sector check, then try to springleap from the floor. Pre-PR `crashLanding(POSITION_SITTING)` handled this case (`movement.cc:3471-3487`). Mirroring the agile branch's flying / fall-sector structure here would close the gap.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,49 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
+    sendTo("You stumble and fall on your butt!\n\r");
+    act("$n stumbles and falls on $s butt!", true, this, nullptr, nullptr, TO_ROOM);
+    addToWait(combatRound(1));
```
wershlak: Non-blocking design note: the codebase's lag idiom elsewhere is `caller invokes addSkillLag(skill, rc)`, which goes through `discArray[skill]->lag` and `lagAdjust(skill)` (`discipline.cc:3123-3139`). The hardcoded `combatRound(1)` here means a bash miss and a defenestrate miss incur identical fall lag where the per-skill discipline table would tune them differently, and `lagAdjust(skill)` is bypassed entirely. Either pass `spellNumT skill` into stumble and call `addSkillLag` internally, or return the lag amount and let the caller add it via the existing path. Same applies to the agile branch's `addToWait(combatRound(1) / 2)` below.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,49 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
+
+    int stumbleDam = dice(1, 4);
+
+    bool victimDeleted = false;
+    TBeing* springTarget = victim ? victim : fight();
```
wershlak: Minor cleanup opportunity - feel free to skip: the `nullptr` default on the parameter (in `being.h`) plus the `fight()` fallback here aren't exercised by any of the 11 call sites (all pass an explicit non-null victim), and `trySpringleap` already falls back to `fight()` itself (`disc_monk_leverage.cc:809`). Dropping the default and the fallback would simplify the API for zero current loss.

code/code/disc/disc_monk_leverage.cc
```diff
@@ -171,8 +169,8 @@ int TBeing::aiHurl(dirTypeT dr, TBeing* victim) {
     if (rc) {
       addSkillLag(SKILL_HURL, rc);
     }
-    if (IS_SET_DELETE(rc, DELETE_THIS))
-      return DELETE_THIS;
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
```
wershlak, line 168: Style consistency nit, optional: the miss path here now uses `if (DELETE_THIS || DELETE_VICT) return rc;` but the hit path immediately above (lines 162-168) still uses the older `if (DELETE_VICT) return DELETE_VICT; if (DELETE_THIS) return DELETE_THIS;` pattern. Same situation in `hurl` (lines 295-321), `shoulderThrow` (lines 519-528), and `defenestrate` (lines 779-803) - the PR converted only the miss branch in each, leaving the hit branch on the old idiom. Both work, but the divergent return idioms within a single if/else block force the reader to context-switch verifying each. Normalizing would be a quick pass.

wershlak (changes requested):
Found one issue I think needs to be addressed before merging - see the inline comment on `stumble()`. The new routine fall damage on misses runs into a class of latent UAF bugs in the command wrappers that the previous `crashLanding` path almost never triggered. A few non-blocking design notes are also inline.

A few FYI items below that aren't requests for changes, just things you might find interesting:

- `tripFail` (`cmd_trip.cc:212-220`) and three `cmd_grapple.cc` branches (`86-90`, `121-125`, `194-198`) are textbook matches for the same `crashLanding(POSITION_SITTING)` + `trySpringleap` pattern that the PR replaces elsewhere. Leaving them unmigrated means future stumble mechanics tweaks (agility check, fall damage, lag scaling) silently won't apply there. Probably worth migrating in a Phase 2 PR (or in this one if scope allows) and noting in the description either way.
- The `IS_SET_ONLY` -> `IS_SET_DELETE` change at `cmd_headbutt.cc:218` is purely a naming cleanup. Both macros are defined identically in `structs.h:373` and `:378` (both compute `(a & b) == b`). Good change for convention, but not the behavior fix the commit message describes.
- When `trySpringleap` succeeds, `springleap()` itself does `caster->setPosition(POSITION_STANDING); caster->updatePos();`. So a successful springleap-from-stumble actually ends with the attacker standing, not sitting from the stumble. Probably fine, just wasn't called out in the audit table.
- The bodyslam generic-fail branch lost its old "ends up falling down" flavor for a "can't get a grip" theme. The new flavor doesn't really pair with the stumble-and-fall outcome that immediately follows - the character can't grip, then falls down anyway.
- `cmd_headbutt.cc:58` still has the `// I don't understand this logic` comment - the logic is actually fine (`$n` is attacker, `v` is victim, so the victim sees themselves dodging). Worth deleting or rewording while the file is open.
- The PR description still includes the doorbash rework, springleap-from-doorbash gameplay row, and `slamIntoWall` audit row, but `cmd_doorbash.cc` isn't in the diff (still has the old `crashLanding(POSITION_RESTING)` at line 44). Worth trimming those sections so future readers / `git log` archaeologists aren't misled about what merged.

code/code/misc/movement.cc
kjsevertson: Addressed in `e3309583e` — chose option 1: removed caster damage entirely from `stumble()`. With no DELETE_THIS path through this code, the latent UAF in the eight wrappers (`addSkillLag` on `this` after the miss returns) is no longer reachable from the unified miss path. The pre-existing issue for the rare `crashLanding` flying/fall-sector case in those wrappers remains, but is out of scope here.

code/code/misc/movement.cc
kjsevertson: Addressed in `e3309583e` and refined in `b44812cca` — `if (isFlying() && roomp->isFallSector())` is now an early return at the top of `stumble()`, emitting the unified "stay aloft" message before either agility branch runs. Non-agile mid-air gap closed.

code/code/misc/movement.cc
kjsevertson: Addressed in `7f3b54494` — chose option 2: dropped `addToWait` from `stumble()` entirely. Lag is now applied by each caller's existing `addSkillLag(skill, rc)` flow, which routes through `discArray[skill]->lag` and `lagAdjust(skill)`. Per-skill tuning preserved.

code/code/misc/movement.cc
kjsevertson: Addressed in `e3309583e` — dropped both the `nullptr` default in `being.h` and the `victim ? victim : fight()` fallback inside `stumble()`. Signature is now `int stumble(TBeing* victim)` and `trySpringleap(victim)` is called directly.

code/code/disc/disc_monk_leverage.cc
kjsevertson: Addressed in `e3309583e` — normalized the hit-path DELETE checks in `hurl`, `shoulderThrow`, `defenestrate`, and the immediate `disc_monk_leverage.cc` site to the combined `if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT)) return rc;` form, matching the miss paths.

kjsevertson:
Thanks for the thorough review. All five inline threads are addressed and resolved.

**Inline (resolved):**
- A — UAF crash on caster damage: removed caster damage entirely (`e3309583e`)
- B — flying/fall-sector non-agile gap: hoisted to early return (`e3309583e`, `b44812cca`)
- C — hardcoded `combatRound(1)` lag: deferred to caller's `addSkillLag` (`7f3b54494`)
- D — `nullptr` default + `fight()` fallback: dropped (`e3309583e`)
- E — DELETE idiom inconsistency in hit paths: normalized (`e3309583e`)

**FYI items:**
- ✅ `tripFail` + grapple miss sites unmigrated → migrated in `b44812cca`
- ✅ Bodyslam generic-fail flavor mismatch → "can't get a grip" replaced with "loses \$s footing" so the cause flows naturally into stumble's effect (`15b74d045`)
- ✅ PR description mentioned doorbash / `slamIntoWall` / springleap-from-doorbash → trimmed
- ⚠️ \`IS_SET_ONLY → IS_SET_DELETE\` commit message inaccuracy → noted, historical, no fix
- ⚠️ Springleap-success ends attacker STANDING (not SITTING) → noted, audit observation only
- ⏸️ \`cmd_headbutt.cc:56\` "I don't understand this logic" comment → kept intentionally as a marker for a separate warrior-skills audit follow-up

Test plan items for trip / grapple / flying-fall-sector recovery are unchecked pending in-game verification.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,57 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
+        nullptr, TO_ROOM);
+    }
+
+    int rc = crashLanding(POSITION_SITTING);
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_THIS;
+
+    rc = trySpringleap(victim);
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+      return rc;
```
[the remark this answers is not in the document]
kjsevertson: Addressed in `10391dfa8`: changed stumble's recovery copy from "falls on $s butt!" to "topples over." which doesn't commit to a specific landing posture, and removed a duplicate "falls on $s butt" `act()` block from `cmd_grapple.cc` that was firing alongside stumble's message and contradicting it in flying-fall-sector cases. The `isFlying() && roomp->isFallSector()` early return at the top of `stumble()` handles the stay-aloft case directly.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,57 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
 
+int TBeing::stumble(TBeing* victim) {
+  if (!hasLegs())
+    return FALSE;
+
+  if (isFlying() && roomp->isFallSector()) {
+    sendTo(format("%sYou falter in the air, but stay aloft!%s\n\r") % green() %
+           norm());
+    act("<g>$n falters in the air, but stays aloft!<1>", true, this, nullptr,
+      nullptr, TO_ROOM);
```
wershlak, line 3501: The raw ansi helpers here seem to be outside conventions. I'm not sure if there's any potential implications though.

Overall I think it's cleaner to be consistent and the act() convention is more flexible and I think consistency would make this more readable:

  act("<g>You falter in the air, but stay aloft!<1>", true, this, nullptr, nullptr, TO_CHAR);
  act("<g>$n falters in the air, but stays aloft!<1>", true, this, nullptr, nullptr, TO_ROOM);

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,57 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
   return TRUE;
 }
 
+int TBeing::stumble(TBeing* victim) {
```
wershlak: I found the "victim" naming confusing here, thinking this was the victim being made to stumble. I think you could make a couple quick changes to this to make it a bit more readable for everyone.

Rename the victim parameter here to something like springleapTarget or opponent

I'd also suggest making it nullable for times when someone stumbles outside a fight TBeing::stumble(TBeing* opponent = nullptr) - This also makes it more intuitive that you aren't making the victim stumble.

code/code/misc/movement.cc
```diff
@@ -3509,6 +3509,57 @@ int TBeing::crashLanding(positionTypeT pos, bool force, bool dam,
...
+    int rc = crashLanding(POSITION_SITTING);
+    if (IS_SET_DELETE(rc, DELETE_THIS))
+      return DELETE_THIS;
+
+    rc = trySpringleap(victim);
+    if (IS_SET_DELETE(rc, DELETE_THIS) || IS_SET_DELETE(rc, DELETE_VICT))
+      return rc;
+  } else {
+    setPosition(POSITION_STANDING);
+    if (isFlying()) {
```
wershlak, line 3522: isFlying() is positional and you've just setPosition so this code is unreachable.

wershlak (changes requested):

code/code/misc/movement.cc
kjsevertson: Addressed in `6d446d726` — converted all five `sendTo`+`green/red/norm` self-message lines in `stumble()` to `act(... TO_CHAR)`, paired with the existing `act(... TO_ROOM)` calls. Same color tags, same audience targeting on both halves.

Saved this as a project-wide convention going forward: prefer `act()` over `sendTo()` for any player-facing message that has an actor to anchor to.

code/code/misc/movement.cc
kjsevertson: Rename done in `7259af6e4` — `stumble(TBeing* opponent)`. Agreed the `victim` framing read as 'the one being made to stumble' rather than the springleap target.

Declining the `= nullptr` default. All eleven current call sites are skill-miss handlers with an explicit non-null victim, and `doSpringleap` already has its own `fight()` fallback + `Springleap at whom?` guard one layer down (`disc_monk_leverage.cc:803-815`). An earlier round of this same decision dropped the prior `nullptr` default and `fight()` fallback inside `stumble()` itself as redundant (coderabbit thread on `3144729619`); re-adding the default would just bounce a hypothetical `nullptr` through to the layer that already handles it. Happy to revisit if a non-combat stumble caller materializes.

code/code/misc/movement.cc
kjsevertson: Good catch — fixed in `75edfd400`.

`isFlying()` is `getPosition() == POSITION_FLYING` (`being.cc:839`), so the check was always false after `setPosition(POSITION_STANDING)` and the agile-flying 'land on your feet' message was never reaching anyone.

Regression I introduced in `b44812cca` when I hoisted the duplicated `setPosition` out of both arms of the original `if (isFlying()) / else` — didn't notice the hoisted call clobbered the position the check reads. Moved `setPosition` to after the `if/else` so the check sees the pre-mutation position, mirroring the non-agile branch's compose-then-mutate order.

code/code/misc/movement.cc
```diff
@@ -3445,70 +3445,94 @@ void TBeing::doLand() {
...
+  if (changed)
+    setPosition(newPos);
+
+  // In fall sectors checkFalling handles the narrative — either the character
+  // plummets (with its own messaging) or is held in place by an upstream
+  // condition. Either way, the sit/sprawl flavor would be misleading, so skip
+  // the position-change messaging here.
+  if (roomp->isFallSector()) {
     if (falling || IS_SET_DELETE(checkFalling(), DELETE_THIS))
       return DELETE_THIS;
```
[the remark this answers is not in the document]
kjsevertson: Fixed in 4cd47f412 — changed the short-circuit from `falling || IS_SET_DELETE(checkFalling(), DELETE_THIS)` to `!falling && IS_SET_DELETE(checkFalling(), DELETE_THIS)`. The flag now correctly suppresses recursion without fabricating a death signal.

code/code/cmd/cmd_doorbash.cc
```diff
@@ -41,7 +41,7 @@ int TBeing::slamIntoWall(roomDirData* exitp) {
   addToWait(combatRound(12));
 #else
   int rc;
-  rc = crashLanding(POSITION_RESTING);
+  rc = crashLanding();
   if (IS_SET_DELETE(rc, DELETE_THIS))
```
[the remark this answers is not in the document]
kjsevertson: Rejecting — `slamIntoWall` has always called `crashLanding` here; this isn't a new substitution. The only PR change is the severity argument: `crashLanding(POSITION_RESTING)` → `crashLanding()`. Under the new design, the agility roll picks SITTING or RESTING. The wall-collision call site itself is unchanged, and the harsher-landing intent is preserved by the severity model (we'd raise severity for wall hits when severity gets tied to skill damage in a follow-up).

code/code/misc/damage.cc
```diff
@@ -682,14 +681,14 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
...
-    // force a doubly failed rideCheck for riders of victim
-    if (tb && !tb->rideCheck(-3) && !tb->rideCheck(-3)) {
-      rc = tb->fallOffMount(v, POSITION_SITTING);
+    if (!tb)
+      continue;
+    rc = tb->knockOffMount(3);
+    if (IS_SET_DELETE(rc, DELETE_THIS)) {
+      delete tb;
+      tb = nullptr;
+    } else if (rc) {
```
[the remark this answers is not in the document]
kjsevertson: Pre-existing convention — this `delete tb` mirrors the original `fallOffMount` callsite at this exact location and the same idiom used at `combat.cc canFight` for secondary-rider cleanup. The migration preserved the deletion pattern unchanged from master. A broader audit of player-rider deletion safety would be its own dedicated effort. Not introduced by this PR.

code/code/cmd/cmd_grapple.cc
```diff
@@ -118,26 +126,17 @@ static int grapple(TBeing* c, TBeing* victim, spellNumT skill) {
...
+      // flavor instead of crashLanding so the messages match the takedown
+      // context and the auto-springleap chain doesn't fire mid-pin.
+      c->setPosition(POSITION_SITTING);
+      victim->setPosition(POSITION_RESTING);
+      act("<g>You kneel over $N, pinning $M to the $g.<1>", true, c, nullptr,
+        victim, TO_CHAR);
+      act("<r>$n kneels over you, pinning you to the $g.<1>", true, c, nullptr,
+        victim, TO_VICT);
+      act("$n kneels over $N, pinning $M to the $g.", true, c, nullptr, victim,
+        TO_NOTVICT);
```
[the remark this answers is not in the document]
kjsevertson: Acknowledged — aerial grapple is allowed by `canGrapple` when both fighters are flying (`isFlying() && !c->isFlying()` is the only flying gate; both-flying passes). The ground-pin behavior is pre-existing — the original code at this site also forced ground positions via `dismount(POSITION_STANDING)` + custom flavor. Grapple is getting a full rework where aerial vs ground can be modeled properly. Deferring to that effort.

kjsevertson:
Addressed in `4cd47f412` and the prior rebased commit `1a45f8c11`. Status of all open review threads:

**CodeRabbit (latest review, 2026-05-03):**

In-diff inline replies posted on each thread. Summary:
- ✅ `movement.cc:3470` — `falling=true` short-circuit fixed.
- ❌ `cmd_doorbash.cc:45` — `slamIntoWall` has always called `crashLanding`; this PR only changed the severity arg. Wall-collision intent preserved by the severity model (will raise severity for wall hits when severity gets tied to skill damage).
- ⏸️ `cmd_grapple.cc:139` — aerial grapple is pre-existing behavior; deferred to the upcoming grapple rework where aerial/ground can be modeled properly.
- ❌ `damage.cc:690` — `delete tb` is pre-existing convention from the original `fallOffMount` callsite, unchanged by this PR.

Outside-diff:
- ✅ `disc_psionics.cc:716` — `IS_SET_ONLY(rc, DELETE_VICT)` was checking the wrong delete flag for `crashLanding`'s callee-side death. Fixed in `4cd47f412` — now `IS_SET_DELETE(rc, DELETE_THIS)` and translates to `DELETE_VICT` for the caller. Real latent UAF that the diff happened to surface.
- ✅ `cmd_bodyslam.cc:134-156` — pre-commitment messaging issue (lift-and-slam fired before knockOffMount). Fixed via a new "first-half action" flavor pattern across bash/bodyslam/spin: the skill describes its setup/wind-up, then `crashLanding`/`knockOffMount` narrates the result (sit/sprawl, thrown clear, hangs on, stays aloft). No more contradictions when a rideCheck save lets the rider hold their seat.
- ❌ `mobact.cc:2002-2018` — shaman drain-spell tier filter, wholly unrelated to this PR.

**wershlak (2026-05-01, pre-rebase):**

All three inline comments were addressed by commits that are now in the rebased branch:
- ✅ `movement.cc:3499` — sendTo→act consistency (`6d446d726`)
- ✅ `movement.cc:3512` — rename `victim` → `opponent` (`7259af6e4`)
- ✅ `movement.cc:3520` — dead `isFlying()` check (`75edfd400`)

**Two pre-existing latent UAFs surfaced by review and deferred to follow-up PRs** (saved to project memory):
- `disc_thief_stealth.cc:981` — `subterfugeMiss` discards `crashLanding`'s return.
- `handler.cc:198` — `affectChange` discards return; the existing "I don't think this can happen" comment is wrong but the proper fix requires changing `affectChange` from `void` to `int` and propagating through 10-20 caller sites.

Neither was introduced or worsened by this branch.

code/code/cmd/cmd_grapple.cc
```diff
@@ -99,11 +98,20 @@ static int grapple(TBeing* c, TBeing* victim, spellNumT skill) {
...
+          // Hung on — grapple can't wrestle a mounted target down.
+          act("Your grapple slips as $N keeps $S seat.", true, c, 0, victim,
+            TO_CHAR);
+          act("$n's grapple slips as $N keeps $S seat.", true, c, 0, victim,
+            TO_NOTVICT);
+          act("$n grapples at you but slips as you keep your seat.", true, c, 0,
+            victim, TO_VICT);
+          c->cantHit += c->loseRound(2);
+          return true;
+        }
```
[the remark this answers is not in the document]
kjsevertson: Intended behavior — grapple is a wrestling takedown, not a maneuver that respects the victim's flight ability. If your dragon-rider is grappled and thrown clear, the grappler then drags them to the ground as part of the pin. The `POSITION_FLYING` overwrite to `POSITION_RESTING` is the takedown completing. (Aerial grapple as a whole still gets revisited in the upcoming grapple rework, but for now this conversion is by design.)

code/code/cmd/cmd_bodyslam.cc
```diff
@@ -148,41 +131,44 @@ int TBeing::bodyslamMiss(TBeing* victim, skillMissT type) {
...
+      return DELETE_VICT;
+    // Only narrate the throw when the rider was actually dismounted; if they
+    // hung on, knockOffMount printed its own "hangs on tight" flavor.
+    if (!victim->riding) {
+      act("You pull $N from $p, throwing $M down!", false, this, mount, victim,
+        TO_CHAR);
+      act("$n pulls you from $p, throwing you down!", false, this, mount,
+        victim, TO_VICT, ANSI_RED);
+      act("$n pulls $N from $p, throwing $M down!", false, this, mount, victim,
+        TO_NOTVICT);
```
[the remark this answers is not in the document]
kjsevertson: Intentionally generic — the payoff phrase is "throwing $M down," not "to the $g." A rider who catches themselves in midair after being thrown clear is still being thrown down by the bodyslammer; the bodyslam's narrative just doesn't commit to where they land. Combined with knockOffMount's own "catches yourself in the air" message that fires for the stays-aloft branch, the room sees both: "$N pulls $N from $p, throwing $M down!" + "$N catches $mself in the air." Reads as a successful throw with a graceful midair recovery.

code/code/misc/damage.cc
```diff
@@ -666,12 +666,19 @@ int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
...
-      rc = v->fallOffMount(v->riding, POSITION_SITTING);
-      v->addToWait(combatRound(2));
+    if (!v->rideCheck(-rideSeverity)) {
+      rc = v->knockOffMount(rideSeverity);
       if (IS_SET_DELETE(rc, DELETE_THIS))
         return DELETE_VICT;
+      if (rc)
+        v->addToWait(combatRound(2));
     }
   } else if (v->riding && dynamic_cast<TMonster*>(v) && !v->desc) {
```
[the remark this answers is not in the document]
kjsevertson: Pre-existing — the old fallOffMount-on-damage block also gated on `dynamic_cast<TBeing*>(v->riding)`, and `fallOffMount` similarly distinguishes TBeing vs TThing mount paths. The TBeing-only gate likely reflects a real semantic distinction (sitting in a chair shouldn't dislodge you from damage like falling off a horse would). Out of scope for this PR; if we want to revisit chair/vehicle dismount-on-damage it's its own design decision.

code/code/misc/riding.cc
```diff
@@ -732,30 +732,106 @@ int TBeing::doMount(const char* arg, cmdTypeT cmd, TBeing* h,
...
+int TBeing::getRideMod() {
+  int mod = 0;
+
+  TBeing* mount = dynamic_cast<TBeing*>(riding);
+  if (!mount)
+    return mod;
+
+  // Roll once and reuse — bSuccess advances skill learning, so multiple calls
+  // would inflate the rider's advanced-riding XP per check.
+  bool advRiding = bSuccess(SKILL_ADVANCED_RIDING);
```
wershlak, line 747: This is a real behavior change worth surfacing: in master, `rideCheck` only called `bSuccess(learn, SKILL_RIDE)` once - it did not advance `SKILL_ADVANCED_RIDING` or any mount-type skill. Now `getRideMod` calls `bSuccess(SKILL_ADVANCED_RIDING)` and `bSuccess(mountSkill)` (further down at line 786), both of which advance learning. Combined with the new call frequency - every damaging hit via `damageEpilog`, every attack swing via `canFight`, plus once per secondary rider per event - advanced-riding and mount-type skills will now train substantially faster than they used to. The "Roll once and reuse" comment correctly explains intra-call dedup but is easy to misread as rate-limiting overall; it doesn't.

Also: `getRideMod` is non-`const` because of these `bSuccess` calls, but the name says "get". A caller reading `learn = getSkillValue(SKILL_RIDE) + getRideMod() + mod;` in `rideCheck` won't suspect that asking for the modifier teaches the skill.

A couple of options:
- Rename to `applyRideMod` or `rollRideMod` to make the side effect part of the contract
- Hoist the `bSuccess(...)` calls out of `getRideMod` into one place where they fire once per logical "ride event" (not per check), and keep `getRideMod` as a `const` value-only computation

Either way, worth updating the comment to acknowledge that XP is awarded per `rideCheck`, not deduplicated across them. Feel free to push back if the accelerated XP rate is what you want - it's just enough of a balance shift that I'd want it called out explicitly somewhere.

code/code/misc/riding.cc
```diff
@@ -732,30 +732,106 @@ int TBeing::doMount(const char* arg, cmdTypeT cmd, TBeing* h,
...
-  if (tbt && tbt->hasSaddle() && tbt->horseMaster() == this)
-    // only guy in saddle gets benefit
+// Sum the situational modifiers that apply to any riding skill check.
+// Positive values favor the rider. Includes equipment, class, combat state,
+// stat reactions, intoxication/hunger/thirst, limb injuries, and exhaustion.
+int TBeing::getRideMod() {
+  int mod = 0;
+
+  TBeing* mount = dynamic_cast<TBeing*>(riding);
+  if (!mount)
```
wershlak, line 742: The early-return-0-for-non-`TBeing`-mounts is correct (chairs and other inert seats don't get situational modifiers) but reads as if the function forgot the chair case. A one-line docblock at the top of `getRideMod` ("Returns 0 for non-`TBeing` mounts - chairs etc.; situational modifiers don't apply to inert seats") would prevent the next confused reader. Feel free to skip if you think it's obvious.

Non-blocking nit, separate from the larger XP concern above.

code/code/misc/damage.cc
```diff
@@ -641,7 +641,7 @@ int TBeing::applyDamage(TBeing* v, int dam, spellNumT dmg_type) {
 }
 
 // DELETE_VICT or FALSE
-int TBeing::damageEpilog(TBeing* v, spellNumT dmg_type) {
+int TBeing::damageEpilog(TBeing* v, int dam, spellNumT dmg_type) {
```
wershlak, line 644: `dam` is a new parameter on a function that does many things, and its only use is one line ~30 lines down (`int rideSeverity = std::min(dam, 50);`). A reader scanning `damageEpilog` won't immediately see why an "epilog" cares about damage magnitude.

Two options:
- Add a brief comment at the parameter declaration in `being.h:1496` explaining that `dam` feeds rider-jostle severity scaling
- Or extract the rider-jostling block (~25 lines) into a private helper like `jostleRiders(TBeing* v, int dam)` so `damageEpilog` doesn't need a new parameter and the rider logic lives in one named place

The second is cleaner but more churn. Either is fine - feel free to leave as-is if you don't want to expand the diff.

code/code/misc/combat.cc
```diff
@@ -4745,15 +4745,17 @@ bool TBeing::canFight(TBeing* target) {
     return FALSE;
   }
 
+  // Pre-check on every attack so swinging from horseback doesn't spam
+  // knockOffMount's hang-on flavor. Only escalate when the rider is already
+  // struggling. Mirror the pattern used in damage.cc.
```
wershlak, line 4750: The "Mirror the pattern used in damage.cc" claim isn't quite accurate. `canFight` uses `rideCheck(0)` and `knockOffMount()` (severity 0 default). `damage.cc:676` uses `rideCheck(-rideSeverity)` and `knockOffMount(rideSeverity)` where `rideSeverity` scales with damage. They diverge intentionally - which is fine - but the comment claims they mirror.

Suggested rewording: "Initiating an attack from the saddle uses a flat severity-0 check; mid-combat hits in `damageEpilog` scale severity to damage taken."

code/code/cmd/cmd_grapple.cc
```diff
@@ -118,26 +126,17 @@ static int grapple(TBeing* c, TBeing* victim, spellNumT skill) {
...
-      rc = victim->trySpringleap(c);
-      if (IS_SET_DELETE(rc, DELETE_THIS) && IS_SET_DELETE(rc, DELETE_VICT))
-        return rc;
-      else if (IS_SET_DELETE(rc, DELETE_THIS))
-        return DELETE_VICT;
-      else if (IS_SET_DELETE(rc, DELETE_VICT))
-        return DELETE_THIS;
+      // Both end up pinned on the ground. Use setPosition + pin-specific
+      // flavor instead of crashLanding so the messages match the takedown
+      // context and the auto-springleap chain doesn't fire mid-pin.
```
wershlak, line 131: The comment justifies bypassing `crashLanding` (auto-springleap shouldn't fire mid-pin) which reads well. Worth noting in the same comment that this also skips `checkFalling` - if a grapple succeeds in a fall sector, both characters are pinned without the fall-sector machinery running. Probably what you want (a wrestler holding their victim isn't going to drop them off a cliff), but a future change to fall-sector handling won't surface this site without an explicit mention.

Non-blocking - just expand the comment if you agree the exemption is intentional.

code/code/misc/being.h
```diff
@@ -460,6 +460,7 @@ class TBeing : public TThing {
     virtual void setWimpy(int);
     virtual short int hitLimit() const;
     virtual int fallOffMount(TThing*, positionTypeT, bool death = FALSE);
+    int knockOffMount(int severity = 0);
```
wershlak, line 463: `fallOffMount` (line above) is virtual with a `TThing` stub override at `riding.cc:769` for chair/object riders. The new `knockOffMount` is non-virtual, so calling it through a `TThing*` rider won't compile. Whether this matters depends on whether non-being mounts are conceptually subject to "hostile dislodge" - probably not, since you can't really throw someone off a chair - but the asymmetry next to `fallOffMount` will read as oversight.

Either add a `TThing` stub matching the `fallOffMount` pattern (returns `FALSE`), or add a one-line comment here: `// TBeing-only - chair/object riders use dismount() directly; hostile dislodge has no meaning for them.`

Feel free to skip if you don't want to expand the surface area.

code/code/disc/disc_monk_leverage.cc
```diff
@@ -788,166 +766,59 @@ int defenestrate(TBeing* caster, TBeing* victim, sstring direction) {
...
-
-  return doSpringleap("", false, vict);
-}
-
-int TBeing::doSpringleap(sstring argument, bool should_lag, TBeing* vict) {
-  TBeing* victim;
-  sstring name_buf;
-  int rc;
-
+int TBeing::doSpringleap() {
```
wershlak, line 783: `doSpringleap` and `springleap` independently check `doesKnowSkill(SKILL_SPRINGLEAP)` and the position range. `doSpringleap` then calls `springleap` which redoes both checks - so the guards in `doSpringleap` are pure duplication once you trace the call.

The duplication exists because `doSpringleap` wants user-facing messages while `springleap` needs to be silent for `crashLanding`. One way to resolve: drop the silent guards from `doSpringleap`, keep only the message-emitting versions, then call `springleap` and let it be the single source of truth for the gating. The user-facing message paths still work because they fire before the call.

Minor DRY suggestion - feel free to leave it if you prefer the current explicit form.

code/code/misc/movement.cc
```diff
@@ -3445,68 +3445,94 @@ void TBeing::doLand() {
...
-    }
+  } else if (changed && getPosition() == POSITION_RESTING) {
+    act("<r>You sprawl across the $g.<1>", true, this, nullptr, nullptr,
+      TO_CHAR);
+    act("<r>$n sprawls across the $g.<1>", true, this, nullptr, nullptr,
+      TO_ROOM);
+  }
+
+  if (doesKnowSkill(SKILL_SPRINGLEAP))
+    springleap();
```
wershlak, line 3484: `springleap()` returns an int but the return is silently discarded here. Today `springleap` always returns `TRUE`/`FALSE` (no DELETE possible), so this is fine. But if `springleap` ever evolves to do damage or anything else that could DELETE, this becomes a silent UAF in `crashLanding`'s callers.

Forward-defensive option: annotate `springleap()` declaration in `being.h:1866` with `[[nodiscard]]` so any future modification has to face this decision explicitly. Or change `springleap` to return `void` and make the contract unambiguous - the only place that uses the return is `doSpringleap` for `addSkillLag`, and that always sees `TRUE`/`FALSE` from the current implementation.

Not blocking - just want to flag it before the contract solidifies.

wershlak (commented):
Read through this carefully - the safety story holds up under scrutiny. DELETE flag propagation through every refactored skill is correct, the rider-iteration loops cache `t2` properly, and the new `knockOffMount` chain composes cleanly with `crashLanding`. Found a handful of design things worth talking through, left inline. None are blocking, but the `getRideMod` one is worth a look before merge - there's a meaningful change to skill-XP behavior that I don't think is obvious from the function name.

A few FYIs that didn't feel worth inline comments:

- Players who type `springleap <target>` will have the argument silently dropped at `parse.cc:316` - old combat-style usage gets a stand-up animation with no feedback that anything was ignored. Worth at least a help-file update if the old syntax was documented.
- The hostile-throw path prints two landing narrations: `knockOffMount`'s "thrown clear of $N" line, then `dismount`, then `crashLanding`'s sit/sprawl flavor. Works in code but worth a quick in-game sanity check that it scans naturally.
- Both pre-existing latent UAFs you flagged in the description (`handler.cc:198` `affectChange` and `disc_thief_stealth.cc:982` `subterfugeMiss`) showed up independently in review - confirming they're not introduced or worsened here.
- `canFight`'s rider loop still processes one rider per attack (the early `return false` short-circuits subsequent riders) - pre-existing behavior, just noting the rewrite didn't fix it.

wershlak:
@kjsevertson I posted the review here from Claude. I looked at it and didn't see much of anything that seemed worth doing. Give it a once over and let me know if you think it's all good or if there's anything you think you want to update first.

wershlak (approved):
I think we've gotten the code here in good shape and the handling seems overall safer so that's awesome.

PR #702, opened 2026-03-22
Rework stab skill lag, mount checks, spear support, and safety fixes

kjsevertson:
## Summary

- **Skill lag timing**: Stab lag moves from before the attack to after. Killing blow reduces lag via `addSkillLag`'s existing `DELETE_VICT` handling.
- **Mounted combat**: Thieves with ride skill >= 80 can stab/backstab while mounted. Flying mounts grant air reach. Replaces the old flat ban on all mounted stab/backstab.
- **Spear support**: Paired pierce weapons (`isSpear()`) can now stab and backstab. Backstab with spears requires 50+ skill. Spear-wielding thief mobs always prefer backstab over throat-slit.
- **Backstab→stab chain**: At 50+ advanced stab learning, a successful backstab chains into stab without double-charging move cost or skill cooldown.
- **Throat-slit height**: Now accounts for victim position — can slit a sitting/prone victim even if they'd be too tall standing.
- **Safety**: Fixes `checkSpec` return handling and `DELETE_THIS` propagation across the entire stab/backstab/throatslit chain.

### Gameplay: skill lag timing (`cmd_stab.cc`, `disc_thief_murder.cc`)

**Before**: `addSkillLag(SKILL_STABBING, 0)` was called inside `stab()` before the attack resolved. The thief always paid full lag regardless of outcome.

**After**: `addSkillLag(SKILL_STABBING, rc)` is called in `doStab()` after `stab()` returns. The `rc` parameter carries `DELETE_VICT` on a kill, which `addSkillLag` uses to cap lag at `f_base` (1.0 multiplier) instead of the full skill lag — per `discipline.cc:3128`:
```cpp
if (IS_SET_DELETE(rc, DELETE_VICT))
    f_lag = min(f_base, f_lag);
```

Same pattern applied to `doBackstab` and `doThroatSlit` — both now pass `rc` to `addSkillLag` instead of `0`.

### Gameplay: mounted combat (`cmd_stab.cc:187-203`, `disc_thief_murder.cc:295-305`)

**Before**: `if (thief->riding) return FALSE` — flat ban, no mounted stabbing ever.

**After**: 
```cpp
TBeing* thiefMount = dynamic_cast<TBeing*>(thief->riding);
if (thiefMount && thief->getSkillLevel(SKILL_RIDE) < 80)
    // reject — not skilled enough
```
- `dynamic_cast` guards against furniture (chairs return nullptr, not a crash)
- Flying victim check: `(victim->isFlying() || (victimMount && victimMount->isFlying())) && !thiefAirborne` — if victim or their mount is flying, thief needs to be airborne too (flying or on a flying mount)
- Exception: `thief->fight() != victim` — if already in combat with the target, reach isn't rechecked (you're already engaged)

### Gameplay: spear support (`obj_general_weapon.cc:168-170`, `cmd_stab.cc:213`, `disc_thief_murder.cc:322-331`)

New method:
```cpp
bool TGenWeapon::isSpear() const {
    return isPierceWeapon() && isPaired();  // paired = two-handed
}
```

- **Stab**: `primCanStab` accepts `canStab() || isSpear()` (primary hand only — spears are two-handed)
- **Backstab**: `canBackstab() || isSpear()`, but spears require `bKnown >= 50`
- **Mob AI** (`mobact.cc:2857`): `canBackstab() || isSpear()` gates entry; spear-wielding mobs always use backstab (never throat-slit)

### Gameplay: backstab→stab chain (`disc_thief_murder.cc:189-194`, `cmd_stab.cc:246-250`)

When backstab hits but doesn't kill, and the thief has 50+ advanced stab learning:
```cpp
// In backstabHit, after reconcileDamage:
int rc = stabChain(this, victim);
```

`stabChain()` calls the internal `stab()` with `isChain=true`, which skips move cost and skill lag (the calling skill owns those). The stab still runs its own skill check (`bSuccess`), precondition checks, and full attack sequence.

### Gameplay: throat-slit height check (`disc_thief_murder.cc:644-645`)

**Before**: `6 * thief->getHeight() < 3 * victim->getHeight() && !thief->isFlying()` — if victim is too tall and thief isn't flying, reject.

**After**: Added `|| victim->getPosition() < POSITION_STANDING` — if the victim is sitting/prone, the height restriction doesn't apply (their throat is reachable).

### Safety: checkSpec return handling (`cmd_stab.cc:132-138`, `disc_thief_murder.cc:159-168`, `disc_thief_murder.cc:524-535`)

**Before** (all three sites):
```cpp
if (weapon->checkSpec(...) == DELETE_VICT)
    return DELETE_VICT;
```

**Problems**: 
1. `==` misses combined flags (e.g. `DELETE_VICT | DELETE_ITEM`)
2. `DELETE_ITEM` (thief killed by spec proc, since thief is passed as `t2`) silently dropped
3. `DELETE_THIS` (weapon destroyed) ignored — subsequent poison check uses dangling pointer

**After** (stabCore shown; backstabHit/throatSlitHit follow same pattern):
```cpp
int specRc = weapon->checkSpec(victim, CMD_STAB,
    reinterpret_cast<char*>(limb), thief);
if (IS_SET_DELETE(specRc, DELETE_VICT))
    return DELETE_VICT;
if (IS_SET_DELETE(specRc, DELETE_ITEM))
    return DELETE_THIS;  // thief (t2) was destroyed
// DELETE_THIS = weapon destroyed; not used after this point
```

Per `spec_objs.cc:101-103` contract: `DELETE_THIS` = calling object (weapon), `DELETE_VICT` = first arg (victim), `DELETE_ITEM` = fourth arg (thief). Translation: spec proc's `DELETE_ITEM` → caller's `DELETE_THIS`.

In backstabHit/throatSlitHit, weapon destruction also nulls the local pointer so the subsequent poison check is safely skipped:
```cpp
if (IS_SET_DELETE(specRc, DELETE_THIS))
    weapon = nullptr;  // weapon destroyed, skip poison
// ...
if (weapon && weapon->isPoisoned())  // guarded
```

### Safety: DELETE_THIS propagation (6 call sites)

The attacker-death signal must travel from `stabCore` all the way back to the top-level `doBackstab`/`doThroatSlit`/`doStab` without any intermediate code touching the dead attacker.

**Propagation chain** (backstab path shown):
```
stabCore returns DELETE_THIS (spec proc killed thief)
  → stabPrimAtk checks DELETE_THIS, returns rc          [cmd_stab.cc:154]
    → stab passes through (direct return)
      → stabChain passes through (direct return)
        → backstabHit checks DELETE_THIS, returns rc     [disc_thief_murder.cc:193]
          → backstab checks DELETE_THIS, returns rc      [disc_thief_murder.cc:377]
            → doBackstab checks DELETE_THIS, returns rc  [disc_thief_murder.cc:267]
              (before addSkillLag — prevents this-> call on dead attacker)
```

Each `do<Skill>` function now has an early return:
```cpp
if (IS_SET_DELETE(rc, DELETE_THIS))
    return rc;  // this is destroyed, don't touch it

if (rc)
    addSkillLag(SKILL_..., rc);  // safe — this is alive
```

### Structural: stabCore extraction (`cmd_stab.cc:95-141`)

`stabPrimAtk` and `stabOffAtk` contained identical attack logic. Extracted into `stabCore()` — both now delegate to it. `stabOffAtk` adds the weapon eligibility check; `stabPrimAtk` adds the dual-wield offhand follow-up.

### Structural: stabChain / doStab cleanup (`cmd_stab.cc:246-250`, `being.h:1990`)

**Before**: `int doStab(const char*, TBeing*, bool chain = false)` — leaks chain concept onto public interface.

**After**: `int doStab(const char*, TBeing*)` — clean public API. `stabChain()` is a free function declared in `cmd_stab.h`, called only from `backstabHit`. `stabPrimAtk` restored to `static` linkage.

### Structural: mob AI simplification (`mobact.cc:2854-2882`)

**Before**: ~50 lines with nested `if` chains, separate victim-mounted/unmounted paths that duplicated `doBackstab`/`doThroatSlit` calls, stale `vlogf` debug pings, and commented-out `backstabHit` direct calls.

**After**: ~25 lines. Single path using `TGenWeapon*` cast and `canBackstab() || isSpear()`. Removed the victim-mount special-casing (the skills themselves now handle mounts). `DELETE_THIS` checks are now live (previously dead code since the signal was never propagated).

## Test plan

- [x] Clean build (`make`) — compiles with no warnings
- [x] Unit tests pass (`make test`) — 63/63
- [x] Functional: direct stab command with pierce weapon — verify attack + lag
- [x] Functional: stab with spear (paired pierce) — verify allowed in primary hand
- [x] Functional: backstab chains into stab at 50+ adv learning
- [x] Functional: mounted stab with ride >= 80 succeeds; ride < 80 rejected
- [x] Functional: stab against flying victim while grounded is rejected
- [x] Functional: throat-slit against tall sitting victim succeeds
- [x] Functional: thief mob uses backstab/slit in takeFirstHit
- [x] Code review: trace DELETE_THIS propagation from stabCore through doBackstab

### DIFF
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -285,18 +285,19 @@ int TBeing::doStab(const char* argument, TBeing* vict) {
   }
   rc = stab(this, victim);
 
+  // Clean up locally-resolved victim before the DELETE_THIS early return so
+  // combined DELETE_THIS | DELETE_VICT doesn't leak it.
+  if (IS_SET_DELETE(rc, DELETE_VICT) && !vict) {
+    delete victim;
+    victim = nullptr;
+    REM_DELETE(rc, DELETE_VICT);
+  }
+
   if (IS_SET_DELETE(rc, DELETE_THIS))
     return rc;
 
   if (rc)
     addSkillLag(SKILL_STABBING, rc);
 
-  if (IS_SET_DELETE(rc, DELETE_VICT)) {
-    if (vict)
-      return rc;
-    delete victim;
-    victim = nullptr;
-    REM_DELETE(rc, DELETE_VICT);
-  }
   return rc;
 }
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -131,11 +131,15 @@ static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
 
   int specRc = weapon->checkSpec(victim, CMD_STAB,
     reinterpret_cast<char*>(limb), thief);
+  // Translate checkSpec flags: DELETE_VICT = victim, DELETE_ITEM = thief (t2),
+  // DELETE_THIS = weapon (not used after this point, no action needed)
+  int rc = 0;
   if (IS_SET_DELETE(specRc, DELETE_VICT))
-    return DELETE_VICT;
+    rc |= DELETE_VICT;
   if (IS_SET_DELETE(specRc, DELETE_ITEM))
-    return DELETE_THIS;  // thief (t2) was destroyed
-  // DELETE_THIS = weapon destroyed; not used after this point
+    rc |= DELETE_THIS;
+  if (rc)
+    return rc;
 
   return TRUE;
 }
@@ -190,7 +194,7 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
     return FALSE;
 
   TBeing* thiefMount = dynamic_cast<TBeing*>(thief->riding);
-  if (thiefMount && thief->getSkillLevel(SKILL_RIDE) < 80) {
+  if (thiefMount && thief->getSkillValue(SKILL_RIDE) < 80) {
     thief->sendTo("You aren't a skilled enough rider to stab while mounted!\n\r");
     return FALSE;
   }
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -99,7 +99,6 @@ static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
     STAT_SPE, STAT_AGI, STAT_PER, false);
   if (specResult != COMPLETE_SUCCESS && specResult != GUARANTEED_SUCCESS) {
     stabMissMsg(thief, victim, weapon);
-    victim->addHated(thief);
     return TRUE;
   }
 
@@ -130,8 +129,11 @@ static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   stabBleedCheck(thief, victim, weapon, limb);
   stabPoisonCheck(victim, weapon);
 
-  if (weapon->checkSpec(victim, CMD_STAB, reinterpret_cast<char*>(limb),
-        thief) == DELETE_VICT)
+  int specRc = weapon->checkSpec(victim, CMD_STAB,
+    reinterpret_cast<char*>(limb), thief);
+  if (IS_SET_DELETE(specRc, DELETE_ITEM))
+    return DELETE_THIS;  // thief destroyed by weapon spec
+  if (IS_SET_DELETE(specRc, DELETE_VICT))
     return DELETE_VICT;
 
   return TRUE;
@@ -236,10 +238,11 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
   int bKnown = thief->getSkillValue(SKILL_STABBING);
   bool stabSuccess = !victim->awake() || thief->bSuccess(bKnown, SKILL_STABBING);
 
+  victim->addHated(thief);
+
   if (!stabSuccess) {
     TGenWeapon* weapon = primCanStab ? primWeapon : offWeapon;
     stabMissMsg(thief, victim, weapon);
-    victim->addHated(thief);
     return TRUE;
   }
 
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -92,7 +92,8 @@ static void stabBleedCheck(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
   }
 }
 
-// Does NOT add skilllag or move cost - caller is responsible for that
+// Resolves a single stab attempt for one weapon — hit roll through weapon
+// spec proc. Shared between primary-hand and offhand attacks.
 static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   int specResult = thief->specialAttack(victim, SKILL_STABBING, 0, STAT_DEX,
     STAT_SPE, STAT_AGI, STAT_PER, false);
@@ -129,17 +130,9 @@ static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   stabBleedCheck(thief, victim, weapon, limb);
   stabPoisonCheck(victim, weapon);
 
-  int specRc = weapon->checkSpec(victim, CMD_STAB,
-    reinterpret_cast<char*>(limb), thief);
-  // Translate checkSpec flags: DELETE_VICT = victim, DELETE_ITEM = thief (t2),
-  // DELETE_THIS = weapon (not used after this point, no action needed)
-  int rc = 0;
-  if (IS_SET_DELETE(specRc, DELETE_VICT))
-    rc |= DELETE_VICT;
-  if (IS_SET_DELETE(specRc, DELETE_ITEM))
-    rc |= DELETE_THIS;
-  if (rc)
-    return rc;
+  if (weapon->checkSpec(victim, CMD_STAB, reinterpret_cast<char*>(limb),
+        thief) == DELETE_VICT)
+    return DELETE_VICT;
 
   return TRUE;
 }
@@ -194,16 +187,26 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
     return FALSE;
 
   TBeing* thiefMount = dynamic_cast<TBeing*>(thief->riding);
+  if (thief->riding && !thiefMount) {
+    act("You can't stab anyone from atop $p!", false, thief, thief->riding,
+      nullptr, TO_CHAR);
+    return false;
+  }
   if (thiefMount && thief->getSkillValue(SKILL_RIDE) < 80) {
     thief->sendTo("You aren't a skilled enough rider to stab while mounted!\n\r");
-    return FALSE;
+    return false;
   }
 
   TBeing* victimMount = dynamic_cast<TBeing*>(victim->riding);
   bool thiefAirborne = thief->isFlying() || (thiefMount && thiefMount->isFlying());
-  if ((victim->isFlying() || (victimMount && victimMount->isFlying())) && !thiefAirborne && thief->fight() != victim) {
+  bool victimAirborne = victim->isFlying() || (victimMount && victimMount->isFlying());
+  // Mid-fight against the same target bypasses the airborne mismatch —
+  // engaging a flying opponent puts them in stab range. (Backstab has no
+  // such carve-out; its mid-fight gate already requires off-feet, which
+  // flying opponents inherently aren't.)
+  if (victimAirborne && !thiefAirborne && thief->fight() != victim) {
     thief->sendTo("You can't stab a flying person with your feet on the ground!\n\r");
-    return FALSE;
+    return false;
   }
 
   if (thief->noHarmCheck(victim))
@@ -214,7 +217,7 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
   TGenWeapon* offWeapon =
     dynamic_cast<TGenWeapon*>(thief->heldInSecHand());
 
-  bool primCanStab = primWeapon && (primWeapon->canStab() || primWeapon->isSpear());
+  bool primCanStab = primWeapon && (primWeapon->canStab() || primWeapon->isPolearm());
   bool offCanStab = offWeapon && offWeapon->canStab();
 
   if (!primCanStab && !offCanStab) {
@@ -230,14 +233,13 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
     thief->addToMove(-STAB_MOVE);
   }
 
-  victim->addHated(thief);
-
   int bKnown = thief->getSkillValue(SKILL_STABBING);
   bool stabSuccess = !victim->awake() || thief->bSuccess(bKnown, SKILL_STABBING);
 
   if (!stabSuccess) {
     TGenWeapon* weapon = primCanStab ? primWeapon : offWeapon;
     stabMissMsg(thief, victim, weapon);
+    victim->addHated(thief);
     return TRUE;
   }
 
@@ -248,7 +250,8 @@ static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
 }
 
 // Chain entry point for backstab. Runs the full stab (preconditions + attack)
-// but skips move cost and skilllag since the calling skill owns those.
+// without charging the stab move cost — the chained stab is a follow-on to
+// backstab, not an independent action.
 int stabChain(TBeing* thief, TBeing* victim) {
   return stab(thief, victim, /*isChain=*/true);
 }
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -284,9 +284,11 @@ int TBeing::doStab(const char* argument, TBeing* vict) {
     return FALSE;
   }
   rc = stab(this, victim);
+  const int lagRc = rc;
 
   // Clean up locally-resolved victim before the DELETE_THIS early return so
-  // combined DELETE_THIS | DELETE_VICT doesn't leak it.
+  // combined DELETE_THIS | DELETE_VICT doesn't leak it. lagRc preserves the
+  // original DELETE_VICT bit so addSkillLag still applies its kill cap.
   if (IS_SET_DELETE(rc, DELETE_VICT) && !vict) {
     delete victim;
     victim = nullptr;
@@ -296,8 +298,8 @@ int TBeing::doStab(const char* argument, TBeing* vict) {
   if (IS_SET_DELETE(rc, DELETE_THIS))
     return rc;
 
-  if (rc)
-    addSkillLag(SKILL_STABBING, rc);
+  if (lagRc)
+    addSkillLag(SKILL_STABBING, lagRc);
 
   return rc;
 }
--- a/code/code/cmd/cmd_stab.cc
+++ b/code/code/cmd/cmd_stab.cc
@@ -5,13 +5,14 @@
 #include "obj_general_weapon.h"
 #include "materials.h"
 #include "skills.h"
+#include "spells.h"
+#include "cmd_stab.h"
 
-// Forward declarations
 static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon);
+
 static int stabPrimAtk(TBeing* thief, TBeing* victim, TGenWeapon* primWeapon,
   TGenWeapon* offWeapon);
 
-// Display hit message for a successful stab
 static void stabHitMsg(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
   wearSlotT limb) {
   sstring limbName(victim->describeBodySlot(limb));
@@ -52,20 +53,17 @@ static void stabHitMsg(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
   }
 }
 
-// Display miss message for a failed stab
 static void stabMissMsg(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   act("You miss your thrust into $N.", FALSE, thief, weapon, victim, TO_CHAR);
   act("$n misses $s thrust into you.", FALSE, thief, weapon, victim, TO_VICT);
   act("$n misses $s thrust into $N.", FALSE, thief, weapon, victim, TO_NOTVICT);
 }
 
-// Apply poison from weapon if present
 static void stabPoisonCheck(TBeing* victim, TGenWeapon* weapon) {
   if (weapon->isPoisoned())
     weapon->applyPoison(victim);
 }
 
-// Check for bleeding based on weapon sharpness vs limb hardness
 static void stabBleedCheck(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
   wearSlotT limb) {
   if (victim->isUndead() || victim->isImmune(IMMUNE_BLEED, limb))
@@ -79,10 +77,8 @@ static void stabBleedCheck(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
     int duration = sharpness * Pulse::UPDATES_PER_MUDHOUR / 2;
 
     if (victim->isLimbFlags(limb, PART_BLEEDING)) {
-      // Already bleeding - increment the stack
       victim->incrementBleedStack(limb, duration);
     } else {
-      // New bleed
       sstring limbName(victim->describeBodySlot(limb));
       act(format("Your stab opens a <r>bleeding wound<z> on $N's %s!") % limbName,
         FALSE, thief, nullptr, victim, TO_CHAR);
@@ -96,12 +92,8 @@ static void stabBleedCheck(TBeing* thief, TBeing* victim, TGenWeapon* weapon,
   }
 }
 
-// Offhand stab attack - uses SKILL_STABBING for specAttack
-static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
-  if (!weapon || !weapon->canStab())
-    return FALSE;
-
-  // Use DEX/SPE for offense, AGI/PER for defense
+// Does NOT add skilllag or move cost - caller is responsible for that
+static int stabCore(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
   int specResult = thief->specialAttack(victim, SKILL_STABBING, 0, STAT_DEX,
     STAT_SPE, STAT_AGI, STAT_PER, false);
   if (specResult != COMPLETE_SUCCESS && specResult != GUARANTEED_SUCCESS) {
@@ -110,12 +102,11 @@ static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
     return TRUE;
   }
 
-  // Calculate damage for offhand
   int level = thief->getSkillLevel(SKILL_STABBING);
   int dam = thief->getSkillDam(victim, SKILL_STABBING, level,
     thief->getAdvLearning(SKILL_STABBING));
   dam = thief->getActualDamage(victim, weapon, dam, SKILL_STABBING);
-  
+
   if (dam <= 0) {
     act("Your $o fails to penetrate $N's thick hide.", FALSE, thief, weapon,
       victim, TO_CHAR);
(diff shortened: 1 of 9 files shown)


code/code/disc/disc_thief_murder.cc
```diff
@@ -154,13 +156,19 @@ int TBeing::backstabHit(TBeing* victim, TThing* obj, int modifier) {
...
-          return DELETE_VICT;
+        if (weapon) {
+          int specRc = weapon->checkSpec(victim, CMD_BACKSTAB,
+            "-special-", this);
+          if (IS_SET_DELETE(specRc, DELETE_VICT))
+            return DELETE_VICT;
+          if (IS_SET_DELETE(specRc, DELETE_ITEM))
+            return DELETE_THIS;  // this (t2=attacker) was destroyed
+          if (IS_SET_DELETE(specRc, DELETE_THIS))
+            weapon = nullptr;  // weapon destroyed, skip poison
```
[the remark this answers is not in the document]
kjsevertson: Fixed in 30e40f0. The three checkSpec handlers now accumulate flags with `rc |=` before returning, so combined flags like `DELETE_VICT | DELETE_ITEM` are fully preserved in the translated return code.

kjsevertson:
@copilot resolve the merge conflicts in this pull request

code/code/disc/disc_thief_murder.cc
```diff
@@ -255,9 +265,12 @@ int TBeing::doBackstab(const char* argument, TBeing* vict) {
...
-      dynamic_cast<TMonster*>(victim)->US(25);
+  rc = backstab(this, victim);
+
+  if (IS_SET_DELETE(rc, DELETE_THIS))
+    return rc;
+
+  if (rc)
     addSkillLag(SKILL_BACKSTAB, rc);
     REMOVE_BIT(specials.affectedBy, AFF_HIDE);
   }
```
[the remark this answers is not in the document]
kjsevertson: Fixed in 391fedb5. The merge conflict resolution in `c5b912f37` mis-integrated master's `REMOVE_BIT(specials.affectedBy, AFF_HIDE)` (from `5f3617bd6`), producing an unbraced `if (rc)` with an orphan `}` in both `doBackstab` and `doThroatSlit`.

I rebased the branch onto current master instead of carrying the merge commit, then re-integrated the hide-break cleanly:

```cpp
  if (rc) {
    addSkillLag(SKILL_BACKSTAB, rc);
    REMOVE_BIT(specials.affectedBy, AFF_HIDE);
  }
```

Same shape in `doThroatSlit`. Full rebuild clean.

kjsevertson:
@coderabbitai the doThroatSlit lagRc + early-cleanup fix landed in 243b9620f, before this review was posted (review at 00:10 UTC, commit at 00:04 UTC). Lines 654–682 in the current HEAD already match the suggested diff. No action needed.

code/code/cmd/cmd_stab.cc
```diff
@@ -126,80 +118,39 @@ static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
...
     return DELETE_VICT;
 
-  // Check for bleeding and poison
   stabBleedCheck(thief, victim, weapon, limb);
   stabPoisonCheck(victim, weapon);
 
-  // Check weapon spec proc
   if (weapon->checkSpec(victim, CMD_STAB, reinterpret_cast<char*>(limb),
         thief) == DELETE_VICT)
     return DELETE_VICT;
```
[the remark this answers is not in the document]
kjsevertson: Dismissing — propagating `DELETE_ITEM`/`DELETE_THIS` from this site would be the bug, not the fix.

`parse.cc:1878-1882` auto-promotes `DELETE_ITEM` → `DELETE_VICT` at the top-level command dispatcher. If `doStab` propagated DELETE_ITEM upward, the dispatcher would interpret it as "victim died" and delete a still-alive victim. The `== DELETE_VICT` exact-match here is intentional — it's a revert to master form, made in `d777214f` after an earlier attempt at fancier flag handling was found to be broken.

A few reasons there's no safety gap:

1. After `checkSpec` returns, `stabCore` returns immediately. Up the chain, `primWeapon` is never dereferenced again — `stabPrimAtk` only touches `offWeapon` (a different object) post-call. So a weapon-destroying spec proc doesn't dangle anything here.
2. Per project convention (`safety-invariants.md`), object spec procs return `DELETE_ITEM` for self-destruction, not `DELETE_THIS`. The "deletes the attacker" scenario isn't a real pattern at this call site — `weapon->checkSpec` operates on the weapon, not the thief.
3. The chain sites that *do* need local weapon-destruction handling (`disc_thief_murder.cc` backstab/slit) keep their `DELETE_THIS → weapon = nullptr` guard, but also don't propagate `DELETE_ITEM` upward — same dispatcher-promotion reason.

The DELETE-flag rule isn't "always propagate every flag" — it's "don't dereference after deletion and propagate correctly for the caller's contract." For a top-level command, the correct contract is `DELETE_VICT` only.

code/code/cmd/cmd_stab.cc
```diff
@@ -235,62 +186,76 @@ static int stab(TBeing* thief, TBeing* victim) {
...
-
-  // Check bSuccess for stab skill
   int bKnown = thief->getSkillValue(SKILL_STABBING);
   bool stabSuccess = !victim->awake() || thief->bSuccess(bKnown, SKILL_STABBING);
 
   if (!stabSuccess) {
-    // Failed - show miss message with whichever weapon we would have used
     TGenWeapon* weapon = primCanStab ? primWeapon : offWeapon;
     stabMissMsg(thief, victim, weapon);
+    victim->addHated(thief);
```
wershlak: Hate is no longer added on a successful stab. Pre-PR `stab()` called `victim->addHated(thief)` unconditionally just before the `bSuccess` check, so any stab attempt (hit or miss) flagged the thief as hated. Post-PR `addHated` only fires here (bSuccess miss), in `stabCore` (specialAttack miss at line 102), and in `stabPrimAtk` (specialAttack miss at line 158). On a path where the special attack succeeds and damage applies, nothing calls `addHated`.

Sibling skills get this right: `backstab()` calls `addHated` after both the `backstabHit` success branch and the miss branch, and `throatSlit()` does the same. Suggest restoring one unconditional `victim->addHated(thief);` in `stab()` immediately before the `bSuccess` branch, and removing the per-path duplicates in `stabCore`/`stabPrimAtk`/`stab`.

code/code/cmd/cmd_stab.cc
```diff
@@ -126,80 +118,39 @@ static int stabOffAtk(TBeing* thief, TBeing* victim, TGenWeapon* weapon) {
...
-  // Apply damage
   if (thief->reconcileDamage(victim, dam, SKILL_STABBING) == -1)
     return DELETE_VICT;
 
-  // Check for bleeding and poison
   stabBleedCheck(thief, victim, weapon, limb);
   stabPoisonCheck(victim, weapon);
 
-  // Check weapon spec proc
   if (weapon->checkSpec(victim, CMD_STAB, reinterpret_cast<char*>(limb),
```
wershlak: The PR description lists this site as one of the checkSpec safety fixes (with `IS_SET_DELETE` plus a `DELETE_ITEM` to caller-`DELETE_THIS` translation), but the actual code here still uses the legacy `==` pattern. Two gaps:

- A combined return like `DELETE_VICT | DELETE_THIS` won't equal `DELETE_VICT`, the function falls through, and `stabPrimAtk`'s offhand step runs against a dangling victim.
- `DELETE_ITEM` from a weapon spec proc means `t2` (the thief) was destroyed per `spec_objs.cc:101-103` - `stabPrimAtk` then continues with a dead `thief`.

Both are latent today (`thiefQuestWeapon` is the only weapon spec on these CMDs and doesn't return either flag), so feel free to ignore if you want to keep this PR scoped to the backstab/slit chain. But the `backstabHit`/`throatSlitHit` updates already do the `IS_SET_DELETE` switch, so it'd be small to bring this site over too.

code/code/disc/disc_thief_murder.cc
```diff
@@ -154,13 +156,17 @@ int TBeing::backstabHit(TBeing* victim, TThing* obj, int modifier) {
 
         auto weapon = dynamic_cast<TBaseWeapon*>(obj);
 
-        if (weapon && weapon->checkSpec(victim, CMD_BACKSTAB, "-special-",
-                        this) == DELETE_VICT) {
-          return DELETE_VICT;
+        if (weapon) {
+          int specRc = weapon->checkSpec(victim, CMD_BACKSTAB,
```
wershlak, line 159: This now correctly handles `DELETE_THIS` (weapon destroyed -> null the local pointer) and `DELETE_VICT` (return), but not `DELETE_ITEM`. Per `spec_objs.cc:101-103`, `DELETE_ITEM` here means `t2` (this thief) was destroyed by the weapon spec. After this block, the function continues to use `this`: `applyPoison(victim)`, `reconcileDamage(victim, ...)`, `getAdvLearning(SKILL_STABBING)`, `stabChain(this, victim)`. If a spec proc ever returns `DELETE_ITEM` we'd UAF.

Same gap exists in `throatSlitHit` (see the parallel comment there). Both are latent today but worth handling alongside the rest of the safety chain. The translation would be `if (IS_SET_DELETE(specRc, DELETE_ITEM)) return DELETE_THIS;` so the existing `DELETE_THIS` guards in `backstab`/`doBackstab` actually have something to catch.

Feel free to defer to a follow-up PR if you'd rather keep scope tight here.

code/code/disc/disc_thief_murder.cc
```diff
@@ -500,13 +550,17 @@ int TBeing::throatSlitHit(TBeing* victim, TThing* obj, int modifier) {
 
         auto weapon = dynamic_cast<TGenWeapon*>(heldInPrimHand());
 
-        if (weapon && weapon->checkSpec(victim, CMD_SLIT, "-special-", this) ==
-                        DELETE_VICT) {
-          return DELETE_VICT;
+        if (weapon) {
+          int specRc = weapon->checkSpec(victim, CMD_SLIT,
```
wershlak, line 555: Same `DELETE_ITEM` gap as `backstabHit` (see disc_thief_murder.cc:160 comment) - if a weapon spec proc here returns `DELETE_ITEM`, this thief is dead but the function keeps using `this` afterwards (`applyPoison`, `reconcileDamage` indirectly via the surrounding flow, etc.). Same suggested fix as that comment.

code/code/disc/disc_thief_murder.cc
```diff
@@ -255,20 +260,32 @@ int TBeing::doBackstab(const char* argument, TBeing* vict) {
...
+  // Clean up locally-resolved victim before the DELETE_THIS early return so
+  // combined DELETE_THIS | DELETE_VICT doesn't leak it. lagRc preserves the
+  // original DELETE_VICT bit so addSkillLag still applies its kill cap.
+  if (IS_SET_DELETE(rc, DELETE_VICT) && !vict) {
+    delete victim;
+    victim = nullptr;
+    REM_DELETE(rc, DELETE_VICT);
+  }
+
+  if (IS_SET_DELETE(rc, DELETE_THIS))
```
wershlak, line 276: A note on the broader `DELETE_THIS` plumbing across this PR (`backstab`, `doBackstab`, `throatSlit`, `doThroatSlit`, `stab`, `stabPrimAtk`, `backstabHit`'s `stabChain` return): no path in the current code actually produces `DELETE_THIS`. `stabCore` doesn't propagate it (see the comment on cmd_stab.cc:133), and `backstabHit`/`throatSlitHit` consume the weapon-side `DELETE_THIS` locally. So these guards are dead code today.

Not necessarily a problem - if you wire spec-proc `DELETE_ITEM` to caller `DELETE_THIS` (per the `backstabHit` checkSpec comment) the guards earn their keep. Otherwise either drop them and let a future PR add them when something actually produces the flag, or leave a one-line comment so a future maintainer knows whether they're load-bearing.

This is just an observation - take or leave.

code/code/disc/disc_thief_murder.cc
```diff
@@ -255,20 +260,32 @@ int TBeing::doBackstab(const char* argument, TBeing* vict) {
     sendTo("Your backstab attempt has no effect on your immortal target.\n\r");
     return FALSE;
   }
-  if ((rc = backstab(this, victim))) {
-    if (!victim->isPc())
-      dynamic_cast<TMonster*>(victim)->US(25);
-    addSkillLag(SKILL_BACKSTAB, rc);
+  rc = backstab(this, victim);
+  const int lagRc = rc;
```
wershlak, line 265: The shadow variable plus the three-line comment above exist because the cleanup strips `DELETE_VICT` before `addSkillLag` would otherwise see it (and `addSkillLag` needs `DELETE_VICT` to apply the kill cap). Sibling `doCudgel` and `doGarrotte` get the same kill cap by simply calling `addSkillLag(rc)` first, then doing the victim cleanup.

With the new `IS_SET_DELETE(rc, DELETE_THIS)` early-return, you could reorder this as: `DELETE_THIS` check -> `addSkillLag(rc)` (rc still has `DELETE_VICT`) -> `DELETE_VICT` cleanup -> return. Same semantics, no shadow var, no explanatory comment. Same applies to the duplicates in `doStab` and `doThroatSlit`.

Non-blocking - just feels like simpler code if you want to make the change.

code/code/disc/disc_thief_murder.cc
```diff
@@ -255,20 +260,32 @@ int TBeing::doBackstab(const char* argument, TBeing* vict) {
...
+    delete victim;
+    victim = nullptr;
+    REM_DELETE(rc, DELETE_VICT);
+  }
+
+  if (IS_SET_DELETE(rc, DELETE_THIS))
+    return rc;
+
+  if (lagRc) {
+    addSkillLag(SKILL_BACKSTAB, lagRc);
```
wershlak, line 280: The cleanup block here (snapshot lagRc -> resolve local victim -> `DELETE_THIS` early return -> `addSkillLag` -> `AFF_HIDE` removal -> suspicion update) appears verbatim in `doStab` and `doThroatSlit` too, with only the skill constant and minor variations differing. Worth pulling into a helper so a future fix doesn't need to land at three sites in lockstep.

Something like `int finalizeStealthAttack(TBeing*& victim, TBeing* originalVict, int rc, spellNumT skill, bool clearHide, bool updateSusp)` would let each caller shrink to a couple of lines. Non-blocking, just flagging the duplication while it's fresh.

code/code/cmd/cmd_stab.cc
```diff
@@ -214,17 +165,17 @@ static int stabPrimAtk(TBeing* thief, TBeing* victim, TGenWeapon* primWeapon,
...
       }
     }
   }
 
-  return TRUE;
+  return rc;
 }
 
-static int stab(TBeing* thief, TBeing* victim) {
+static int stab(TBeing* thief, TBeing* victim, bool isChain = false) {
```
wershlak, line 180: The `bool isChain` parameter only exists to skip the move check at lines 230-235, and there's no other "isChain" pattern in the codebase to anchor on. Extracting the move check plus deduct into a small static helper that `doStab` calls before `stab()` would let `stabChain` call `stab()` directly and the boolean parameter could go away - reads more naturally at the call site too.

Feel free to ignore if you'd rather not churn the function signature.

code/code/cmd/cmd_stab.cc
```diff
@@ -235,62 +186,76 @@ static int stab(TBeing* thief, TBeing* victim) {
...
     return FALSE;
 
-  // Get weapons
   TGenWeapon* primWeapon =
     dynamic_cast<TGenWeapon*>(thief->heldInPrimHand());
   TGenWeapon* offWeapon =
     dynamic_cast<TGenWeapon*>(thief->heldInSecHand());
 
-  bool primCanStab = primWeapon && primWeapon->canStab();
+  bool primCanStab = primWeapon && (primWeapon->canStab() || primWeapon->isPolearm());
```
wershlak, line 222: Two small notes on `isPolearm()`:

1. It's composed entirely from `TBaseWeapon` predicates (`isPierceWeapon() && isPaired()`), and its sibling type predicates (`isPierceWeapon`, `isSlashWeapon`, `isBluntWeapon`) live on `TBaseWeapon`. By the existing convention, the new method belongs on `TBaseWeapon` next to those, not on `TGenWeapon` (which is where the capability checks `canStab`, `canBackstab`, `canCudgel` live).
2. `ITEM_PAIRED` is a per-instance flag toggled by `TBaseWeapon::wear`, not an inherent type marker, so a small paired pierce weapon would test as a "polearm" too. Either rename to something accurate (e.g. `isTwoHandedPierce`) or drop a one-line comment about what the check actually says.

Minor - take or leave.

code/code/misc/mobact.cc
```diff
@@ -2853,68 +2851,21 @@ int TMonster::takeFirstHit(TBeing& vict) {
...
-              if (IS_SET_DELETE(rc, DELETE_VICT)) {
-                return DELETE_VICT;
-              }
-              return TRUE;
-            }
-          }
+  if (hasClass(CLASS_THIEF)) {
+    TGenWeapon* tgw = dynamic_cast<TGenWeapon*>(heldInPrimHand());
+    if (tgw) {
+      if (tgw->canBackstab() || tgw->isPolearm()) {
```
wershlak, line 2877: Worth noting in the PR description: the predicate narrowed from "any pierce weapon" (`isPierceWeapon()`) to `canBackstab() || isPolearm()`. A pierce weapon with `getVolume()` between 1501-2000 that isn't paired no longer triggers the thief-mob opener - the mob falls through to a normal hit. Probably fine, just want to confirm that's intentional.

Also: the master version of `takeFirstHit` redirected the attack at the rider when the victim was mounted. The new code doesn't, and `backstab`/`throatSlit` reject mounted victims outright now. Net effect is mob thieves no longer use their opener against mounted PCs at all. Same question - intentional?

code/code/obj/obj_general_weapon.cc
```diff
@@ -195,17 +198,17 @@ int thornsHit(TBeing* victim, TBeing* ch, wearSlotT chLimb, wearSlotT vicLimb) {
 
   // Calculate chance based on hardness difference
   int hardnessDiff = chLimbHardness - vicLimbHardness;
```
wershlak, line 197: Most of the diff in this file is trailing-whitespace cleanup in functions unrelated to the PR's stated scope (`thornsHit`, `hardHit`, `impactSpec`, `spikesHit`). Per the project rule about keeping PRs focused, suggest reverting the whitespace-only edits and leaving just the new `isPolearm()` here. Or run `clang-format` on the file in a separate PR.

Non-blocking - this just inflates the diff for reviewers.

code/code/disc/disc_thief_murder.cc
```diff
@@ -1,3 +1,5 @@
+#include "enum.h"
```
wershlak: This file already used `POSITION_DEAD`/`POSITION_STUNNED` pre-PR without an explicit `enum.h` include, so the new `POSITION_SITTING`/`POSITION_FIGHTING` references will resolve through the same transitive chain (via `being.h`/`handler.h`). The include is unnecessary - safe to drop.

wershlak (changes requested):
Did a pass on the safety chain rework. The DELETE_VICT/DELETE_THIS plumbing in `doBackstab`/`doThroatSlit` and the new spec-proc handling in `backstabHit`/`throatSlitHit` look good in shape, but I caught one behavioral regression in `stab()` that I think wants fixing before merge - see the inline comment. The rest of what I left inline is non-blocking design/quality observations, take or leave as you see fit. A couple of FYI items below for the broader picture.

### FYI

- `cmd_stab.h` switched to `#pragma once` - fine, just noting it.
- `parse.cc` lowering the `CMD_BACKSTAB` minimum position from STANDING to FIGHTING is consistent with the new `attackers`/`fight()` gating done inside `backstab()` itself.
- In `backstab()` the success and miss branches each end with `victim->addHated(thief);` now. Functionally fine, but a future cleanup could pull the call out below the if/else.

code/code/cmd/cmd_stab.cc
kjsevertson: Fixed in baa4fc8. Restored one unconditional `victim->addHated(thief)` just before the `bSuccess` branch in `stab()` and removed the per-path duplicates in `stabCore` and the miss path. Functionally matches backstab/throatSlit (both branches add hated, equivalent to one pre-branch call).

code/code/cmd/cmd_stab.cc
kjsevertson: Fixed in baa4fc8. Converted to `IS_SET_DELETE` with the `DELETE_ITEM` → `DELETE_THIS` translation so `stabPrimAtk`'s offhand swing can't run against a dead thief, and combined returns are now handled.

code/code/disc/disc_thief_murder.cc
kjsevertson: Fixed in baa4fc8. Added `if (IS_SET_DELETE(specRc, DELETE_ITEM)) return DELETE_THIS;` between the existing DELETE_THIS (weapon-dead) and DELETE_VICT handlers. The existing caller guards in `doBackstab` now catch the thief-dead case.

code/code/disc/disc_thief_murder.cc
kjsevertson: Fixed in baa4fc8. Same pattern as `backstabHit` — `DELETE_ITEM` now returns `DELETE_THIS` so the caller's existing guards catch it.

code/code/disc/disc_thief_murder.cc
kjsevertson: Resolves itself with the `DELETE_ITEM` → `DELETE_THIS` translation added in baa4fc8 to backstabHit/throatSlitHit/stabCore. The DELETE_THIS guards now have a real producer and earn their keep.

code/code/disc/disc_thief_murder.cc
kjsevertson: Declining. The suggested reorder leaks in the combined `DELETE_THIS | DELETE_VICT && !vict` case: the early `DELETE_THIS` return would propagate `DELETE_VICT` to a caller that has no reference to the locally-resolved victim. The shadow var + comment exist precisely because the cleanup must happen *before* the `DELETE_THIS` early return so combined returns don't leak. `doCudgel`/`doGarrotte` don't hit this because `cudgel()`/`garrotte()` never return `DELETE_THIS` — the stab chain now does (after the DELETE_ITEM translation), so the combined case is plausible here.

code/code/disc/disc_thief_murder.cc
kjsevertson: Deferring. The `skulk` branch already adds a `breakStealth()` helper (loseSneak + loseSkulk) with an explicit comment anticipating `AFF_HIDE` folding in once `willBreakHide`'s pipeline is unified. Once that follow-up lands, the explicit `REMOVE_BIT(AFF_HIDE)` lines in the cleanup blocks dissolve and what remains is small enough not to warrant a dedicated helper. Plan captured for the follow-up.

code/code/cmd/cmd_stab.cc
kjsevertson: Deferring. Cosmetic only, and the lag/move-check region is touchier than I want to refactor in this PR. Will revisit if the bool becomes a maintenance burden.

code/code/cmd/cmd_stab.cc
kjsevertson: Done in f612887. Moved `isPolearm()` from `TGenWeapon` to `TBaseWeapon` next to `isPierceWeapon`/`isSlashWeapon`/`isBluntWeapon`. Kept the name and added a comment at the implementation noting `ITEM_PAIRED` is a per-instance wear flag (not a type marker), so small paired pierce weapons also qualify — that's intentional.

code/code/misc/mobact.cc
kjsevertson: Both intentional. The pierce-weapon-volume-1501-2000-not-paired case: master's `isPierceWeapon()` was too loose — mobs would attempt to backstab with weapons that can't actually backstab, then fail inside `backstab()`. The new `canBackstab() || isPolearm()` is more precise.

The mounted-PC case: master's `takeFirstHit` called `doBackstab("", v2)` where `v2` is the mount, but master's own `backstab()` had (and still has) a rider-rejection gate (`if (dynamic_cast<TBeing*>(victim->rider))`) — and the mount has the PC as rider, so master's own gate rejected it. The horse-scrag workaround was already dead code. Current branch falls through to a normal swing, same effective behavior without the wasted call path.

code/code/obj/obj_general_weapon.cc
kjsevertson: Reverted in f612887. All four functions (`thornsHit`, `hardHit`, `impactSpec`, `spikesHit`) back to master. `obj_general_weapon.{cc,h}` now have zero diff vs master — the new `isPolearm()` lives in `obj_base_weapon.cc` alongside the sibling predicates.

code/code/disc/disc_thief_murder.cc
kjsevertson: Dropped in baa4fc8. Confirmed `POSITION_*` symbols resolve transitively through `being.h`/`handler.h`.

wershlak (approved):

Issue #678, opened 2026-03-05
Undefined behavior: value is outside range of representable values of type 'int' in `engraveCost()`


drewhershey:
```
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-2026|0221|02:11:46 :: Everlast (569):value dev
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-/home/runner/work/sneezymud/sneezymud/code/code/spec/spec_mobs_customizers.cc:43:15: runtime error: 1.78154e+14 is outside the range of representable values of type 'int'
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #0 0x5f34aafd33a0 in engraveCost(TObj*, TBeing*, unsigned int) /home/runner/work/sneezymud/sneezymud/code/code/spec/spec_mobs_customizers.cc:43:15
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #1 0x5f34aafcee8a in customizer(TBeing*, cmdTypeT, char const*, TMonster*, TObj*, customizerType) /home/runner/work/sneezymud/sneezymud/code/code/spec/spec_mobs_customizers.cc:290:14
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #2 0x5f34aafc6475 in engraver(TBeing*, cmdTypeT, char const*, TMonster*, TObj*) /home/runner/work/sneezymud/sneezymud/code/code/spec/spec_mobs_customizers.cc:47:10
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #3 0x5f34aae6b6d7 in TMonster::checkSpec(TBeing*, cmdTypeT, char const*, TThing*) /home/runner/work/sneezymud/sneezymud/code/code/spec/spec_mobs.cc:107:10
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #4 0x5f34a9eb0cd8 in TBeing::triggerSpecial(TThing*, cmdTypeT, char const*) /home/runner/work/sneezymud/sneezymud/code/code/misc/parse.cc:2442:15
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #5 0x5f34a9e99fe5 in TBeing::doCommand(cmdTypeT, sstring const&, TThing*, bool) /home/runner/work/sneezymud/sneezymud/code/code/misc/parse.cc:359:10
[...]

Issue #675, opened 2026-03-05
Undefined behavior: value outside range of representable values of type 'int' in `TBaseWeapon::suggestedPrice()`


drewhershey:
```
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-2026|0221|01:51:23 :: Redwood renting in <d>a diamond shoe<1>. (max: 9999, cur: 27)
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-/home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_weapon.cc:1249:20: runtime error: 4.88287e+09 is outside the range of representable values of type 'int'
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #0 0x5f34aac2d2fb in TBaseWeapon::suggestedPrice() const /home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_weapon.cc:1249:20
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #1 0x5f34a9300f65 in read_object(int, readFileTypeT) /home/runner/work/sneezymud/sneezymud/code/code/sys/db.cc:2569:23
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #2 0x5f34aa00e439 in ItemLoad::raw_read_item() /home/runner/work/sneezymud/sneezymud/code/code/misc/rent.cc:552:13
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #3 0x5f34aa01fdb4 in ItemLoad::objsFromStore(TObj*, int*, TBeing*, TRoom*, bool) /home/runner/work/sneezymud/sneezymud/code/code/misc/rent.cc:1038:21
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #4 0x5f34aa0365b2 in TPerson::loadRent() /home/runner/work/sneezymud/sneezymud/code/code/misc/rent.cc:2100:10
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #5 0x5f34a91d8104 in TPerson::genericLoadPC() /home/runner/work/sneezymud/sneezymud/code/code/sys/connect.cc:1378:3
/var/tmp/sneezymud-logs/sneezy-20260221-141406.log-    #6 0x5f34a91cf224 in Descriptor::nanny(sstring) /home/runner/work/sneezymud/sneezymud/code/code/sys/connect.cc:1283:47
[...]

Issue #674, opened 2026-03-04
Undefined behavior: Value outside range of representable values of type 'int' in `TBaseClothing::armorPriceStruct()`


drewhershey:
```
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-2026|0208|23:00:20 :: Virion renting in <k>an ebon-grey pendant<z>. (max: 9999, cur: 42)
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-/home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_clothing.cc:351:23: runtime error: 1.72714e+10 is outside the range of representable values of type 'int'
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #0 0x5977e58f1d98 in TBaseClothing::armorPriceStruct(armorLevT, double*) const /home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_clothing.cc:351:23
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #1 0x5977e58f203e in TBaseClothing::suggestedPrice() const /home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_clothing.cc:386:15
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #2 0x5977e58f81e5 in virtual thunk to TBaseClothing::suggestedPrice() const /home/runner/work/sneezymud/sneezymud/code/code/obj/obj_base_clothing.cc
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #3 0x5977e3f15d80 in read_object(int, readFileTypeT) /home/runner/work/sneezymud/sneezymud/code/code/sys/db.cc:2569:23
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #4 0x5977e4ce6fba in ItemLoad::raw_read_item() /home/runner/work/sneezymud/sneezymud/code/code/misc/rent.cc:552:13
/var/tmp/sneezymud-logs/sneezy-20260210-214254.log-    #5 0x5977e4cf96d0 in ItemLoad::objsFromStore(TObj*, int*, TBeing*, TRoom*, bool) /home/runner/work/sneezymud/sneezymud/code/code/misc/rent.cc:1038:21
[...]

PR #706, opened 2026-04-01
Fix UB engravecost and friends

wershlak:
I tried to just fix one thing but in order to test with the UB sanitizer on I had to fix some adjacent issues.

I think this should fix these issues:

  - #678 — engraveCost() in spec_mobs_customizers.cc — fixed (the original issue)
  - #675 — TBaseWeapon::suggestedPrice() in obj_base_weapon.cc — fixed (the std::clamp on the hitroll formula)
  - #674 — TBaseClothing::armorPriceStruct() in obj_base_clothing.cc — fixed (clamping ac_comp/str_comp)

### DIFF
--- a/code/code/misc/extern.h
+++ b/code/code/misc/extern.h
@@ -16,8 +16,14 @@
 
 #include <sys/select.h>  // for fd_set
 #include <cmath>         // for pow
+#include <climits>       // for INT_MAX, INT_MIN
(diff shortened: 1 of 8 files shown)


drewhershey (dismissed):
Looks good to me. Left a few non-blocking inline comments you can just ignore if you don't find them relevant or want to address them.

drewhershey (approved):

PR #717, opened 2026-04-06
Improve hide behavior: feedback message and expanded whitelist

kjsevertson:
## Summary
- Adds "You move from your hiding spot." message when a player breaks hide, giving feedback instead of silently dropping the affect
- Expands the hide-preserving whitelist with thematically appropriate commands: rest, sit, stand, sign, ptell, conceal, eat, drink
- Fixes a bug in `parseCommand` where a legacy hide-removal path bypassed the `willBreakHide()` whitelist — two separate if-statements (one exempting backstab, one exempting slit) together exempted neither

## Test plan
- [ ] Hide, then use a whitelisted command (look, rest, sit, eat, etc.) — should stay hidden with no message
- [ ] Hide, then use a non-whitelisted command (say, move, cast) — should see "You move from your hiding spot." and lose hide
- [ ] Hide, then backstab/slit — should execute the attack from hiding, then break hide after

### DIFF
--- a/code/code/disc/disc_thief_stealth.cc
+++ b/code/code/disc/disc_thief_stealth.cc
@@ -936,7 +936,9 @@ int hide(TBeing* thief, spellNumT skill) {
 
   if (thief->bSuccess(bKnown, skill)) {
     SET_BIT(thief->specials.affectedBy, AFF_HIDE);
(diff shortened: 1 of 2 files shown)


drewhershey:
We decided to pull these changes into https://github.com/sneezymud/sneezymud/pull/536 instead.

PR #728, opened 2026-05-16
Allow slam and whirlwind unarmed (or with gauntlets)

kjsevertson:
## Summary
- Drop the "must be holding a weapon" gate on `slam` and `whirlwind`. Empty primary hand is now fine; non-weapon items (lights, scrolls, instruments) are still rejected with a descriptive message.
- Unarmed damage source falls through to the primary hand slot, so a worn gauntlet/glove contributes via `impactSpec`. Truly bare hands deal `DAMAGE_NORMAL` with no blunt/pierce/slash class.
- Messages substitute the implement's noun (`$o`) when present and fall back to the race-appropriate body-part noun (`describeBodySlot`) when bare — so trolls slam with claws, centaurs with hooves, etc.
- `whirlwind` also picks up the per-victim `impactSpec` contribution that `slam` already had; the previous code skipped impact for whirlwind entirely.

## Test plan
- [ ] `slam` armed with a blunt/pierce/slash weapon still produces the right damage class and messages
- [ ] `slam` unarmed with no glove → `DAMAGE_NORMAL`, bare-hand noun in act messages, no crash
- [ ] `slam` unarmed wearing a gauntlet → impactSpec from the gauntlet applies; messages still use body-part noun
- [ ] `slam` while holding a non-weapon (light, scroll) → rejection message names the held item
- [ ] `whirlwind` armed: unchanged behavior plus impactSpec now applies per victim
- [ ] `whirlwind` unarmed / with gauntlet: parallels slam behavior
- [ ] `whirlwind` holding a non-weapon → rejected
- [ ] Berserk-mode slam messages render correctly with both `$o` and bare-hand noun

### DIFF
--- a/code/code/cmd/cmd_slam.cc
+++ b/code/code/cmd/cmd_slam.cc
@@ -67,11 +67,10 @@ int TBeing::doSlam(const char* argument, TBeing* vict) {
     return false;
   }
 
-  // Ensure the player has a weapon equipped
-  if (!weapon) {
-    sendTo(
-      "You need to hold a weapon in your primary hand to make this a "
-      "success.\n\r");
+  // Empty hand is fine (unarmed slam). Reject only when something non-weapon
+  // is held (light, scroll, instrument, etc.).
+  if (auto* held = heldInPrimHand(); held && !weapon) {
+    act("You can't slam with $p.", false, this, held, nullptr, TO_CHAR);
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


code/code/cmd/cmd_whirlwind.cc
```diff
@@ -19,6 +20,13 @@ static int whirlwind(TBeing* caster, TBeing* victim, int castLevel,
       TO_VICT);
     dam = caster->getSkillDam(victim, SKILL_WHIRLWIND, castLevel,
       caster->getAdvLearning(SKILL_WHIRLWIND));
+
+    // Add weapon/gauntlet/bare-hand impact contribution per victim.
+    auto* weapon = dynamic_cast<TBaseWeapon*>(caster->heldInPrimHand());
+    wearSlotT damSource =
+      weapon ? caster->getPrimaryHold() : caster->getPrimaryHand();
+    wearSlotT targetLimb = victim->getPartHit(caster, true);
+    dam += impactSpec(caster, victim, damSource, targetLimb);
```
wershlak, line 29: The `weapon` and `damSource` here don't depend on the victim, but `whirlwind()` is called once per victim from `whirlwindSuccess`, so this re-runs the `dynamic_cast` and slot resolution every iteration. Lifting these two lines into `whirlwindSuccess` and passing `damSource` into this helper as a parameter (the same way `damageType` is already plumbed through) avoids the per-victim work and matches the existing pattern.

Feel free to leave it for a follow-up if you'd rather keep this PR narrowly scoped to the unarmed support.

code/code/cmd/cmd_slam.cc
```diff
@@ -139,60 +138,86 @@ int TBeing::slamSuccess(TBeing* victim) {
   // Apply the scaling constant
   dam = max((int)(victim->hitLimit() * scalingConstant), dam);
 
+  // Resolve the implement: weapon if held, else worn glove/gauntlet, else
+  // null (truly bare). Picks the impact-damage source slot and the noun shown
+  // in the act messages.
+  TThing* implement =
```
wershlak, line 144: The four-line `weapon` / `implement` / `noun` resolution is now repeated in `slamSuccess` and `slamFail`, with a parallel pattern in `cmd_whirlwind.cc`. `doSlam` already does the `dynamic_cast<TBaseWeapon*>(heldInPrimHand())` at the top of the function too. Now that "what counts as an implement" is non-trivial (weapon then worn glove then bare hand), a small TBeing helper returning `{implement, damSource, noun}` - or threading the resolved values down from `doSlam` to `slamSuccess`/`slamFail` - would give one place to update if the rule ever evolves (off-hand weapon, bracers, etc.).

Nit-level; happy to ship as-is if you'd rather keep the scope tight.

wershlak (commented):
Nothing blocking - the unarmed path is well-thought-out and the safety story is solid (impactSpec already handles the null-equipment case, and the body-part noun substitution stays clear of format-string hazards). A couple of small design suggestions inline, both optional. One non-PR FYI below.

### FYI

- The `slamFail` rewrite quietly fixes two pre-existing typos in the TO_NOTVICT message at the old `cmd_slam.cc:189-191`: `$N's` (extra possessive on the victim) and `$s his weapon` (`$s` already expands to a possessive). Probably worth a line in the commit/PR description so future bisects can attribute the rendering change.

wershlak (approved):

